PR #2, opened 2019-07-12
Add Victor Stinner's Performance talk

vstinner:


vstinner:
cc @antocuni

vstinner:
Well, it should also copy relevent part of my slides into the document, but I don't have the bandwidth right now.

PR #8, opened 2019-07-24
mention Jython in README

deronnax:
It's a pretty important one (Django documentation even have a page dedicated to Jython)

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ The goal of the project is to design a better API for extending Python
 in C. The current API is specific to the current implementation of CPython: it
 exposes a lot of internal details which makes it hard:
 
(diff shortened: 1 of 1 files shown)


vstinner:
Jython only supports Python 2, whereas this project targets Python 3 support only...

antocuni:
True but it's also a good and well-known example of "alternative Python implementation". It will not be immediately useful for Jython in its current state, but it will be in case Jython will ever decide to migrate to Python 3

deronnax:
IronPython also only target Python2 and it was listed :)

Le lun. 29 juil. 2019 à 15:58, Antonio Cuni <notifications@github.com> a
écrit :

arigo:
And, fwiw, I don't expect Python 2 support to be hard to do so I may do it once we have a Python 3 story.

PR #14, opened 2019-12-18
Change handle content type to 'void *'.

fangerer:
This changes the type of member `_i` (`struct _HPy_s`) to `void *`.
The reason for this change is that `void *` is a more abstract type and does not force a runtime implementation to really represent their handles via integers.
In particular, in GraalVM's implementation of Python, we do not run the C code natively. Therefore, a pointer would allow us to keep the handle virtual whereas an integer will force us to actually allocate the handle.

For runtime implementations written in C/C++, the change should not impose any overhead since the `void *` can still be casted to `HPy_ssize_t`.

### DIFF
--- a/hpy-api/hpy_devel/include/universal/hpy.h
+++ b/hpy-api/hpy_devel/include/universal/hpy.h
@@ -7,7 +7,7 @@
 
 typedef intptr_t HPy_ssize_t;
 
-struct _HPy_s { HPy_ssize_t _i; };
+struct _HPy_s { void* _i; };
 typedef struct _HPy_s HPy;
 
 typedef struct _HPyContext_s *HPyContext;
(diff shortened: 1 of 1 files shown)


arigo:
You're missing the point of ``intptr_t``.  It's supposed to be an integer large enough to store a ``void *``.  That means that the two types are more or less equivalent nowadays, but ``intptr_t`` is actually *more* abstract than ``void *`` in the sense that if an implementation wants to represent handles with pointers, it can (because it's guaranteed that a cast from any pointer to ``intptr_t`` and back gives the same pointer), but it can also use integers that mean anything else.  Of course in practice your commit doesn't really change that, because we can very likely cast any integer to a ``void *`` and back to an integer, too.  It's just not guaranteed a priori.

timfel:
@arigo but the point is that a `void*` really doesn't have to be an integer, and that the compiler will generate different code for pointer arithmetic than if it already thinks it's an integer. that forces us to allocate an integer eagerly and create a mapping, even though we don't actually need that

arigo:
I apologize for not understanding you at all.  I'm talking about casting your pointers to  ``intptr_t``; not using some mapping.  Don't think of ``intptr_t`` as a type that can only store small integers.  It's not ``int``.  If you look at the ``intptr_t`` as an integer then it will contain large values like 0x7ffff1029a8.  You should think of it as N bytes of memory where ``N = sizeof(void *)``.

arigo:
Ah, maybe you mean that on Graal you're compiling C files with a custom compiler, in which the cast from ``void *`` to ``intptr_t`` is expensive.  If that's the case, then OK, now I see the point.  Maybe fix it by using ``#ifdef GRAAL`` to compile different code, with a comment explaining the difference?

timfel:
@arigo sounds good, we weren't sure if an `#ifdef` for this would be acceptable

arigo:
In fact, given that you are in a situation where recompiling is needed anyway, you can declare the HPy type freely like you want, if another variant would be more helpful to you.

arigo:
(also, I'm sorry, I didn't actually mean to close the pull request already after my first comment - I realise now that I did but I'm not sure how)

antocuni:
@timfel do you plan to load universal extensions or to require them to be recompiled specifically for graal?
In the first case, the C compiler will emit the very same code for `intptr_t` or `void*`. In the second case, you probably want to distribute a custom header file anyway, so you are free to put whatever you want for the types, unless I am missing anything

PR #19, opened 2020-03-01
Implement support for keyword arguments to methods.

hodgestar:


hpy/devel/include/cpython/hpy.h
```diff
@@ -21,6 +21,12 @@
 
 #define HPyAPI_FUNC(restype) HPyAPI_STORAGE restype
 
+#ifdef __GNUC__
+#define HPyAPI_RUNTIME_FUNC(restype) __attribute__((visibility("hidden"))) restype
```
antocuni: should this use `_HPy_HIDDEN`?

hpy/devel/include/universal/hpy.h
```diff
@@ -13,7 +13,8 @@ typedef struct _HPy_s HPy;
 typedef struct _HPyContext_s *HPyContext;
 struct _object;  /* that's PyObject inside CPython */
 typedef struct _object *(*_HPy_CPyCFunction)(struct _object *self,
-                                             struct _object *args);
+                                             struct _object *args,
```
antocuni: This is wrong. `_HPy_CPyCFunction` is a direct translation of CPython's `PyCFunction`, so it should stay the same

hpy/devel/include/universal/meth.h
```diff
@@ -35,15 +35,18 @@ typedef struct {
...
 #define HPy_DECL_METH_VARARGS(NAME) HPy_DECL_METH_NOARGS(NAME)
+#define HPy_DECL_METH_KEYWORDS(NAME) HPy_DECL_METH_NOARGS(NAME)
 
 
 #define HPy_DEF_METH_NOARGS(fnname)                                            \
     static HPy fnname##_impl(HPyContext ctx, HPy self);                        \
     static struct _object *                                                    \
-    fnname##_trampoline(struct _object *self, struct _object *noargs)          \
+    fnname##_trampoline(struct _object *self, struct _object *noargs,          \
+                        struct _object *nokw)                                  \
```
antocuni: ...and this is wrong, because CPython will call it using its own `PyCFunction` signature

hpy/universal/src/api.c
```diff
@@ -121,6 +121,29 @@ ctx_CallRealFunctionFromTrampoline(HPyContext ctx, struct _object *self,
     }
 }
 
+typedef HPy (*HPyMeth_Keywords)(HPyContext, HPy self, HPy *args, HPy_ssize_t,
+                                HPy kw);
+
+static struct _object *
+ctx_CallRealFunctionWithKeywordsFromTrampoline(HPyContext ctx,
```
antocuni: strictly speaking this is not wrong. However, it is possible to implement the same functionality by adding a case to the existing `CallRealFunctionFromTrampoline`, by adding a `*kw` param which will be `NULL` in the non-keyword cases

hpy/universal/src/api.c
```diff
@@ -151,52 +174,6 @@ ctx_Dup(HPyContext ctx, HPy h)
     return _py2h(obj);
 }
 
-
```
antocuni: hoorray for removing code duplication 🎉

hpy/devel/include/universal/meth.h
hodgestar: Fixed!

hpy/devel/include/universal/hpy.h
hodgestar: Fixed!

hpy/universal/src/api.c
hodgestar: :D

test/test_argparse.py
```diff
@@ -0,0 +1,286 @@
...
+
+    def test_blank_keyword_argument_exception(self):
+        import pytest
+        mod = self.make_module("""
+            HPy_DEF_METH_KEYWORDS(f)
+            static HPy f_impl(HPyContext ctx, HPy self,
+                              HPy *args, HPy_ssize_t nargs, HPy kw)
+            {
+                HPy a, b, c;
+                b = HPyLong_FromLong(ctx, 5);
```
antocuni: this is leaking!

test/test_argparse.py
```diff
@@ -0,0 +1,286 @@
...
+
+    def test_positional_only_argument(self):
+        import pytest
+        mod = self.make_module("""
+            HPy_DEF_METH_KEYWORDS(f)
+            static HPy f_impl(HPyContext ctx, HPy self,
+                              HPy *args, HPy_ssize_t nargs, HPy kw)
+            {
+                HPy a, b;
+                b = HPyLong_FromLong(ctx, 5);
```
antocuni: this as well

test/test_basic.py
```diff
@@ -245,11 +213,22 @@ def test_decl_meth(self):
                     return HPy_NULL;
                 return HPyLong_FromLong(ctx, 10*a + b);
             }
+            HPy_DEF_METH_KEYWORDS(i)
+            static HPy i_impl(HPyContext ctx, HPy self, HPy *args, HPy_ssize_t nargs,
+                              HPy kw)
+            {
+                // TODO: Also parse kw args.
```
antocuni: fix the TODO

tools/public_api.h
```diff
@@ -51,6 +52,7 @@ int HPyList_Append(HPyContext ctx, HPy h_list, HPy h_item);
 /* dictobject.h */
 HPy HPyDict_New(HPyContext ctx);
 int HPyDict_SetItem(HPyContext ctx, HPy h_dict, HPy h_key, HPy h_val);
+HPy HPyDict_GetItem(HPyContext ctx, HPy h_dict, HPy h_key);
```
antocuni: this is probably "wrong": we don't want API which return borrowed references. Please add a comment so we remember to fix

tools/autogen.py
```diff
@@ -289,9 +267,11 @@ def gen_func_implementations(self):
                        'HPy_Dup',
                        'HPy_Close',
                        'HPyArg_Parse',
```
antocuni: kill these two: they are no longer in `public_api.h`, so there is no point to exclude them

hpy/universal/src/api.c
hodgestar: Fixed!

test/test_basic.py
hodgestar: Fxied!

tools/autogen.py
hodgestar: Fixed!

tools/public_api.h
hodgestar: Comment added!

test/test_argparse.py
hodgestar: Fixed!

test/test_argparse.py
hodgestar: Fixed!

hpy/devel/include/cpython/hpy.h
hodgestar: Fix by defining _HPy_HIDDEN more nicely and then defining HPyAPI_RUNTIME_FUNC in terms of _HPy_HIDDEN.

PR #20, opened 2020-03-03
WIP autogen: Rust definition of HPyContext

gracinet:
This is good enough to go on experimenting, but there are
chances that eventually, using the
[bindgen crate](https://crates.io/crates/bindgen)
on the generated C definition would be a better idea.

This depends on whether we'll need more than just struct definition
translations.

### DIFF
--- a/tools/autogen.py
+++ b/tools/autogen.py
@@ -25,6 +25,40 @@ def toC(node):
 toC.gen = CGenerator()
 
 
+RUST_TYPE_MAPPING = {
+    ('unsigned', 'long', 'long'): 'c_ulonglong',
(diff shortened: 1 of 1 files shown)


thedrow:
This is very useful for PyO3 and similar projects.

auvipy:
> This is very useful for PyO3 and similar projects.

indeed. we can try on celery as well.

thedrow:
#325 proposes a more generic solution using bindgen as @gracinet has already suggested.
Since this PR is stale and isn't going to be merged at the time being, I'm going to close it.

Thanks for the contribution!

PR #21, opened 2020-03-05
Implement object protocol item and attribute access.

hodgestar:


hodgestar:
Ready for review!

test/test_object.py
```diff
@@ -0,0 +1,63 @@
...
+
+class TestObject(HPyTest):
+    def test_getitem(self):
+        import pytest
+        mod = self.make_module("""
+            HPy_DEF_METH_O(f)
+            static HPy f_impl(HPyContext ctx, HPy self, HPy arg)
+            {
+                HPy key, result;
+                key = HPyLong_FromLong(ctx, 0);
```
antocuni: I think it would be better to use a non-common value here, instead of `0`, to minimize the chances that a test passes "by mistake" in case somewhere someone uses a default value or something like that

test/test_object.py
```diff
@@ -0,0 +1,63 @@
...
+            mod.f([])
+
+    def test_getitem_i(self):
+        import pytest
+        mod = self.make_module("""
+            HPy_DEF_METH_O(f)
+            static HPy f_impl(HPyContext ctx, HPy self, HPy arg)
+            {
+                HPy result;
+                result = HPy_GetItem_i(ctx, arg, 0);
```
antocuni: same as above

test/test_object.py
```diff
@@ -84,3 +84,31 @@ def test_getitem_s(self):
...
+
+    def test_setitem(self):
+        import pytest
+        mod = self.make_module("""
+            HPy_DEF_METH_O(f)
+            static HPy f_impl(HPyContext ctx, HPy self, HPy arg)
+            {
+                HPy key;
+                int result;
+                key = HPyLong_FromLong(ctx, 0);
```
antocuni: same here, better to use a more unique value

test/test_object.py
```diff
@@ -112,3 +112,26 @@ def test_setitem(self):
...
             mod.f([])
+
+    def test_setitem_i(self):
+        import pytest
+        mod = self.make_module("""
+            HPy_DEF_METH_O(f)
+            static HPy f_impl(HPyContext ctx, HPy self, HPy arg)
+            {
+                int result;
+                result = HPy_SetItem_i(ctx, arg, 0, ctx->h_True);
```
antocuni: same

test/test_object.py
```diff
@@ -10,6 +10,48 @@
...
+                HPy_Close(ctx, name);
+                if (HPy_IsNull(result))
+                    return HPy_NULL;
+                return result;
+            }
+            @EXPORT f HPy_METH_O
+            @INIT
+        """)
+
+        class Attrs:
```
antocuni: good idea to use many different kinds of attrs to test! 👍👍👍

test/test_object.py
hodgestar: Being addressed in #27.

test/test_object.py
hodgestar: Being addressed in #27.

test/test_object.py
hodgestar: Being addressed in #27.

test/test_object.py
hodgestar: Being addressed in #27.

test/test_object.py
hodgestar: :D

PR #22, opened 2020-03-06
Implement types

rlamy:


antocuni:
I think that the branch is ready to be reviewed and hopefully merged.
Highlights of the branch:
  - Introduce the new struct `HPyType_Spec`
  - Introduce `HPyType_FromSpec`, `HPy_CAST` and `HPy_New`

It implements enough functionalities that it is now possible to implement simple types, with methods and slots.
During the design, special care has been taken to ensure that it will be possible to make an incremental transition of existing types: in particular, it will be possible to:
  - Have some methods written in HPy and some others written using the CPython API
  - For a given `struct FooObject`, it will be possible to choose whether to include `PyObject_HEAD` or not, by setting `has_pyobject_head` in the TypeSpec. This will make it easier to port the code which casts `PyObject *` to e.g. `FooObject *`. Once the porting is complete, `PyObject_HEAD` should be removed (and `Python.h` no longer included).
  - Note that none of this has been implement yet, though.

Note that in the CPython ABI mode, `HPyType_Spec` and `HPyType_FromSpec` are **not** a aliases for their CPython counterpart. This was needed to support `has_pyobjecy_head`, as described above.

Among the other things, the branch shuffled some code around. In particular, `hpy.universal` used to have all the code inside one big `api.c` file, which now has been split into many `ctx_*.c`, together with the appropriate headers.

Finally, it also contains a refactoring of the `test/support.py`: now it is easier to introduce new `@FOO` "macros" to use inside `self.make_module`. The new `@EXPORT_TYPE` "macro" has been introduced.

@arigo I would like to have your feedback, since I discussed most of the subtle details with you :)

hpy/devel/include/universal/autogen_trampolines.h
```diff
@@ -236,6 +240,14 @@ static inline int HPy_SetItem_s(HPyContext ctx, HPy obj, const char *key, HPy va
      return ctx->ctx_SetItem_s ( ctx, obj, key, value ); 
 }
 
+static inline void *_HPy_Cast(HPyContext ctx, HPy h) {
+     return ctx->ctx_Cast ( ctx, h ); 
+}
+
+static inline HPy HPy_New(HPyContext ctx, HPy h_type, void **data) {
+     return ctx->ctx_New ( ctx, h_type, data );
```
arigo, line 248: A pure performance hack... Writing it this way is probably better for callers:

`void *data_result; HPy h = ctx->ctx_New(ctx, h_type, &data_result); *data = *data_result; return h;`

Without it, the C compiler must assume that if the caller passes, say, `&point_data` then anybody could write a different value at any time into this local variable because a pointer to it escaped.  With this hack, it's no longer the case: what escaped is the address of `data_result` instead and that local variable disappears.

hpy/devel/src/runtime/ctx_type.c
```diff
@@ -0,0 +1,212 @@
...
+    // CPython ABI mode
+    dst->pfunc = src->pfunc;
+#endif
+}
+
+
+_HPy_HIDDEN void*
+ctx_Cast(HPyContext ctx, HPy h)
+{
+    // XXX: how do we implement ctx_Cast when has_pyobject_head==1?
```
arigo, line 86: For now we should check the type object to see if has_pyobject_head==1 or not.  It doesn't seem like a big overhead to me.  We can store that info in a quickly-accessible place, like there are probably tp_flags bits we can abuse.

hpy/devel/src/runtime/ctx_type.c
```diff
@@ -0,0 +1,212 @@
...
+    spec->slots = create_slot_defs(hpyspec);
+    if (spec->slots == NULL) {
+        PyMem_Free(spec);
+        PyErr_NoMemory();
+        return HPy_NULL;
+    }
+    PyObject *result = PyType_FromSpec(spec);
+    /* note that we do NOT free the memory which was allocated by
+       create_method_defs, because that one is referenced internally by
+       CPython (which probably assumes it's statically allocated) */
```
arigo, line 191: We should probably come up with a general way to attach extra C-level information to type objects created by HPy.  There doesn't seem to be an official way.  We can find hacks though...

hpy/devel/src/runtime/ctx_type.c
antocuni: Yes, I was also thinking about (ab)using `tp_flags`.
What I don't like is that in theory `HPy_CAST` was supposed to be a purely static cast for the sake of the typesystem, now it becomes something which has a runtime behavior, with a bit of overhead (although I admit it's probably very small).

But I suppose that for now it's fine.

hpy/devel/src/runtime/ctx_type.c
antocuni: Well, for this specific case I claim that it's CPython's fault, but I agree about the general idea. Are you ok to merge this PR now and think about this "eventually"? :)

hpy/devel/include/universal/autogen_trampolines.h
antocuni: interesting, I'd have never thought about that.
Fixed by 1396a4a.

hpy/devel/src/runtime/ctx_type.c
antocuni: Resolve the issue and merging it anyway, we can always fix it later

hpy/devel/include/common/runtime/ctx_type.h
```diff
@@ -0,0 +1,18 @@
+#ifndef HPY_COMMON_RUNTIME_CTX_TYPE_H
+#define HPY_COMMON_RUNTIME_CTX_TYPE_H
+
+#include <Python.h>
```
hodgestar, line 4: Why do we include Python.h here?

hpy/devel/src/runtime/argparse.c
```diff
@@ -4,36 +4,45 @@
...
+    case 'l': {
+        long *output = va_arg(vl, long *);
+        _BREAK_IF_OPTIONAL(current_arg);
+        long value = HPyLong_AsLong(ctx, current_arg);
+        if (value == -1 && HPyErr_Occurred(ctx))
+            return 0;
+        *output = value;
+        break;
+    }
+    /* case 'd': { */
```
hodgestar, line 26: Is something still needed so that we can uncomment this?

hpy/devel/include/common/runtime/ctx_type.h
antocuni: Because this is the header file corresponding to `ctx_type.c`, which is used in two places:
1. in the CPython-ABI mode (in this case we are obviously allowed to include Python.h)
2. to implement `hpy.universal` for CPython (and we are again allowed to include Python.h)

Note that this file is **not** included nor used if you just `#include <hpy.h>` in universal mode.
However, I agree that it's confusing to have it under `hpy/devel/include/common`: probably this means that we should restructure the layout in a way which makes more clear what is the interface and what is the implementation.

Maybe something like:
1. `hpy/devel/include`, which contains everything needed to compile an universal extension. No `Python.h` allowed there
2. `hpy/devel/src` (or `common`, `runtime`, `cpython`, ...) which contains implementation files which are used **both** by `hpy/devel/include/cpython` and `hpy/universal/src/` ?

hpy/devel/src/runtime/argparse.c
antocuni: I think it was added by @rlamy when he was experimenting with types, but there are no tests for it and it was broken (for example, it doesn't even compile because it calls `HPyErr_Occurred()` without passing the ctx.
It is probably easy to fix but it should have a test and should be done in its own branch.

PR #23, opened 2020-03-06
Avoid usage of void * for generic function pointers

gracinet:
On some architectures, data and instruction pointers have
different sizes, meaning that void * and function pointers
become incompatible.

We introduce _HPy_FuncPtr as the generic pointer type for
functions returning HPy, so that it can be used in HPyMeth

### DIFF
--- a/hpy/devel/include/universal/meth.h
+++ b/hpy/devel/include/universal/meth.h
@@ -1,6 +1,9 @@
 #ifndef HPY_UNIVERSAL_METH_H
 #define HPY_UNIVERSAL_METH_H
 
+/* Pointer to a function returning HPy, with unspecified arguments */
+typedef HPy (*_HPy_FuncPtr)();
+
 /* in universal mode, an HPyMeth is a function which returns two output
  * arguments:
  *
@@ -12,7 +15,8 @@
  * generate them on the fly, and for the sake of simplicity it is easier to
  * just let the C compiler to generate it. This is done by the DEF macros.
  */
-typedef void (*HPyMeth)(void **out_func, _HPy_CPyCFunction *out_trampoline);
+typedef void (*HPyMeth)(_HPy_FuncPtr *out_func,
+                        _HPy_CPyCFunction *out_trampoline);
 
 /* ml_flags can be:
  *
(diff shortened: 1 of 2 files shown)


hpy/devel/include/universal/meth.h
```diff
@@ -1,6 +1,9 @@
 #ifndef HPY_UNIVERSAL_METH_H
 #define HPY_UNIVERSAL_METH_H
 
+/* Pointer to a function returning HPy, with unspecified arguments */
```
antocuni: I would add a comment here explaining why we can't use `void*`

gracinet:
@antocuni good idea, will add links to appropriate documentation.

Note that since then, I've spotted a few other cases, and we discussed with @arigo and @rlamy about the that possibility that new style type declarations of current CPython could possibly ignore the thing altogether. If that's true, what we can do will be limited anyway (but I still think having the correct types where we can is a good idea).

antocuni:
> Note that since then, I've spotted a few other cases, and we discussed with @arigo and @rlamy about the that possibility that new style type declarations of current CPython could possibly ignore the thing altogether. 

What are "new style type declaration"?

arigo:
@antocuni: that's the new way in CPython to declare C extension types, using PyType_Spec and PyType_Slot instead of a static PyTypeObject

antocuni:
> @antocuni: that's the new way in CPython to declare C extension types, using PyType_Spec and PyType_Slot instead of a static PyTypeObject

Sorry, I'm not following. How is this related to using the correct C type for function pointers?

arigo:
PyType_Slot is basically a ``void *`` pointer that is actually often (but maybe not always) a function pointer.

encukou:
Hi! CPython dev here.
I noticed this issue in CPython when I tried to play around with WebAssembly. And while it's not a super high priority for me, I'd like to stop using `void*` for functions, and I believe it will be possible for `PyType_Slot`.

To not steal this issue, this can be discussed in: https://github.com/encukou/abi3/issues/14 (or in a CPython issue/PR, but I often find smaller forums helpful.)

hodgestar:
@gracinet Is this an issue for using HPy from Rust in practice? If so, could you explain a bit how the problem unfolds?

@encukou Thank you for mentioning the WebAssembly case and it's great that CPython is looking at this too.

I think HPy will probably try sort this pointer issue out to the extent that we can, but it probably won't be right now (although I'll definitely review an updated PR if someone decides to tackle this).

thedrow:
> I think HPy will probably try sort this pointer issue out to the extent that we can, but it probably won't be right now (although I'll definitely review an updated PR if someone decides to tackle this).

In that case, I think it's a good idea to close this PR, create an issue which references it and tackle this later.
Just to keep things neat and tidy.

hodgestar:
I created https://github.com/hpyproject/hpy/issues/174. Closing this for now.

PR #24, opened 2020-03-09
Improve autogen

rlamy:
Reduce duplication between the 2 ABIs by:
* Allowing functions whose name isn't just like CPython's with an 'H' added to be auto-generated.
* Moving other duplicated implementations to `common/implementation.h`

Sadly, dealing with names like `HPy_Foo` required adding another macro. (autogen could only create implementations for `HPyFoo_Bar` so far)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# HPy autogen
+tools/autogen_pypy.txt
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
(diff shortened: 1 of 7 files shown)


rlamy:
`_HPy_IMPL_NAME(_GetAttr)` wouldn't work, it would generate `ctx__GetAttr` (with 2 underscores) for the universal ABI. I don't think it's possible to follow our conventions so far with a single macro.

PR #25, opened 2020-03-15
Feature/number functions

arigo:
Add the simple PyNumber_Xxx() functions


antocuni:
didn't we decide NOT to include these in the main HPy API, but to use more generic `HPy_Add`, etc. instead?

PR #29, opened 2020-04-01
WIP: Document the API design discussion decisions from Leysin 2020

hodgestar:


docs/leysin-2020-design-decisions.md
```diff
@@ -0,0 +1,295 @@
...
+  int size;
+} PointObject;
+
+// When using HPy:
+
+typedef struct {
+    HPyObject_HEAD
+    double x;
+    double y;
+} HPy_Point;
```
paugier, line 214: Just a remark: it seems to me that it's confusing that consumers of the HPy API define variables starting with HPy. Could names starting with HPy be reserved for the HPy API? The equivalent of `PointObject` could be `PointHPy`. At least, it's clear that `PointHPy` is not part of the HPy API but a user defined struct. I don't know if it makes sense?

antocuni: I did a quick search of `PyType_Ready` on github to check how people are currently calling their C types. A small number of extensions use the `Py` prefix, but the majority seems not to use any particular prefix, so +1 to remove the `HPy_` prefix.

timfel:
Should we merge this? Just thinking if people come by and have some of these questions, it's pretty hard to find right now :)

antocuni:
+1 to merge it even if it's not in a perfect state

docs/leysin-2020-design-decisions.md
hodgestar: Agreed.

hodgestar:
Merging & will make a follow up clean up PR.

PR #31, opened 2020-04-07
Write some overview docs

antocuni:


antocuni:
I think that this document is ready to be reviewed. There are still a couple of XXX left (mostly missing links), but I removed the WIP with the hope that someone will start an actual review :)

docs/overview.rst
```diff
@@ -0,0 +1,350 @@
...
+CPython
+  In this mode, HPy is implemented as a set of C macros and ``static inline``
+  functions which translate the HPy API into the CPython API at compile
+  time. The result is a compiled extension which is indistinguishable from a
+  "normal" one and can be distributed using all the standard tools and will
+  run at the very same speed. The ABI tag is defined by the version of CPython
+  which is used to compile it (e.g., ``cpython-37m``),
+
+HPy Universal
+  As the name suggests, the HPy Universal ABI is designed to be loaded and
```
mattip, line 145: Worth mentioning that such packages can be distributed as wheels using the [PEP 384] stable ABI tag `abi3` (see also [the PEP 384 section of PEP 3149](https://www.python.org/dev/peps/pep-3149/#pep-384)? In any case, does this part need a PEP or would this be an import hook that adds that so signature? 

Also: the HPy Hybrid mode would force the "regular" ABI tag `cpython-37m` for the packaged wheels.

Maybe worth adding a short "from the user's perspective" section that would explain this: the CPython mode would not change anything, the HPyUniversal and HPy Hybrid would require an additional install on CPython.

docs/overview.rst
```diff
@@ -0,0 +1,350 @@
...
+    PyPy (using the PyPy `hpy branch <https://foss.heptapod.net/pypy/pypy/tree/branch/hpy>`_
+
+  - it is possible to load HPy Universal extensions on GraalPython (XXX link)
+
+
+However, there is still a long road before HPy is usable for the general
+public. In particular, the following features are on our roadmap but has not
+been implemented yet:
+
+  - it is not possible to write custome types in C. There is already a WIP
```
mattip, line 262: Typo and expanding the idea
```suggestion
  - it is not possible to write custom types (like NumPy's ndarray) in C. There is already a WIP
```

docs/overview.rst
```diff
@@ -0,0 +1,350 @@
...
+    branch to address this issue
+
+  - only a handful of the original Python/C functions has been ported to
```
mattip, line 265:
```suggestion
  - only a handful of the original Python/C functions have been ported to
```

docs/overview.rst
```diff
@@ -0,0 +1,350 @@
...
+    HPy. Porting most of them is straighforward, so for now the priority is to
+    work on the "hard" features to prove that the HPy approach works, and we
+    port new functions as needed
+
+  - the debug mode simply does not exist (yet!)
```
mattip, line 270: Link to a discussion of how to implement? Otherwise it sounds like a pie-in-the-sky idea that may or may not work

docs/overview.rst
```diff
@@ -0,0 +1,350 @@
...
+
+  - there is not a standard/easy way to integrate HPy in a
+    ``distutils/setuptools`` based workflow. The only way to compile HPy
+    extensions right now is to manually close the git repo and tweak your
```
mattip, line 274:
```suggestion
    extensions right now is to manually clone the git repo and tweak your
```

docs/overview.rst
```diff
@@ -0,0 +1,350 @@
...
+    ``setup.py``. Eventually, we plan to offer a workflow which integrates
+    seamlessly with ``pip``, ``setuptools``, etc.
```
mattip, line 276: Worth thinking about [PEP 517](https://www.python.org/dev/peps/pep-0517/) type of build environment (not for this doc, but something to think about)

docs/overview.rst
```diff
@@ -0,0 +1,350 @@
...
+just that some of the people involved contributed to the
+code/design/discussions of HPy:
+
+  - PyPy
+
+  - CPython
+
+  - Cython
+
+  - Numpy (XXX for mattip: do you think it is good/correct to include numpy here?)
```
mattip, line 327: Probably not. I didn't discuss it on any public mailing list nor with the NumPy steering council.

docs/overview.rst
antocuni: > Worth mentioning that such packages can be distributed as wheels using the [PEP 384] stable ABI tag abi3 (see also the PEP 384 section of PEP 3149? In any case, does this part need a PEP or would this be an import hook that adds that so signature?

Actually, at the moment `hpy.universal` on CPython does NOT use the limited API, e.g. we use `PyTuple_GET_ITEM` in `ctx_CallRealFunctionFromTrampoline`. We would need to run benchmarks and so how much it costs to use the limited ABI.

> Maybe worth adding a short "from the user's perspective" section that would explain this: the CPython mode would not change anything, the HPyUniversal and HPy Hybrid would require an additional install on CPython.

good idea

docs/overview.rst
antocuni: We don't have any discussion in a written form AFAIK.
I originally plan to write also a document `debugging.rst` which explains some of the ideas, but I thought it is better to start merging this at first, to allow smaller reviews

docs/overview.rst
antocuni: I didn't know PEP 517. Yes, it's worth thinking about it

docs/overview.rst
```diff
@@ -0,0 +1,350 @@
...
+If you are writing an alternative Python implementation, there is a good
+chance that you already know how painful it is to support the Python/C
+API. HPy is designed to be both faster and easier to implement!
+
+You have two choices:
+
+  - support the Universal ABI: in this case, you just need to export the
+    needed functions and to add a hook to ``dlopen()`` the desired libraries
+
+  - use a custom ABI: in this case, you have to write your own replacement for
```
hodgestar, line 231:
```suggestion
  - use a custom ABI: in this case, you have to write your own replacement for
  + use a custom ABI: in this case, you have to write your own replacement for JUST A TEST
```

docs/overview.rst
antocuni: fixed by 4cbf1ee

docs/overview.rst
antocuni: @mattip I'm going to merge this PR. We can always improve this paragraph later, but at least if we merge this we put online some docs

PR #33, opened 2020-04-20
API docs

antocuni:
I wrote something to document a bit of the API. It's not a reference manual of course, but it should be enough to have a grasp on hpy ideas

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -114,3 +114,23 @@ identity, you can use ``HPy_Is()``::
    indices into a list, which is itself managed by the GC. When an object
    moves, the GC fixes the address into the list, without having to touch all
    the handles which have been passed to C.
+
+
+HPyContext
+-----------
+
+All HPy function calls take a ``HPyContext`` as a first argument, which
+represents the the Python interpreter all the handles belong to.  Strictly
+speaking, it would be possible to design the HPy API without using
+``HPyContext``: after all, all HPy function calls are ultimately mapped to
+Python/C function call, where there is no notion of context.
+
+One of the reasons to include ``HPyContext`` from the day one is to be
+future-proof: it is conceivable to use it to hold the interpreter or the
+thread state in the future, in particular when there will be support for
+sub-interpreter.  Another possible usage could be to embed different versions
+or implementations of Python inside the same process.
+
+Moreover, ``HPyContext`` is used by the HPy Universal ABI to contain a sort of
+virtual function table which is used by the C extensions to call back into the
+Python interpreter.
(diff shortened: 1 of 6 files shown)


docs/api.rst
```diff
@@ -0,0 +1,282 @@
...
+
+The concept of handles is certainly not unique to HPy. Other examples include
+Unix file descriptors, where you have ``dup()`` and ``close()``, and Windows'
+``HANDLE``, where you have ``DuplicateHandle()`` and ``CloseHandle()``.
+
+
+Handles vs ``PyObject *``
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. XXX I don't like this sentence, but I can't come up with anything better
```
ammaraskar, line 30: It reads fine to me :)

I'd maybe phrase it as:

> The biggest difference is that in the old Python/C API, multiple ``PyObject
> *`` references to the same objects are completely equivalent to each other. They can be passed to Python/C API functions interchangeably _including_ memory management APIs.
> 
> In particular, it does not matter which particular reference you call ``Py_INCREF`` and ``Py_DECREF`` on, as long as the total number of increfs and decrefs to the underlying object is the same at the end of the object lifetime.

docs/api.rst
```diff
@@ -0,0 +1,282 @@
...
+      HPy x = HPyLong_FromLong(ctx, 42);
+      HPy y = HPy_Dup(ctx, x);
+      /* ... */
+      // we need to close x and y independently
+      HPy_Close(ctx, x);
+      HPy_Close(ctx, y);
+  }
+
+Calling ``HPy_Close()`` on the same handle twice is an error.  Forgetting to
+call ``HPy_Close()`` on a handle results in a memory leak.  When running in
```
ammaraskar, line 64: What about operating on a `HPy_Close()`'d handle?

docs/api.rst
```diff
@@ -0,0 +1,282 @@
...
+:ref:`debug mode`, HPy actively checks that you that you don't close a handle
+twice and that you don't forget to close any.
+
+
+.. note::
+  The debug mode is a good example of how powerful it is to decouple the
+  lifetime of handles and the lifetime of an objects.  If you find a memory
+  leak on CPython, you know that you are missing a ``Py_DECREF`` somewhere but
```
ammaraskar, line 72: If you have the willingness, it might be worth mentioning the other case as well. That is, you have an extra `Py_DECREF` somewhere and then later in an unrelated part of the code you end up calling `Py_DECREF` *after* the object has been free'd.

docs/api.rst
antocuni: thank you, fixed by 278ffa4

docs/api.rst
antocuni: fixed by 278ffa4

docs/api.rst
antocuni: I tried to mention it in ca3b3af, do you think it's good enough?

PR #38, opened 2020-05-09
Add link to documentation in README

ammaraskar:
Should make the compiled sphinx docs a bit more discoverable + show their build status


antocuni:
good idea, thanks!

PR #42, opened 2020-06-08
Refactor methods and slots

antocuni:
This PR refactors how method and trampolines are defined, as discussed on the mailing list.
Current status:
- universal tests works; cpython-mode tests fail (will fix them later)
- methods and slots are represented by the new `HPyMeth` struct, which explicitly contains the impl and the trampoline function pointers:
```c
typedef struct {
    int slot;
    void *impl;                     // function pointer to the implementation
    void *cpy_trampoline;           // used by CPython to call impl
    HPyMeth_Signature signature;    // Indicates impl's expected the signature

    // the following fields are used only if slot == HPy_meth
    const char *name;               // The name of the built-in function/method
    const char *doc;                // The __doc__ attribute, or NULL
} HPyMeth;
```
- `.slot` will be `HPy_meth` for normal methods and things like `HPy_tp_repr`, etc. for slots
- `.signature` comprises both the "standard" method signatures and the slot ones. So it can have values like `HPyMeth_O`, `HPyMeth_NOARGS`, but will also contain things like `HPyMeth_NEW`, `HPyMeth_RICHCMP`, `HPyMeth_UNARY`, etc.
- There is a bit of redundancy between `.slot` and `.signature`, because most slots support only one specific signature. One alternative is to specify only `.slot` and have some logic to deduce the signature automatically. In that case, instead of a single `HPy_meth` value for the slot, we will need `HPy_meth_o`, `HPy_meth_noargs`, etc.

The preferred way to define a method is to use a macro, but you can also define the `HPyMeth` struct explicitly:
```c
HPyMeth_DEFINE(f, "f", f_impl, HPyMeth_NOARGS)
static HPy f_impl(HPyContext ctx, HPy self) { ... }
```
which roughly expands to:
```c
static HPy f_impl(HPyContext ctx, HPy self); // prototype
HPyMeth_TRAMPOLINE(f_trampoline, f_impl, HPyMeth_NOARGS)
HPyMeth f = {
    .slot = HPy_meth,
    .impl = f_impl,
    .cpy_trampoline = f_trampoline,
    .signature = HPyMeth_NOARGS,
    .name = "f"
};
```
The idea is that the `HPyMeth` struct will be fully documented and it will be possible/supported to write them by hand (although you still need to manually call `HPyMeth_TRAMPOLINE` to generate the cpy trampoline). However, we will also provide `HPyMeth_DEFINE` to make the job easier; moreover, since it also declares a prototype for `f_impl`, you get a nice compiler error if by mistake you specify the wrong flag.

One defined, you attach methods to a module in this way (note that you no longer need to specify the various `ml_*` fields in the `moduledef`: they are already specified in the `HPyMeth` struct):
```c
/* in this case, f and g are defined in another C file */
extern HPyMeth f;
extern HPyMeth g;
    .m_name = "mytest",
    .m_doc = "some test for hpy",
    .m_size = -1,
    .methods = { &f, &g, NULL }
};
```
One big advantage over the old solution is that now `HPyMeth` contains explicit pointers to both `impl` and `cpy_trampoline`, instead of using the ugly hack to have a function which returns them when called.

In case you are porting an existing extension, you can easily use your old Python/C methods by using `.legacy_methods`:
```c
#include <Python.h>
PyMethodDef old_methods[] = {
    {"a", (PyCFunction)a, METH_NOARGS},
    {"b", (PyCFunction)b METH_O},
};
    .m_name = "mytest",
    .m_doc = "some test for hpy",
    .m_size = -1,
    .legacy_methods = old_methods,
    .methods = { &f, &g, NULL }
};
```

`HPyType_FromSpec` works similarly; the old `tp_*` slots are represented as `HPyMeth` as well; Again, you can define it manually or use `HPyMeth_SLOT`.  Slots and normal methods are listed together inside `HPyType_Spec.methods[]`, and inside `hpy.universal` there is some logic to "filter out" the `HPy_meth` and put them into an appropriate `PyMethodDef[]` for CPython:
```c
....
HPyMeth_SLOT(Point_new, HPy_tp_new, Point_new_impl, HPyMeth_KEYWORDS);
HPyMeth_DEFINE(Point_foo, "foo", Point_foo_impl, HPyMeth_NOARGS);
static HPyType_Spec Point_spec = {
    .name = "mytest.Point",
    .basicsize = sizeof(PointObject),
    .methods = {
        &Point_new,
        &Point_foo,
        NULL
    }
};
```

The plan is to also have `HPyType_Spec.legacy_slots` which contains the old "cpython-style" slots, to be used when migrating an existing extension piece by piece. It's not implemented yet, though.

There are some more open questions:
- CPython makes a distinction between some function types which are essentially the same; e.g. `reprfunc` is the same as `unaryfunc` which is the same as `METH_NOARGS`. Do we want to keep the distinction and give each one its own value inside the `HPyMeth_Signature` enum, or we want to simplify and collapse all of them into a single value?

- for each different signature, we need a macro to generate a forward-declaration of the "impl" function, and a macro to generate a trampoline. I am considering to autogenerate all of them.

- Do we want to keep the distinction between `.slot` and `.signature` or we want the signature to be automatically deduced by the slot?


antocuni:
I think that the branch is ready to be merged.
It contains many internal changes, which are in part explained by my first message and in part  were discussed on the ML, on IRC.
Internally, the biggest changes are:
1. Create a generic macro-based internal API to declare impl functions and generate trampolines: see `HPyFunc_Signature`, `HPyFunc_DECLARE` and `HPyFunc_TRAMPOLINE`

2. Use the universal `HPyModuleDef` also in the CPython-ABI case; this means that it is no longer an alias for `PyModuleDef`, and that we need to share the code between universal and cpython, as we already do for `HPyType_Spec`. See `hpy/devel/src/runtime/ctx_module.c`, which is shared between the two.

From the user-API point of view, the biggest change is to unify methods and slots into the new `HPyDef` API. The easiest way to explain the change is to look at commit 325ac1c, which refactors `pof.c` to use the new API.

arigo:
(Sorry, don't wait for me and go forward here)

timfel:
This looks very good from my PoV (I would love to get rid of slots vs methods altogether, as discussed on IRC, but alas....) . I asked @fangerer to also take a look, but don't wait if you want to go ahead

antocuni:
> I would love to get rid of slots vs methods altogether, as discussed on IRC, but alas....

I would like that as well. In theory, once we settle down the `nb_add` vs `sq_concat` issue, we could simply list slots as methods whose names are `__add__`, `__getitem__`, etc.
However, the biggest problems to go down that route are:
1. special methods/slots have a different C signature than normal methods, in CPython,
2. CPython makes a difference between e.g. `tp_getattr` and `tp_getattro`, which does not exist at the Python-level: again

These two points makes it harder (impossible?) to unify the two concepts of "normal methods" and "special methods/slots" at the C level, IMHO.

mattip:
NumPy is adopting the approach that assigning to both `nb_add` and `sq_concat` is an error. I think the same thing can be said for any other construct like in (2), and we should feel free to raise an issue on CPython. @vstinner thoughts?

PR #60, opened 2020-08-09
Setuptools support

antocuni:
Factor out the logic to build extensions from `support.py`, and add an official setuptools way to build hpy extensions.

Once this branch is merged, we should also merge the corresponding branch in PyPy, which is `hpy-setuptools-extension`


antocuni:
(this branch also renames `listbuilder.c` into `ctx_listbuilder.c`, to avoid having to special case this file inside `hpy.devel`

PR #64, opened 2020-08-19
HPyDef_GET, HPyDef_GETSET

arigo:
Implement getsets (C-level properties)

### DIFF
--- a/hpy/devel/include/common/hpydef.h
+++ b/hpy/devel/include/common/hpydef.h
@@ -126,17 +126,34 @@ typedef struct {
         }                                           \
     };
 
-#define HPyDef_GET(SYM, NAME, IMPL, ...)                        \
-    static HPy IMPL(HPyContext ctx, HPy self, void *closure);   \
-    HPyFunc_TRAMPOLINE(SYM##_trampoline, IMPL, HPyFunc_GETTER); \
-    HPyDef SYM = {                                              \
-        .kind = HPyDef_Kind_GetSet,                             \
-        .getset = {                                             \
-            .name = NAME,                                       \
-            .getter_impl = IMPL,                                \
-            .getter_cpy_trampoline = SYM##_trampoline,          \
-            __VA_ARGS__                                         \
-        }                                                       \
+#define HPyDef_GET(SYM, NAME, IMPL, ...)                            \
+    static HPy IMPL(HPyContext ctx, HPy self, void *closure);       \
+    HPyFunc_TRAMPOLINE(SYM##_get_trampoline, IMPL, HPyFunc_GETTER); \
+    HPyDef SYM = {                                                  \
+        .kind = HPyDef_Kind_GetSet,                                 \
+        .getset = {                                                 \
+            .name = NAME,                                           \
+            .getter_impl = IMPL,                                    \
+            .getter_cpy_trampoline = SYM##_get_trampoline,          \
+            __VA_ARGS__                                             \
+        }                                                           \
+    };
+
+#define HPyDef_GETSET(SYM, NAME, GETIMPL, SETIMPL, ...)                     \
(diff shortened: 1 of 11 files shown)


hpy/devel/include/common/hpydef.h
```diff
@@ -126,17 +126,34 @@ typedef struct {
...
+        .kind = HPyDef_Kind_GetSet,                                 \
+        .getset = {                                                 \
+            .name = NAME,                                           \
+            .getter_impl = IMPL,                                    \
+            .getter_cpy_trampoline = SYM##_get_trampoline,          \
+            __VA_ARGS__                                             \
+        }                                                           \
+    };
+
+#define HPyDef_GETSET(SYM, NAME, GETIMPL, SETIMPL, ...)                     \
```
antocuni, line 142: 1. do we need two separate macros `GET` and `GETSET`? Can't we just have a single `GETSET` and pass `NULL` to setter if we don't have? I suppose this would make things more complicate because we would need to ensure we do the correct thing in `HPyFunc_DECLARE` and `HPyFunc_TRAMPOLINE`, which brings me to the next question:

2. do we want to support the case in which we provide a setter but not a getter?

hpy/devel/include/common/hpydef.h
```diff
@@ -126,17 +126,34 @@ typedef struct {
...
-    HPyDef SYM = {                                              \
-        .kind = HPyDef_Kind_GetSet,                             \
-        .getset = {                                             \
-            .name = NAME,                                       \
-            .getter_impl = IMPL,                                \
-            .getter_cpy_trampoline = SYM##_trampoline,          \
-            __VA_ARGS__                                         \
-        }                                                       \
+#define HPyDef_GET(SYM, NAME, IMPL, ...)                            \
+    static HPy IMPL(HPyContext ctx, HPy self, void *closure);       \
```
antocuni, line 130: why didn't you use `HPyFunc_DECLARE` here?

hpy/devel/include/common/hpydef.h
```diff
@@ -126,17 +126,34 @@ typedef struct {
...
+            .name = NAME,                                           \
+            .getter_impl = IMPL,                                    \
+            .getter_cpy_trampoline = SYM##_get_trampoline,          \
+            __VA_ARGS__                                             \
+        }                                                           \
+    };
+
+#define HPyDef_GETSET(SYM, NAME, GETIMPL, SETIMPL, ...)                     \
+    static HPy GETIMPL(HPyContext ctx, HPy self, void *closure);            \
+    static int SETIMPL(HPyContext ctx, HPy self, HPy value, void *closure); \
```
antocuni, line 144: same as above, `HPyFunc_DECLARE` ?

antocuni (changes requested):
Looks good to me, apart the two minor things I commented

hpy/devel/include/common/hpydef.h
arigo: Ah, right.  I wrote these macros before I realized I needed `HPyFunc_GETTER` and `HPyFunc_SETTER` anyway.  Will fix

hpy/devel/include/common/hpydef.h
arigo: 1. Yes, it's more complicated to do the right thing and have the `static` function declaration present only if the user didn't give exactly `NULL` as argument to the macro.  I think it's quite a mess to do it, even, and I thought it would go too far on the side of macro magic.

2. Right, maybe?  It doesn't cost a lot to have `HPyDef_SET` in the same style.

antocuni: Agreed, then I vote for adding `HPyDef_SET` and be happy :)

antocuni (approved):

PR #65, opened 2020-08-24
Improve autogen

antocuni:


antocuni:
This PR does 3 things:
- tweak a bit the code, to allow the pypy repo to run its own autogen
- use `.gitattributes` to teach gihub about which files are autogenerated, which brinds some UI niceties (the most important being that they are hidden by default in diffs)
- add a test which checks that we run `make autogen` after we change something

PR #66, opened 2020-08-24
Improve slot tests and HPyDef_SLOT

antocuni:
The goal of this branch is to simplify writing new tests for custom types, and write tests for each slot.

At the moment the new tests are still being written, but reviews are already welcome for the first commits, which introduce the new `ExtensionTemplate` logic to remove some boilerplate.


antocuni:
Contrarily to the branch name, this PR does NOT introduce any new test about slots. Instead:
1. refactor the existing slot tests to make it easier to write new ones
2. change `HPyDef_SLOT`, now it no longer requires to manually specify the signature of the slot
3. autogen the `HPySlot_Slot` and the boilerplate needed to implement (2)

PR #68, opened 2020-09-01
HPy_FatalError(), a bit of a mess

arigo:


hpy/devel/include/universal/hpy.h
```diff
@@ -71,5 +82,11 @@ static inline HPy _HPy_New(HPyContext ctx, HPy h_type, void **data) {
     return h;
 }
 
+static inline _HPy_NO_RETURN void
+HPy_FatalError(HPyContext ctx, const char *message) {
+    ctx->ctx_FatalError(ctx, message);
+    abort();
```
antocuni, line 88: is there any particular reason why you put the `abort()` here instead of inside `ctx_FatalError`?

test/test_hpyerr.py
```diff
@@ -15,3 +15,16 @@ def test_NoMemory(self):
...
+            HPyDef_METH(f, "f", f_impl, HPyFunc_NOARGS)
+            static HPy f_impl(HPyContext ctx, HPy self)
+            {
+                HPy_FatalError(ctx, "boom!");
+            }
+            @EXPORT(f)
+            @INIT
+        """)
+        # Calling mod.f() gives a fatal error, ending in abort().
+        # How to check that?  For now we just check that the above compiles
```
antocuni, line 30: I fear the only solution is to write some code in a tmp file, spawn a subprocess and check the output/return code. It will be messy to test this on PyPy, though.

antocuni (approved):

hpy/devel/include/universal/hpy.h
arigo: Yes.  It's needed for gcc to figure out that the call above never returns.  The `abort()` should never be executed.  Note that it seems to be needed even though we say `_HPy_NO_RETURN` above, I think because it's an inlined function and during inlining it forgets this detail.  I'll write a comment in the test.

test/test_hpyerr.py
arigo: Yes, and I think for now it's fine like it is

hpy/devel/include/universal/hpy.h
antocuni: I think it would be good to have a comment also here in the code

antocuni:
LGTM

PR #70, opened 2020-09-01
add the HPyTupleBuilder API

arigo:
Also tweak a little bit the type 'HPyListBuilder'.

Done as a PR to the existing 'more-slot-tests-2' branch, to reduce the risk of conflicts with that branch.

### DIFF
--- a/hpy/devel/include/universal/hpy.h
+++ b/hpy/devel/include/universal/hpy.h
@@ -18,7 +18,8 @@ typedef intptr_t HPy_ssize_t;
 typedef intptr_t HPy_hash_t;
 struct _HPy_s { HPy_ssize_t _i; };
 typedef struct _HPy_s HPy;
-typedef struct { HPy _lst; } HPyListBuilder;
+typedef struct { HPy_ssize_t _lst; } HPyListBuilder;
+typedef struct { HPy_ssize_t _tup; } HPyTupleBuilder;
 typedef struct _HPyContext_s *HPyContext;
 
 /* compatibility CPython types */
--- a/hpy/devel/include/universal/hpy.h
+++ b/hpy/devel/include/universal/hpy.h
@@ -16,10 +16,15 @@
 /* HPy types */
 typedef intptr_t HPy_ssize_t;
 typedef intptr_t HPy_hash_t;
-struct _HPy_s { HPy_ssize_t _i; };
-typedef struct _HPy_s HPy;
+
+/* The following types, when compiled with the present universal-mode header,
+   are each just a pointer-sized integer.  What this integer (or pointer)
+   means is up to the implementation.  For example, on both CPython and PyPy,
+   the HPy structure contains an index in a global array. */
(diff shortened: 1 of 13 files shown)


hpy/devel/include/universal/hpy.h
```diff
@@ -18,7 +18,8 @@ typedef intptr_t HPy_ssize_t;
 typedef intptr_t HPy_hash_t;
 struct _HPy_s { HPy_ssize_t _i; };
 typedef struct _HPy_s HPy;
-typedef struct { HPy _lst; } HPyListBuilder;
+typedef struct { HPy_ssize_t _lst; } HPyListBuilder;
```
antocuni, line 21: maybe add a comment explaining that `_lst` is just an opaque pointer-like field which is implementation-specific. Else people might think that it contains the length or something like that.

hpy/devel/src/runtime/ctx_tuplebuilder.c
```diff
@@ -0,0 +1,53 @@
...
+#include "hpy.h"
+
+#ifdef HPY_UNIVERSAL_ABI
+   // for _h2py and _py2h
+#  include "handles.h"
+#endif
+
+
+_HPy_HIDDEN HPyTupleBuilder
+ctx_TupleBuilder_New(HPyContext ctx, HPy_ssize_t initial_size)
```
antocuni, line 12: should it be just `size`? It's not that it can change anyway

hpy/devel/src/runtime/ctx_tuplebuilder.c
```diff
@@ -0,0 +1,53 @@
...
+{
+    PyObject *tup = PyTuple_New(initial_size);
+    if (tup == NULL)
+        PyErr_Clear();   /* delay the MemoryError */
```
antocuni, line 16: why do you want to delay it?

antocuni (changes requested):

hpy/devel/src/runtime/ctx_tuplebuilder.c
arigo: Right now yes, but the C API gives `_PyTuple_Resize()`.  You can resize tuples while you build them (and you're not supposed to use that after the tuple is exposed to Python, of course).

hpy/devel/src/runtime/ctx_tuplebuilder.c
arigo: It's like in ListBuilder.  I'll write a comment.

hpy/devel/src/runtime/ctx_tuplebuilder.c
antocuni: ah, I didn't know. The question is whether we want to expose this functionality in `TupleBuilder`. But I am fine for not having to discuss it now and leave the argument name as it is

antocuni (approved):
LGTM

PR #71, opened 2020-09-01
Py27 compatibility

antocuni:
Add a test which checks that some file are importable on python2.7, needed by pypy tests

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -113,5 +113,6 @@ jobs:
     displayName: 'Use Python 2.7'
 
   - script: |
+      pip install pathlib
       python test/check_py27_compat.py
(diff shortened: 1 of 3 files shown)


rlamy:
TBH, I don't like having so much py2-compatibility hacks, but this should reduce a lot the risk of accidental pypy breakage, so +1.

hodgestar (approved):
Looks good!

Would it be possible to extent this later to also check that there are no imports outside of test functions that PyPy tests would be unhappy with?

antocuni:
> TBH, I don't like having so much py2-compatibility hacks, but this should reduce a lot the risk of accidental pypy breakage, so +1.

I also don't like these hacks, but I can't think of a better solution :(

PR #74, opened 2020-09-03
Give HPyType_FromSpec() an extensible way to list extra HPy objects

arigo:


hpy/devel/include/common/hpytype.h
```diff
@@ -31,6 +31,18 @@ typedef struct {
...
+typedef enum {
+    HPyType_SpecParam_Base = 1,
+    HPyType_SpecParam_Bases = 2,
+    //HPyType_SpecParam_Metaclass = 3,
+    //HPyType_SpecParam_Module = 4,
+} HPyType_SpecParam_Kind;
+
+typedef struct {
+    HPyType_SpecParam_Kind kind;
+    HPy object;
```
antocuni, line 43: what about using `void *` instead of `HPy`?
I can't think of any reason to use something different than `HPy` right now, but why close the door for the future?

hpy/tools/autogen/public_api.h
```diff
@@ -90,7 +91,8 @@ HPy HPyErr_NoMemory(HPyContext ctx);
 
 /* object.h */
 int HPy_IsTrue(HPyContext ctx, HPy h);
-HPy HPyType_FromSpec(HPyContext ctx, HPyType_Spec *spec);
+HPy HPyType_FromSpec(HPyContext ctx, HPyType_Spec *spec,
+                     HPyType_SpecParam *optional_object_parameters);
```
antocuni, line 95: I'm not sure about the name `optional_object_parameters`:
  - `object`: see above the my comment for `void *` vs `HPy`: if we choose `void *`, they are not necessarily objects. If we choose `HPy`, it's "obvious" they are objects

 - `optional`: there are tons of params which accept `NULL`, we never put `optional` in the name

Maybe `params` is enough?

hpy/devel/include/common/hpytype.h
```diff
@@ -33,7 +33,7 @@ typedef struct {
 
 typedef enum {
     HPyType_SpecParam_Base = 1,
-    HPyType_SpecParam_Bases = 2,
+    HPyType_SpecParam_BasesTuple = 2,
```
antocuni, line 36: do we really need/want `BasesTuple`? Can't we just use `Base`?

test/test_hpytype.py
```diff
@@ -288,3 +288,62 @@ def test_HPyDef_SET(self):
...
+
+    def test_specparam_basestuple(self):
+        mod = self.make_module("""
+            static HPyType_Spec Dummy_spec = {
+                .name = "mytest.Dummy",
+                .itemsize = 0,
+                .flags = HPy_TPFLAGS_DEFAULT | HPy_TPFLAGS_BASETYPE,
+            };
+
+            @INITCODE
```
antocuni, line 328: personally, I find this `@INITCODE` usage very confusing: by reading the test it's not immediately clear what's happening, and seeing C code outside a function is ugly.
Some random suggestion for alternatives:

1. Don't use `@EXPORT_TYPE` at all; instead, we use `@EXTRA_INIT_FUNC` which takes a **function name** and calls it in the init; inside the function, we manually create the type and put it in the dict:
```C
void create_Dummy(HPyContext ctx, HPy mod) {
    HPyType_SpecParam params[] = { ... };
    HPy h_dummy = HPyType_FromSpec(..., params);
    HPy_SetAttr(ctx, mod, HPyUnicode_FromString(ctx, "Dummy", h_dummy);
}

@EXTRA_INIT_FUNC(create_Dummy)
@INIT // inside INIT, we cal all the EXTRA_INIT_FUNC registered above
```

2. use`@EXPORT_TYPE`, but in the 3rd argument pass a C function which *RETURNS* the `HPyType_SpecParam[]` array. This is probably more messy because you need to malloc/free the array, or to return a static which you fill at runtime, etc.

antocuni (changes requested):

hpy/devel/include/common/hpytype.h
arigo: And use ``HPy_AsVoidP()``?  That seems strange to require this API, which is advanced and maybe not really supported on all implementations, just for that.

hpy/devel/include/common/hpytype.h
arigo: I don't see a problem to support ``BasesTuple`` and it's closer to what is done in the CPython C API.  It avoids a potentially messy deconstruction of the tuple of bases in user code if he starts with that.

test/test_hpytype.py
arigo: "by reading the test it's not immediately clear what's happening": well, I'd argue that it's already not immediately clear what's happening, with all the other ``@`` custom-pseudo-macros.  But yes, your suggestion number 1 looks at least like progress.

hpy/devel/include/common/hpytype.h
antocuni: Right, I didn't think that we can't cast `HPy` to `void *` directly. Then ok, I am fine with leaving it as it is.

antocuni (approved):
LGTM

PR #75, opened 2020-09-04
Fix closing-vs-borrowed issues in HPyArg_ParseKeywords()

arigo:
- get rid of HPyDict_GetItem and HPyDict_SetItem.  The first one was returning a "borrowed handle"

- HPyArg_ParseKeywords() cannot accept the "O" format character because the handle to return might be obtained by calling HPy_GetItem() in the keywords dictionary

- as a replacement, add the "N" character, which returns a new handle


hpy/devel/src/runtime/argparse.c
```diff
@@ -39,45 +39,82 @@ int _HPyArg_ParseItem(HPyContext ctx, HPy current_arg, const char **fmt, va_list
         *output = current_arg;
         break;
     }
+    case 'N': {
```
antocuni, line 42: sorry for not having thought of this before: I think that `N` is a bad choice: in `struct.pack` it is used for `size_t`, as opposed to `n` which is `ssize_t` (and PyArg_ParseTuple tries to be consistent with `struct.pack`, apparently). I can't think of any nice char to use, though; random ideas:
- `A` for hAndle
- `J` for obJect

hpy/devel/src/runtime/argparse.c
```diff
@@ -88,81 +125,114 @@ HPyArg_Parse(HPyContext ctx, HPy *args, HPy_ssize_t nargs, const char *fmt, ...)
 
     va_end(vl);
     return 1;
+
+  error:
+    va_end(vl);
+
+    va_start(vl, fmt);
+    clear_already_allocated_handles(ctx, fmt, fmt_end, &vl);
```
antocuni, line 133: what about `close_already_allocated_handles`?

hpy/devel/src/runtime/argparse.c
```diff
@@ -39,45 +39,82 @@ int _HPyArg_ParseItem(HPyContext ctx, HPy current_arg, const char **fmt, va_list
...
 
+static void
+skip_next_format_character(HPyContext ctx, const char **fmt, va_list *vl)
+{
+    if (!parse_item(ctx, HPy_NULL, fmt, vl))
+        HPy_FatalError(ctx, "argparse.c: skip_next_format_character failed");
+}
+
+static void
+clear_already_allocated_handles(HPyContext ctx, const char *fmt,
```
antocuni, line 65: it would be nice to have a test which check that this has been called, although it's probably hard to write. Maybe once we have a proper debug mode, it will be easier?

caseyc37:
Tracked down the issue with `pytest test` to the two functions test_SetItem and test_GetItem in test_hpydict.py.

These two functions attempt to call the functions HPyDict_SetItem and HPyDict_GetItem, testing the return values of these two functions. Now, this branch has removed both of these functions (as specified in the original comment); hence, both tests will fail, due to function not existing.

Given this, would it make sense to simply delete those two tests from the test file?

hodgestar:
@caseyc37 Yes, deleting those tests makes sense!

I'm also planning to do a rewrite of the argument parsing to include some of the ideas from this PR though. The argument parsing grew organically, and now it is time to implement it cleanly with all the functionality it needs.

hpy/devel/src/runtime/argparse.c
hodgestar: I'm going to put in a vote for `N` anyway. `A` and `J` seem fairly horrible, where `N` has a nice interpretation as `New`. Staying in sync with `struct.pack` is a good overall goal, but the formats here already deviate from struct.pack quite a bit anyway. Do we have a need for `HPy_size_t` as opposed to `HPy_ssize_t`?

antocuni: that's true, I notice that for example `p` is a Pascal String for struct and a bool in `PyArg_ParseTuple`.
However, I still think that `N` should be our last choice until we come up with something better. Another reason for **not** to chose `N` is that in `PyArg_ParseTuple` we already have `n` which is `Py_ssize_t`.

Other random ideas:
- `X`: no particular meaning but it seems to be free
- `O`: we can't support the current meaning of `O` anyway, so we might want to just change its semantics. I'm -0.5 on this though, because it would lead to segfaults if the user forget to update the types of the pointers it passes
- `O^`: we already have the modifiers to tweak the meaning of `O`, so we can introduce one to say "it's a object, please return a handle". We can also use a different punctuation  sign of course, e.g.: `O%`, `O'`, `O+`, `O/`, `O.`, `O*`. `O@`, `O~`, ...
- just a punctuation mark without a letter; e.g. `@`

hodgestar: Note that `O` is still supported and useful, but only for `HPyArg_Parse`. I'm also -1 on giving `O` different meanings between the `HPyArg_Parse` and `HPyArg_ParseKeywords`.

antocuni: > Note that O is still supported and useful, but only for HPyArg_Parse. I'm also -1 on giving O different meanings between the HPyArg_Parse and HPyArg_ParseKeywords.

that's true, I didn't think of that. Then the idea of adding a modifier to `O` looks even better to me:

  * `O` returns an existing handle (which you don't need to close), and it's not supported by `HPyArg_ParseKeywords`
  * `O^` (or equivalent) returns a new handle which you need to close.

So, what is the best punctuation mark to say "dup the O"? I would like `O²` but it's not ASCII 😅. `O'` is nice because it has some kind of mathematical sense (it's a "new O"), but I worry it's too easy not to notice it.

@arigo do you have any opinion?

arigo: Maybe ``O+``?  But ideally something that CPython isn't likely to add in the next version with an unrelated meaning, and I have no idea how likely they are to pick ``O+``.

antocuni: I agree we should try to avoid clashing with the future CPython.
I think it is unlikely to happen, though. I quickly looked through [git blame Doc/c-api/arg.rst](https://github.com/python/cpython/blame/master/Doc/c-api/arg.rst) and it seems that the format arguments are added at a very slow pace, the last many years ago.

+1 for `O+`

hodgestar: I've implemented `O+` in #111, but it's ugly for something that will be used everywhere. Specifically, having to use a `+` modifier to achieve the most basic format code is not a great solution.

What about making `O` always dup the handle? Then at least switching from `HPyArg_Parse` to `HPyArg_ParseKeywords` doesn't require one to also change how handles are closed.

antocuni: > What about making O always dup the handle?

it might be a good idea.

However, before we go on: I am a bit confused. Didn't we introduce `HPyTracker` specifically to have a way to close at once all handles opened by `HPyArg_ParseKeywords`? I mean, something like that:

```c
HPyTracker t = HPyArg_ParseKeywords(..., "OO", ..., &h1, &h2);
HPyTracker_Close(ctx, t);
```

If we do so, it's easy to implement `O`: if it comes from the positional args, we don't dup it. If it comes from the keywords dict, we dup and add it to the tracker.

Am I missing anything?

arigo: Uh, I think you're right.

I would make sure that the syntax ``if (!HPyArg_ParseKeywords(...))`` is not allowed, which should be the case if ``HPyTracker`` is a struct type (and not a pointer type), to prevent a naïve translation from the CPython API which appears to work but leaks.

Ideally the tracker could be extended later to also close the PyBuffer equivalent, or in pypy free a ``const char *`` version of a Python unicode string, for example.

antocuni: > I would make sure that the syntax if (!HPyArg_ParseKeywords(...)) is not allowed, which should be the case if HPyTracker is a struct type (and not a pointer type), to prevent a naïve translation from the CPython API which appears to work but leaks.

Agreed. We also have the option of continuing returning `int` and pass an output argument to hold the tracker. And also, we need to decide what to do with `HPyArg_Parse`. I think we have three options:

1. `HPyArg_Parse` returns `int` and `HPyArg_ParseKeywords` returns `HPyTracker`
2. they both return `HPyTracker` (but it will be unused  for `HPyArg_Parse`)
3. they both return `int` and we return the tracker via an output arg. This has also the advantage to avoid allocating the tracker if we happen not to need it.

hodgestar: I hadn't thought about returning the HPyTracker (I was thinking only of using it internally) but that seems like an interesting idea.

hodgestar:
I've replaced the fixes to argument parsing with #115 and the removal of HPyDict_GetItem/_SetItem with #116 so I'm going to close this PR for now. Please shout if I missed anything from here.

PR #72, opened 2020-09-02
Check the source code using CPPCheck

thedrow:
Since this project is written in C, we should use a linter to ensure we don't have issues which will bite us later.
This PR runs cppcheck on the source code and reports the errors back.

We could invest time in the tool's configuration to enforce our own rules, if we so desire.

PR #76, opened 2020-09-06
Replace alloca() with variable length arrays

thedrow:
This fixes the warning in #72 regarding the usage of alloca.
Merging this PR means that we aim to support C99 and above, which is in my opinion a reasonable assumption these days.

### DIFF
--- a/hpy/universal/src/misc_win32.h
+++ b/hpy/universal/src/misc_win32.h
@@ -28,8 +28,6 @@ static void *dlsym(void *handle, const char *symbol)
         */
         int i;
         char mangled_name[1 + strlen(symbol) + 1 + 3 + 1];
-        if (!mangled_name)
-            return NULL;
         for (i = 0; i < 32; i++) {
             sprintf(mangled_name, "_%s@%d", symbol, i * 4);
             address = GetProcAddress((HMODULE)handle, mangled_name);
--- a/hpy/universal/src/misc_win32.h
+++ b/hpy/universal/src/misc_win32.h
@@ -1,6 +1,3 @@
-#include <malloc.h>   /* for alloca() */
-
-
 /************************************************************/
 /* Emulate dlopen()&co. from the Windows API */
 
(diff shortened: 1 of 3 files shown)


thedrow:
If you'd like to keep support for older versions of C, we could do:
```c
#if __STDC_VERSION__ >= 199901L
#define alloca(array_name, array_type, array_size) array_type array_name[array_size];
#else
#define alloca(array_name, array_type, array_size) array_type *array_name = alloca(array_size);
#endif
```

But that seems more complicated than dropping support for very old versions of C.
It's your call.

antocuni:
we have already decided to target C99 (although this is probably not explicitly written anywhere), so this should be a safe.
@arigo what do you think?

antocuni (approved):

hpy/universal/src/misc_win32.h
```diff
@@ -30,7 +27,7 @@ static void *dlsym(void *handle, const char *symbol)
            for __stdcall functions.
         */
         int i;
-        char *mangled_name = alloca(1 + strlen(symbol) + 1 + 3 + 1);
+        char mangled_name[1 + strlen(symbol) + 1 + 3 + 1];
         if (!mangled_name)
             return NULL;
```
thedrow, line 32:
```suggestion
```

We should remove this check too IMO.

arigo:
I don't see what is wrong with using an old functionality which works, even when a newer but equivalent one is available.  But I also won't oppose merging this PR anyway.

antocuni:
> I don't see what is wrong with using an old functionality which works, even when a newer but equivalent one is available. But I also won't oppose merging this PR anyway.
I think the point is that `alloca` is not standard, while variable length arrays are.

thedrow:
> > I don't see what is wrong with using an old functionality which works, even when a newer but equivalent one is available. But I also won't oppose merging this PR anyway.
> 
> I think the point is that `alloca` is not standard, while variable length arrays are.

It is also a source of very obscure bugs as [StackOverflow](https://stackoverflow.com/questions/1018853/why-is-the-use-of-alloca-not-considered-good-practice) demonstrates.

antocuni:
It looks like variable lengths arrays are not supported by visual c++, so I fear we need to rever this PR. See this comment for more details: https://github.com/hpyproject/hpy/pull/125#discussion_r542044386

PR #77, opened 2020-09-06
Use pytest-xdist to run tests

thedrow:
Using xdist cuts the build time significantlly.
I see no reason why we shouldn't use it.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -22,7 +22,7 @@ jobs:
     displayName: 'Use Python $(python.version)'
 
   - script: |
-      python -m pip install --upgrade pip
+      python -m pip install --upgrade pip wheel
     displayName: 'Install dependencies'
 
   - script: |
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -22,16 +22,16 @@ jobs:
...
     displayName: 'Install dependencies'
 
   - script: |
       python -m pip install .
     displayName: 'Build project'
 
   - script: |
-      pip install pytest pytest-azurepipelines
-      pytest test/
+      pip install pytest pytest-azurepipelines pytest-xdist
```
antocuni, line 33: shouln't it be `pytest-xdist[psutil]` to enable the detection of the number of CPU?

thedrow: Oh, I guess we can do that although it detects the number of CPUs anyway.

thedrow: The answer seems to be, no we don't need it since we've got `os.sched_getaffinity()` in all supported Python versions we run pytest with.

psutil is only required if we run pytest on the codebase using Python 2.7.

PR #78, opened 2020-09-06
Check for memory leaks with valgrind

thedrow:
I've included suppression files that exclude the memory leaks we already have.
We should remove them once we resolve the memory leaks.

Note that I've included suppressions for two memory leaks that seem to originate from CPython itself.


fangerer:
Hi @thedrow, thank you for this contribution. We just had our monthly HPy call and discussed this PR. We are basically fine with adding this but our concern is that it will influence design (and implementation) just to satisfy it. Of course, this is somewhat the point of such a tool but we also need to keep in mind that it checks a specific implementation (namely CPython). So, I think we can add it but we might also again remove it if it starts being annoying (due to false positives, bugs in CPython or whatever).

fangerer (approved):

thedrow:
Couldn't we run the test suite with valgrind on multiple implementations of Python?

mattip:
> test suite with valgrind on multiple implementations of Python

Why? Can you give an example of a true positive in HPy that would depend on the python version?

thedrow:
> > test suite with valgrind on multiple implementations of Python
> 
> Why? Can you give an example of a true positive in HPy that would depend on the python version?

Isn't it possible that we leak memory when HPy is used with PyPy but not with CPython?
After all, the internals are completely different.

antocuni:
> Isn't it possible that we leak memory when HPy is used with PyPy but not with CPython?
After all, the internals are completely different.

technically, in that case it would be a pypy bug, not an hpy bug. We are not even running tests on pypy here (they are run as part of pypy's buildbot). Note that when you run an HPy module on top of PyPy, you are not executing the code which is in this repo, but the code which is in pypy repo (in the `pypy/module/_hpy_universal` directory).

So, I think that having a job which checks that we are not leaking on top of CPython is enough here (moreover: I think that valgrind would be very confused by PyPy's GC, but that's another story).

antocuni:
FWIW, now valgrind is findind a new leak when running `test_specparam_base`; here is a [link to the report](https://dev.azure.com/pyhandle/hpy/_build/results?buildId=555&view=logs&j=70f75f5e-8692-5c2b-5974-34c6ae62b7d6&t=e326006a-17d5-5be5-4070-9a19165eeed4&l=67)

I didn't investigate the failure, so I don't know whether it's a false positive or a true leak. If it's a true leak it would be a great example of why this PR is useful :)

fangerer:
I absolutely agree with @antocuni and need to add that we would also have a dep cycle if we would run tests on PyPy etc. here. A non-CPython engine first needs to adopt changes of this repo.

Concerning the failure: Yes, I noticed. I will have a look ASAP.

hodgestar:
I'm continuing this PR in #89 so that I can fix the merge conflicts & some of the memory leaks.

(If anyone knows of a better way to work with PRs from branches in external repos, please let me know).

PR #79, opened 2020-09-06
Cache pip and extension compilation

thedrow:
Since we constantly compile code when testing and the tests rarely change,
it'd be wise to setup ccache for them.

I've also went ahead and cached pypi packages for good measure.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -15,11 +15,34 @@ jobs:
         python.version: '3.7'
       py38:
         python.version: '3.8'
+  variables:
+    PIP_CACHE_DIR: $(Pipeline.Workspace)/.pip
+    CCACHE_DIR: $(Pipeline.Workspace)/ccache
   steps:
   - task: UsePythonVersion@0
     inputs:
       versionSpec: '$(python.version)'
     displayName: 'Use Python $(python.version)'
+
+  - script: |
+      sudo apt update && sudo apt install -y ccache
+      echo "##vso[task.prependpath]/usr/lib/ccache"
+    displayName: Install ccache and update PATH to use linked versions of gcc, cc, etc
+
+  - task: Cache@2
+    inputs:
+      key: 'ccache | "$(Agent.OS)" | "$(python.version)"'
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -69,12 +69,35 @@ jobs:
...
   steps:
   - task: UsePythonVersion@0
     inputs:
       versionSpec: '$(python.version)'
     displayName: 'Use Python $(python.version)'
 
+  - script: |
+      sudo apt update && sudo apt install -y ccache
+      echo "##vso[task.prependpath]/usr/lib/ccache"
+    displayName: Install ccache and update PATH to use linked versions of gcc, cc, etc
```
antocuni, line 84: where is the code which update the `PATH`? I see only a `echo` command here

azure-pipelines.yml
```diff
@@ -69,12 +69,35 @@ jobs:
...
+  - task: Cache@2
+    inputs:
+      key: 'ccache | "$(Agent.OS)" | "$(python.version)"'
+      path: $(CCACHE_DIR)
+    displayName: Cache C compilation with ccache
+
+  - task: Cache@2
+    inputs:
+      key: 'python | "$(Agent.OS)"'
+      restoreKeys: |
```
antocuni, line 95: this is more out of curiosity than a review, but why do we need `restoreKeys` for the pip cache and not for the ccache?

azure-pipelines.yml
thedrow: The echo command prepends it to the path.
This is copied from azure pipelines' documentation.

azure-pipelines.yml
thedrow: I didn't give this much thought.
I just copied it from their documentation.

thedrow:
As you can see the PoC tests' runtime has been cut by 50%.
It also cuts our test suite's runtime by 50%.

You can try this locally by installing ccache and running the test suite twice while using it.

hodgestar:
Continuing in #99 so that I can update it with the recent changes to master.

PR #80, opened 2020-09-08
Improve versioning

antocuni:
- use setuptools_scm to automatically set the correct setuptools version
    based on git tags

  - automatically generate hpy/universal/src/version.h at build time

  - introduce hpy.universal.get_version(). The idea is to introduce it also in
    the PyPy version: it will be useful to know which hpy revision we are
    targeting as soon as the pypy hpy branch gets merged and we start to do
    nightly builds


arigo:
A general note, freezing the current ``_hpy_universal`` version inside the next pypy release is likely pointless.  For now, every slight change in ``public_api`` will cause the layout of the ``HPyContext`` to change.  We don't have any binary compatibility at all so far.

antocuni:
re @hodgestar 
> I haven't used setuptools_scm itself, but I've generally found such tools that read the version from the version control system to be quite fragile -- e.g. if one copies the code, or has a tarball, or changes version control or doesn't have quite the right commit -- then these tools break in odd ways.

I have used it for capnpy and it basically worked flawlessly: as soon as you `git tag` a certain commit, this becomes the version used by `setup.py`.
For tarballs: `sdist` tarballs work out of the box. I think that the only unsupported/partially supported use case if you export a tarball directly from the git repo, but I don't think it's something we want to support, do we?

> I've generally preferred to use tools like bumpversion where the version is written into a file and there is just a tool to update the version when making a release.

I don't have any experience with it, but I don't have any prejudice either. If you think it's a good idea, I don't mind using it.

re @arigo 
> A general note, freezing the current _hpy_universal version inside the next pypy release is likely pointless. For now, every slight change in public_api will cause the layout of the HPyContext to change. We don't have any binary compatibility at all so far.

I know, but I am more thinking of nightly builds of pypy3, since the plan is to merge the pypy `hpy` branch into `py36`.
The idea is that we will be able to tell which hpy version a certain nightly build support by calling `hpy.universal.get_version()`: it's mostly for our own convenience :)

hodgestar:
@antocuni Happy to go with this & plus I will get to try something new. If it ever bites us it should be easy to change.

antocuni:
> Happy to go with this & plus I will get to try something new. If it ever bites us it should be easy to change.

cool, I'll merge it then. Thanks!

PR #81, opened 2020-09-10
Fix type refcount in _HPy_New()

rlamy:
On CPython <3.8, PyObject_New doesn't incref the type, so we need to do it ourselves. On 3.8+, it does incref the type iff it's a heap type, so we don't have to do anything.

Cf. https://docs.python.org/3.8/whatsnew/3.8.html#changes-in-the-c-api and https://bugs.python.org/issue35810


antocuni (approved):
LGTM

rlamy:
Note that we should also make sure to decref the type in tp_dealloc. subtype_dealloc does it, but it doesn't get called if we have an HPy_tp_destroy...

PR #82, opened 2020-09-12
Added a TODO comment

thedrow:
I find that it is easier to search what to remove once something happens in code which has TODO comments.
I added this one for good measure.


antocuni (approved):
looks like a good idea to me, I'll leave the merge to @rlamy since he's the original author of the code

PR #87, opened 2020-09-30
Run facebook infer in CI

thedrow:
This PR runs facebook infer in CI.
[infer](https://fbinfer.com/) is a static analysis tool much like coverity, only free for use with much better UX.

I run it locally and got two potential null pointer de-references.
Unlike CPPCheck, it does not check for style or unused code etc.
Instead, it analyzes the code by traversing the IR bytecode from LLVM.

### DIFF
--- a/hpy/devel/src/runtime/ctx_type.c
+++ b/hpy/devel/src/runtime/ctx_type.c
@@ -420,10 +420,7 @@ _HPy_HIDDEN HPy
 ctx_New(HPyContext ctx, HPy h_type, void **data)
 {
     PyObject *tp = _h2py(h_type);
-    if (tp == NULL) {
-        PyErr_SetString(PyExc_RuntimeError, "HPy_New arg 1 must not be a null handle");
-        return HPy_NULL;
-    }
+    assert(tp != NULL);
     if (!PyType_Check(tp)) {
         PyErr_SetString(PyExc_TypeError, "HPy_New arg 1 must be a type");
         return HPy_NULL;
@@ -446,10 +443,7 @@ _HPy_HIDDEN HPy
 ctx_Type_GenericNew(HPyContext ctx, HPy h_type, HPy *args, HPy_ssize_t nargs, HPy kw)
 {
     PyObject *tp = _h2py(h_type);
-    if (tp == NULL) {
-        PyErr_SetString(PyExc_RuntimeError, "HPy_Type_GenericNew arg 1 must not be a null handle");
-        return HPy_NULL;
-    }
+    assert(tp != NULL);
     if (!PyType_Check(tp)) {
         PyErr_SetString(PyExc_TypeError, "HPy_Type_GenericNew arg 1 must be a type");
         return HPy_NULL;
--- a/hpy/devel/src/runtime/ctx_type.c
+++ b/hpy/devel/src/runtime/ctx_type.c
@@ -420,6 +420,10 @@ _HPy_HIDDEN HPy
 ctx_New(HPyContext ctx, HPy h_type, void **data)
 {
     PyObject *tp = _h2py(h_type);
+    if (tp == NULL) {
+        PyErr_SetString(PyExc_RuntimeError, "HPy_New arg 1 must not be a null handle");
+        return HPy_NULL;
+    }
     if (!PyType_Check(tp)) {
         PyErr_SetString(PyExc_TypeError, "HPy_New arg 1 must be a type");
         return HPy_NULL;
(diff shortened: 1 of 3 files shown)


arigo:
Can you also tell us where the two null pointer dereferences are?

thedrow:
@antocuni Yes, of course. I wanted to show you using the build output but if you can't wait... :smile: 

```
make infer
python3 setup.py build_ext -if | compiledb
Capturing using compilation database...
Starting translating and linting 11 files


Found 11 source files to analyze in /home/thedrow/Documents/Projects/hpy/infer-out


Analysis finished in 8.736ss

Found 2 issues

hpy/devel/src/runtime/ctx_type.c:423: error: NULL_DEREFERENCE
  pointer `tp` last assigned on line 422 could be null and is dereferenced at line 423, column 10.
  421.   {
  422.       PyObject *tp = _h2py(h_type);
  423. >     if (!PyType_Check(tp)) {
  424.           PyErr_SetString(PyExc_TypeError, "HPy_New arg 1 must be a type");
  425.           return HPy_NULL;

hpy/devel/src/runtime/ctx_type.c:445: error: NULL_DEREFERENCE
  pointer `type` last assigned on line 444 could be null and is dereferenced at line 445, column 21.
  443.   {
  444.       PyTypeObject *type = (PyTypeObject *)_h2py(h_type);
  445. >     PyObject *res = type->tp_alloc(type, 0);
  446.       return _py2h(res);
  447.   }
```

thedrow:
Please wait before merging.
Infer runs but the job doesn't fail correctly.

thedrow:
> Please wait before merging.
> Infer runs but the job doesn't fail correctly.

This is ready for review & merge.

hodgestar:
@thedrow Thanks! I fixed the NULL dereferences infer found and pushed the changes into this PR.

@antocuni (or others): Let me know if you have any thoughts on the fixes. The checks seem cheap and likely to catch errors in calls to object creation, so they seem like a good idea to me. And we use `Py_XINCREF / Py_XDECREF` on handles everywhere, so having a check here seems consistent.

hpy/devel/src/runtime/ctx_type.c
```diff
@@ -420,6 +420,10 @@ _HPy_HIDDEN HPy
 ctx_New(HPyContext ctx, HPy h_type, void **data)
 {
     PyObject *tp = _h2py(h_type);
+    if (tp == NULL) {
+        PyErr_SetString(PyExc_RuntimeError, "HPy_New arg 1 must not be a null handle");
```
antocuni, line 424: I'm not sure this is the correct way to go:
1. CPython never (or almost never? I don't know) does these kind of checks: it just assumes that the `PyObject*` you pass around are not-NULL

2. The HPy style is to DO these checks, but in debug mode only (debug mode which doesn't exist yet, but that's another story), to not have impact on performances. I'm in favor of saying that it is responsibility of the caller to ensure that the `HPy` handles you pass around are non-NULL. 

3. What about all the other functions? Do you want to introduce NULL checks on *all* of them?

antocuni (changes requested):
As a follow-up of my inline comment: this example showed that `infer` found a false positive: it complained about a possible NULL pointer, but the actual semantics is that it's the caller responsibility to ensure it's non-NULL. So my question is: is there a way to tell the tool "please trust me, I guarantee that this pointer is non-null?"

hpy/devel/src/runtime/ctx_type.c
hodgestar: I think the situation is a bit more complex:

1. Internally HPy mostly passes the result of `_h2py` directly to CPython functions, and assume they do something sane.  In these cases HPy is doing a bit more with the result. If we're not going to do the NULL check, why bother with the PyType_Check either? Why not just pass the result directly to `PyObject_New` and let it sort it out?

2. HPy is already a little inconsistent in places. E.g. `ctx_Dup` does XINCREF, but `ctx_FromPyObject` and `ctx_AsPyObject` do INCREF.

3. We do also have HPy_IsNull, so it's definitely expected that handles can contain null objects in some places.

Happy for you to make the call here though -- my commits were mostly so that we could have this conversation.

hodgestar:
After some discussion in IRC, @antocuni suggested the following:

* Handles passed into functions (and especially context functions mostly called by the Python runtime) can be assumed to be not null handles (and this should be checked by the long-awaited HPy debug mode).
* Handles returned by functions (e.g. `HPyLong_FromLong`) might be null and should be checked by the receiver.

From this point of view, these "null dereferences" are false positives.

@antocuni Then suggested using asserts to silence the warnings, which seems to have worked (and I've pushed the changes to this PR).

hodgestar (approved):

antocuni:
> Handles passed into functions (and especially context functions mostly called by the Python runtime) can be assumed to be not null handles (and this should be checked by the long-awaited HPy debug mode).

More specifically, I think that it is the caller's responsibility to ensure that we pass non-NULL handles, and we should clearly document it.

antocuni (approved):

hodgestar:
@thedrow Merged -- thanks!

PR #90, opened 2020-10-05
Add flake8 to CI

hodgestar:
I've run flake8 locally and there are lots of suggestions from it. They range from completely stylistic (blank lines before nested function definitions), to clean up things (unused imports), to things that should probably be looked at (variables assigned to but never used, f-strings without braces in them).

I'm in favour of using flake8 -- otherwise its too easy for the code base to develop little warts all over the place which make contributing hard (e.g. is this unused import here for a reason or not?).

For those who are less familiar with flake8, turning off checks for a particular instance of an error is extremely easy -- just add `# noqa` to the end of the appropriate line (and optionally specify which error code you want to ignore). See https://flake8.pycqa.org/en/latest/user/violations.html#in-line-ignoring-errors.

Comments from others?


hodgestar:
Failures we have currently: https://dev.azure.com/pyhandle/hpy/_build/results?buildId=575&view=logs&j=3a0d994b-f8b6-55ae-5573-4cdc8287c652&t=d5bda00a-77aa-5b7e-2fd8-ff69c93f1cd8

hodgestar:
I've merged master in, but I'm going to close this PR for now and instead make a case for using black at the next HPy call. Manually fixing all the flake8 complaints is the old outdated way of doing this. :)

Issue #92, opened 2020-10-06
Cannot install without the Git repository


paugier:
Sorry to bother with such a silly thing.

In `setup.py` (https://github.com/hpyproject/hpy/blob/master/setup.py#L18), there is 
```
    gitrev = subprocess.check_output('git describe --abbrev=7 --dirty '
                                     '--always --tags --long'.split(), encoding='utf-8')
```
so that hpy cannot be installed from source without the Git repository.

It seems quite uncommon because I'm able to use hg-git for many Python related projects on Github.

I modified `setup.py` to support also hg-git (https://github.com/hpyproject/hpy/compare/master...paugier:do-not-assume-git-repo). Do you think something like that could be included in master (yes it is a bit specific :slightly_smiling_face:)? Or could there be another solution to install hpy from source without the Git repository?

Is the Git revision use for anything important in HPy?

arigo:
+1 to include this in master.

hodgestar:
@paugier Thank you for pointing this out -- it is indeed a bit of a mess that needs tidying!

hodgestar:
After some discussion in IRC, the consensus seems to be to apply a simpler fix for now. The simpler fix ignores failures and writes a fixed "__UNKNOWN__" git revision if one is not found. I've implemented this in #94, so I'll close this PR.

hodgestar:
@paugier I've merged #94. Please re-open this issue if this still fails for you.

PR #93, opened 2020-10-06
Support install without Git repo when using hg-git

paugier:
The situation is explained in #92. To summarize, it is not possible to install hpy from source (currently the only possibility) without the Git repository. 

With the commit in this PR, it becomes possible to install from source with hg-git.

The correct value of `HPY_GIT_REVISION` is found in hg-git data located in `.hg`.

One difference with this simple version is that Git adds "dirty" if the working directory is modified. It is not done here for simplicity. But at least, hg-git users are not blocked and hpy can be installed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,9 +14,31 @@ def get_scm_config():
     import subprocess
     import pathlib
     import setuptools_scm
+
     version = setuptools_scm.get_version()
-    gitrev = subprocess.check_output('git describe --abbrev=7 --dirty '
-                                     '--always --tags --long'.split(), encoding='utf-8')
+    # After setuptools_scm.get_version, we know that there is a repository
+    try:
+        gitrev = subprocess.check_output(
+            "git describe --abbrev=7 --dirty --always --tags --long".split(),
+            encoding="utf-8",
+        )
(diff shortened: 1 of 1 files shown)


hodgestar:
I think `setup.py:get_scm_config` cannot work as it at all. `use_scm_version` expects a function that returns some modified configuration options -- writing files from there or calling `.get_version` doesn't follow setuptools_scm's expected usage and is already causing problems (e.g. one cannot unpack a source tarball and build something from it).

I'd like to simply remove all this complexity somehow.

hodgestar:
@paugier After some discussion in IRC, the consensus seems to be to apply a simpler fix for now. The simpler fix ignore failures and writes a fixed "__UNKNOWN__" git revision if one is not found. I've implemented this in #94, so I'll close this PR.

Longer term, once HPy is making versioned releases of some kind, I'd like to remove the use of setuptools_scm entirely (if enough other people agree :).

paugier:
Yes, simplicity is better. I wasn't sure whether the value of `HPY_GIT_REVISION` was important or not.

Anyway, thanks it works fine now.

PR #95, opened 2020-10-10
Add dev requirements.

hodgestar:
At the moment this just install pytest and pytest-xdist so that contributors can easily see what's needed to run the tests.

This allows one to do `pip install -e '.[dev]'` to get the pytest requirements.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,10 +39,18 @@ def get_scm_config():
 
     return {} # use the default config
 
+dev_requirements = [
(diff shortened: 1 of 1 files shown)


kimvanwyk (approved):
Works for me, with these commands in a fresh venv:
```bash
python setup.py install
pip install -e .[dev]
```
All tests pass with 
```
pytest -n auto test
```

PR #99, opened 2020-10-11
Cache pip and extension compilation

hodgestar:
Continuation of #79.


thedrow:
There is no ccache on Mac. We should install it using brew.

thedrow:
It might be useful to use the cache when running Valgrind as well since that also runs the test suite.

hodgestar:
> It might be useful to use the cache when running Valgrind as well since that also runs the test suite.

@thedrow:

Thanks. Valgrind didn't seem to generate a ccache folder so I removed it while I was trying to get other things running (to have one fewer things to debug for the moment). I'll try adding it back once CI is passing here again.

Do you have any idea why cppcheck is suddenly finding a lot of errors? Running it locally shows them too.

hodgestar:
> Do you have any idea why cppcheck is suddenly finding a lot of errors? Running it locally shows them too.

There was a new release of the cppcheck yesterday, so possibly that is the reason. The snap installs cppcheck "2.3 dev" and it's a bit unclear what that is (HEAD?).

hodgestar:
> Do you have any idea why cppcheck is suddenly finding a lot of errors? Running it locally shows them too.

I solved the problem by suppressing `nullPointerRedundantCheck` which seems to be a new warning added by cppcheck.

I also removed the PYTHON_INCLUDE_DIR because cppcheck's documentation recommends against including these (despite issuing a notice about header files that weren't found).

hodgestar:
This is not a massive win in overall performance (tests are about 40s faster, but we also lose some time installing ccache), but the test suite is not going to get smaller, so the wins should improve over time.

It also encouraged me to get to grips with Azure Pipelines in a bit more detail, which was probably a good thing.

@thedrow Thanks again for suggesting this and putting in the initial work.

hodgestar:
> I specified to install it from the stable channel for a reason. I guess it isn't that stable after all. neutral_face

Indeed. Not our faults, but fixed for now. :)

PR #101, opened 2020-10-15
A simple manager for closing handles in complicated cases.

hodgestar:


arigo:
I think it shouldn't be fixed-size like that, and there shouldn't be a separate pair of functions ClearAll and Free when in most cases we want to call both and it's wrong to call only one of them.

hodgestar:
> I think it shouldn't be fixed-size like that, and there shouldn't be a separate pair of functions ClearAll and Free 
> when in most cases we want to call both and it's wrong to call only one of them.

Re-resizing -- I didn't implement auto resizing straight away because I didn't want to have to check whether HList_Track encountered a memory allocation error (because that would clutter the code even more). Maybe I can just store the error somewhere and deal with it at the end if needed though (possibly better than returning -1 and not checking for errors like the code does now).

Re _Free and _CloseAll -- I think often in the success case (e.g. insides HPyArg_ParseKeywords) one will want to return the handles in the success case and not close them, but one still needs to free the HList handles. Perhaps I can add _FreeAndCloseAll though? Or perhaps _Free should close everything and there should be an _Clear which explicitly says "I no longer need to track these handles".

arigo:
Re-resizing: yes, memory reallocation errors are annoying.  The problem is that you can't store the error for later, because of the argument passed in Track().  It must either be

arigo:
Note that if we assume it's always an error to put the same handle in more than one HList, then PyPy might have a more efficient implementation for it: allocate a global extra list of integers, and `Track(h)` (where `h` is essentially a small integer) would work by building a chained list with the head inside the HList and adding the node number `h` to it.  It's faster because it doesn't require any extra allocation.

hodgestar:
> Note that if we assume it's always an error to put the same handle in more than one HList, then PyPy might have 
> a more efficient implementation for it: allocate a global extra list of integers, and `Track(h)` (where `h` is essentially 
> a small integer) would work by building a chained list with the head inside the HList and adding the node number > `h` to it. It's faster because it doesn't require any extra allocation.

It would definitely be rather odd to put a handle in more than one HList. I can document it as a rule, but it seems tricky to enforce.

If we think PyPy might want to special case this, should it move outside `hpy/devel/src/runtime` and get placed in `universal` somewhere?

arigo:
Should it move in universal: probably, but also I'm fine if it's put here at first.  It can always move later when we figure out a better PyPy implementation.

thedrow:
I have a related design question: How would Cython use HList?
Should we decide on a threshold of allocations per scope to switch to using an HList?
Are there any advantages for doing so except readability?

hodgestar:
> I have a related design question: How would Cython use HList?
> Should we decide on a threshold of allocations per scope to switch to using an HList?
> Are there any advantages for doing so except readability?

HList is foremost about making it easier to write code that is correct.

If a function creates handles but which and how many handles depends on the inputs, then some dynamic way is needed to keep track of the handles in case they need to be closed later. HList is such a way. If the handles that need to be closed can be determined statically, then HList is probably not necessary in generated code (e.g. inside Cython).

Readability and low overhead are important too, of course.

The first use case for HList will inside HPy itself -- I'm intending to use it to track handles inside `HPyArg_Parse` and `HPyArg_ParseKeywords` to track handles created by the `N` argument specified which returns a new handle. See #75.

hpy/devel/src/runtime/hlist.c
```diff
@@ -0,0 +1,135 @@
...
+        PyErr_NoMemory();
+        return NULL;
+    }
+    hl->size = _HLIST_INITIAL_SIZE;
+    hl->next = 0;
+    return hl;
+}
+
+HPyAPI_RUNTIME_FUNC(int)
+HList_Resize(HPyContext ctx, HList hl, HPy_ssize_t size)
```
antocuni, line 89: does it need to be a public function? It looks to me that the only valid/reasonable use case it to call it from within `HList_Track`?

hpy/devel/src/runtime/hlist.c
```diff
@@ -0,0 +1,135 @@
...
+
+struct _HList_s {
+    HPy_ssize_t size;
+    HPy_ssize_t next;
+    HPy *handles;
+};
+
+
+HPyAPI_RUNTIME_FUNC(HList)
+HList_New(HPyContext ctx)
```
antocuni, line 69: maybe we could add an optional param which specifies the initial size of the pool?

hpy/devel/src/runtime/hlist.c
```diff
@@ -0,0 +1,135 @@
...
+ *    HList_Free(ctx, hl);
+ *    HPy_Close(ctx, dict);
+ *    HPyErr_SetString(ctx, ctx->h_ValueError, "Failed!");
+ *    return HPy_NULL;
+ */
+
+#include <Python.h>
+#include "hpy.h"
+
+#define _HLIST_INITIAL_SIZE (5 + 1)
```
antocuni, line 59: what about `static const HPy_ssize_t` instead of `#define` ?

hpy/devel/src/runtime/hlist.c
```diff
@@ -0,0 +1,135 @@
...
+    hl->next = 0;
+    return hl;
+}
+
+HPyAPI_RUNTIME_FUNC(int)
+HList_Resize(HPyContext ctx, HList hl, HPy_ssize_t size)
+{
+    HPy *new_handles;
+    if (size <= hl->next) {
+        // refuse a resize that would either 1) lose handles or  2) not leave
```
antocuni, line 93: shouldn't we raise a proper exception in this case? (Probably `RuntimeError`?)

hpy/devel/src/runtime/hlist.c
```diff
@@ -0,0 +1,135 @@
...
+ *    return dict;
+ *
+ * error:
+ *    HList_Free(ctx, hl);
+ *    HPy_Close(ctx, dict);
+ *    HPyErr_SetString(ctx, ctx->h_ValueError, "Failed!");
+ *    return HPy_NULL;
+ */
+
+#include <Python.h>
```
antocuni, line 56: If you put it in `devel/src/runtime` you are NOT allowed to include `Python.h`: these files are automatically included in all `hpy_ext_modules`, including the universal ones.
So, two solutions:
1. keep it in `hpy/devel/src/runtime` but implement everything in terms of other HPy primitives (as `argparse.c`)
2. move it to `hpy/universal`, and let each implementation to write its own version. In this case all functions like `HList_New` & co. will probably need to go inside `public_api.h`

antocuni (changes requested):
I am generally +1 on the idea, but see my inline comments; in particular, the one about `Python.h` surely needs to be solved before merging this.

Another topic of discussion: I am not sure I like the name `HList`, for a number of reasons:
1. it's too easy to confuse it with `HPyList`
2. so far, every API function starts with the `HPy` prefix, I think it would be better to stick with it
3. It's not immediately clear what `HList` is for: it's not even a "list" of things in the python sense, because you cannot iterate over it and/or access individual arguments

Random suggestions for a better name, feel free to add more:
- `HPyPool`
- `HPyHandlePool`
- `HPyHandleManager`
- `HPyManager`
- `HPyTracker`
- `HPyHandleTracker`
- `HPyHandles`
- `HPys` (the plural of `HPy`... this is very obscure :) )

hpy/devel/src/runtime/hlist.c
hodgestar: Agreed about the name. I like HPyTracker.

Re Python.h -- note that, e.g., ctx_listbuilder in runtime also includes Python.h. Are those different somehow? Should they not use Python.h? Or not be in runtime? Or something else?

hodgestar:
Agreed about the name. I like HPyTracker.

Re Python.h -- note that, e.g., ctx_listbuilder in runtime also includes Python.h. Are those different somehow? Should they not use Python.h? Or not be in runtime? Or something else?

hpy/devel/src/runtime/hlist.c
hodgestar: I felt that it would be very helpful to expose it because if one can calculate the number of handles needed, then one can avoid a few mallocs by calling Resize explicitly oneself.

hodgestar: That said, I exposed Resize before the extra handle logic was added, which perhaps makes Resize currently a bit too complex to use well. Perhaps Resize should internally allocate space for one more handle than requested.

hpy/devel/src/runtime/hlist.c
hodgestar: I was thinking about this but was trying to avoid adding more va_list things. Perhaps I can just add a second function for creating the tracker.

hpy/devel/src/runtime/hlist.c
hodgestar: Can do. Is there a particular advantage to defining it as a variable here?

hpy/devel/src/runtime/hlist.c
hodgestar: Not sure. I see the tracker as largely a C helper that shouldn't raise Python errors, but I could be convinced otherwise.

hpy/devel/src/runtime/hlist.c
hodgestar: Done.

hpy/devel/src/runtime/hlist.c
hodgestar: Second function added.

hpy/devel/src/runtime/hlist.c
antocuni: another alternative is to offer only `HPyTracker_New(ctx, initial_size)`. If you know it in advance, you pass a precise number, else you pass 0 (or -1, if you want a default) and let `HPyTracker` to take care of that.

hpy/devel/src/runtime/hlist.c
antocuni: > Is there a particular advantage to defining it as a variable here?

generally speaking I think that variables are better, more modern and more type safe than macros. Maybe it's just a matter of taste though

hpy/devel/src/runtime/hlist.c
antocuni: well, it raises `PyErr_Mem` few lines below. I am fine to say that it never raises anything, but then we should be consistent.

antocuni:
> Re Python.h -- note that, e.g., ctx_listbuilder in runtime also includes Python.h. Are those different somehow? Should they not use Python.h? Or not be in runtime? Or something else?

You are correct, it might be that `hpy/devel/runtime` is too overloaded and should be split in two; currently, there are two kind of files:

- `argparse.c`: this needs to be in `hpy/devel` because it is automatically compiled in all `hpy_ext_modules`: it offers a C API without putting functions in the ABI/ctx.

- `ctx_*.c`: these files are compiled in the final extension ONLY when targeting the CPython ABI. When you target the universal ABI, they are not included at all. Coincidentally, the `hpy.universal` module does include them, so when you load an universal extension on CPython, you end up calling this code again, but you always go through the ctx indirection. From a purely theoretical point of view, you could have two totally separate implementations for the CPython and Universal cases; but in practice, we try to share the code of course :) . That's also why some `ctx_*.c` files are un `hpy/devel` and some in `hpy/universal`: the latter are universal only and have their own (different) implementation in `hpy/devel` for the CPython-ABI case.

So, back to `HPyTracker`: if you put the functions in `public_api.h`, you are allowed to use `Python.h` because it's CPython-only code. If you DO NOT put them in `public_api.h`, they need to be CPython-agnostic

hpy/devel/src/runtime/ctx_tracker.c
```diff
@@ -0,0 +1,154 @@
...
+ *    handle.
+ *
+ *    Calling HPyTracker_NewWithSize(ctx, n) will ensure that at least n handles
+ *    can be tracked without the need for a resize.
+ *
+ * Example usage (inside an HPyDef_METH function)::
+ *
+ * long i;
+ * HPy key, value;
+ * HPyTracker hl;
```
TeamSpen210, line 23: Minor issue, but this still has an old `HList` variable name. Maybe `trk` or `track`?

hodgestar: Good catch! I renamed them to `ht` for `HPyTracker`.

hpy/devel/src/runtime/ctx_tracker.c
```diff
@@ -10,10 +10,10 @@
  *    HPyTracker_Add.
  *
  *    Space for this extra handle is created automatically, so
- *    HPyTracker_NewWithSize(ctx, 0) will actually allocated space for one
- *    handle.
+ *    HPyTracker_New(ctx, 2) will actually allocated space for three
```
antocuni, line 13: isn't this an implementation detail now that resize is private? From the user point of view, it is enough to know that HPyTracker_Add can fail, and that in that case he/she needs to call HPyTracker_Free, isn't it?

hodgestar: It is more of an implementation detail, but the "Calling HPyTracker_New(ctx, n) will ensure that HPyTracker_Add cannot fail for the first n handles" is intended to be a guarantee across implementations.

Maybe that is not a good idea though?

hodgestar:
@antocuni Thank you for all of the reviewing and feedback. I think I have now addressed all the comments, but let me know if I missed anything or if you find anything more.

antocuni (approved):
LGTM, thanks for this!

hodgestar:
Yay! Merging.

Issue #103, opened 2020-10-18
Error extension name with --hpy-abi=universal


paugier:
I try to build piconumpy with the flag `--hpy-abi=universal`. I use `python setup.py --hpy-abi=universal build_ext -if` and I get the following error:

```
running build_ext
building 'piconumpy._piconumpy_cpython_capi' extension
C compiler: gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC

creating build
creating build/temp.linux-x86_64-3.8
creating build/temp.linux-x86_64-3.8/piconumpy
compile options: '-I/home/pierre/.pyenv/versions/3.8.2/include/python3.8 -c'
extra options: '-Wfatal-errors -Werror'
gcc: piconumpy/_piconumpy_cpython_capi.c
creating build/lib.linux-x86_64-3.8
creating build/lib.linux-x86_64-3.8/piconumpy
gcc -pthread -shared -L/home/pierre/.pyenv/versions/3.8.2/lib -L/home/pierre/.pyenv/versions/3.8.2/lib build/temp.linux-x86_64-3.8/piconumpy/_piconumpy_cpython_capi.o -o build/lib.linux-x86_64-3.8/piconumpy/_piconumpy_cpython_capi.cpython-38-x86_64-linux-gnu.so
building 'piconumpy._piconumpy_hpy' extension
C compiler: gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC

creating build/temp.linux-x86_64-3.8/home
creating build/temp.linux-x86_64-3.8/home/pierre
creating build/temp.linux-x86_64-3.8/home/pierre/Dev
creating build/temp.linux-x86_64-3.8/home/pierre/Dev/hpy
creating build/temp.linux-x86_64-3.8/home/pierre/Dev/hpy/hpy
creating build/temp.linux-x86_64-3.8/home/pierre/Dev/hpy/hpy/devel
creating build/temp.linux-x86_64-3.8/home/pierre/Dev/hpy/hpy/devel/src
creating build/temp.linux-x86_64-3.8/home/pierre/Dev/hpy/hpy/devel/src/runtime
compile options: '-DHPY_UNIVERSAL_ABI -I/home/pierre/Dev/hpy/hpy/devel/include -I/home/pierre/.pyenv/versions/3.8.2/include/python3.8 -c'
[...]

PR #105, opened 2020-10-20
Improve setuptools integration

### LINKED ISSUE #86: Test proof-of-concept in universal mode
We're only testing `proof-of-concept/` in native CPython mode, but should also make sure that the universal mode is actually usable (which it isn't right now, see #85).

antocuni:
generate the correct .hpy.so extension when building an universal module which is inside a package.

See the docstring for the details of what is the problem and what is the proposed solution.
I marked this as WIP because I don't want to merge it yet but I'd like to receive feedback and/or ideas for alternative solutions.

Currently this code is untested by our CI, so I plan to add an azure job to install/test pof in universal mode; this would also fix #86


antocuni:
@arigo adding you as a reviewer since IIRC we wrote this code together.

arigo:
If this workaround helps, then sure.  Sorry but I don't have anything more precise to offer here.

hodgestar:
@antocuni I have an alternative suggestion: Since in `fix_distribution` we add the hpy extensions to `dist.ext_modules` ourselves, why don't we just tag them there? So if we see `foo.bar`, we add `hpy:foo.bar` to the extensions and in `build_extension` we can just check if the extension starts with `hpy:` and strip that from the name as part of the fixing we do?

antocuni:
> I have an alternative suggestion: Since in fix_distribution we add the hpy extensions to dist.ext_modules ourselves, why don't we just tag them there? So if we see foo.bar, we add hpy:foo.bar to the extensions and in build_extension we can just check if the extension starts with hpy: and strip that from the name as part of the fixing we do?

I think we already tried to go down that route, but it doesn't work because the rest of distutils expect it to be a "real" name. I don't remember the details, but after a while we either gave up or decided that it was not going to work, and we want with the current approach

hodgestar:
> I think we already tried to go down that route, but it doesn't work because the rest of distutils expect it to be a "real" name. I 
> don't remember the details, but after a while we either gave up or decided that it was not going to work, and we want with the 
> current approach

Bleh. I see. In that case I am fine with this approach.

antocuni:
I think the PR is read to be reviewed. It improves over the current situation w.r.t. packaging and installing:
- fix issue #103: the problem was about `hpy_ext_modules` inside packages, so I added one in `proof-of-concept` so we can test this case

- `setup.py --hpy-abi=universal install` does not work, so we detect the case and complain early about it.

- fix issue #86: now we test pof in three different modes: using wheels, using `setup.py install` (cpython ABI) and using `setup.py --hpy-abi=universal build_ext --inplace`. We need to import universal modules "by hand": that's bad, but it's the only supported way for now

It is obvious that eventually we need a better strategy for how to compile/install/distribute universal extensions, but I think that we need upstream support in setuptools and/or pip for it. I think that for now this is good enough to let people experiment.

azure-pipelines.yml
```diff
@@ -58,14 +58,17 @@ jobs:
...
-  - script: python -m pytest proof-of-concept/
-    displayName: 'Run pof tests'
+
+  - script: proof-of-concept/test_pof.sh wheel
+    displayName: 'Test bdist_wheel'
+
+  - script: proof-of-concept/test_pof.sh setup_py_cpython
+    displayName: 'Test setup.py --hpy-abi=cpython install'
+
+  - script: proof-of-concept/test_pof.sh build_ext_inplace_universal
```
hodgestar, line 68: Very minor nitpick -- I've been trying to not have blank lines within a job so that one can easily see where jobs start and end (only mentioning this because I arranged the rest of pipeline file to be like that).

hodgestar (approved):
The situation is indeed a mess at the moment, but this shows what is needed to get things installed right now and runs tests against universal builds, so +1.

azure-pipelines.yml
antocuni: fine for me, addressed in 3c347ef.
I don't like azure-pipelines way of organizing things: it would be MUCH easier if you could write the `displayName` BEFORE the script, but I don't think it's possible. Too bad, I suppose

hodgestar: I think putting the displayName before the script should be fine? It's just YAML so the `-` is "this is a list item" and "script" and "displayName" are just keys that are part of the dict that is said list item.

antocuni: that's what I would think as well, but it turns out not to be the case; see e.g. commit 
1b8bda9 and the corresponding [azure pipelines error](https://github.com/hpyproject/hpy/runs/1288900114)

hodgestar: O_O

PR #106, opened 2020-10-20
Handle the memory error case in ctx_ListBuilder_Cancel and ctx_TupleBuilder_Cancel.

hodgestar:
Currently ctx_ListBuilder_Cancel and ctx_TupleBuilder_Cancel don't cater for the case where a memory error occurred when the builder was created.

Caveats:
* This doesn't raise a memory error like _Build does, but on the other hand the result of the building wasn't needed, so maybe its fine to swallow the memory error? Since _Cancel is likely to be called when an error is already set, it probably shouldn't raise errors of its own.
* No idea how to test this case.

Assuming _Cancel is called when an error is detected, it might be more likely that _Cancel is called after a MemoryError rather than _Build.


antocuni (approved):
LGTM

PR #107, opened 2020-10-21
use __VA_ARGS__ also for HPyDef_METH

antocuni:
use `__VA_ARGS__` also for `HPyDef_METH`, as we already do for `_MEMBER`, `_GETSET`, etc.


hodgestar (commented):
:+1:

PR #108, opened 2020-10-21
Enable and test all the remaning sq_* slots

antocuni:
As the title says


test/test_slots.py
```diff
@@ -287,3 +263,163 @@ def test_nb_ops_power(self):
...
+            @EXPORT_POINT_TYPE(&Point_x, &Point_y, &Point_setitem)
+            @INIT
+        """)
+        p = mod.Point()
+        p[0] = 100
+        assert p.x == 100
+        p[1] = 200
+        assert p.y == 200
+        with pytest.raises(IndexError):
+            p[2] = 300
```
hodgestar, line 330: Should we also test what happens with `del p[i]` here?

hodgestar (approved):
I left a small request for an additional test, but looks good to me otherwise.

test/test_slots.py
antocuni: good point, added by 4208706

PR #110, opened 2020-10-25
Improve tests

antocuni:
This PR improves/extends our test suite to check for corner cases which were unchecked before.

Most of these corner cases were discovered while implementing the PyPy version of hpy, so they are probably very useful for other implementations as well. All of them are already integrated in the pypy hpy branch.


hodgestar (approved):
Woot. Thanks for adding these extra tests!

PR #111, opened 2020-10-30
HPyArg_Parse improvements

hodgestar:
Including:

* Add a new handle argument format (O+).
* Add support for `:` and `;` error message handling.
* Add SystemError.
* Add HPyErr_Clear.
* Convert appropriate error messages to the SystemError (i.e. ones that are caused by incorrect arguments to HPyArg_* not by incorrect calls to the Python function).

To do in follow up PRs:

* Implement handle tracking in `HPyArg_ParseKeywords`.
* Consider replacing `O+` with `O` in `HPyArg_ParseKeywords` and return an HPyTracker allowing the handles to be closed on success.
* Add the remaining simple parsing types (with suitable error checking).


hpy/devel/src/runtime/argparse.c
```diff
@@ -34,54 +100,108 @@ int _HPyArg_ParseItem(HPyContext ctx, HPy current_arg, const char **fmt, va_list
...
     }
     return 1;
 }
 
+static const char *
+parse_errstring(const char *fmt, const char **err_funcname, const char **err_message)
+{
+    const char *fmt1 = fmt;
+
+    for (fmt1 = fmt; *fmt1 != 0; fmt1++) {
```
thedrow, line 131:
```suggestion
    for (const char *fmt1 = fmt; *fmt1 != 0; fmt1++) {
```

Since we're using C99 this is possible and I also think it should be encouraged.
Infer is complaining about this anyway for some reason.

hpy/devel/src/runtime/argparse.c
```diff
@@ -34,54 +100,108 @@ int _HPyArg_ParseItem(HPyContext ctx, HPy current_arg, const char **fmt, va_list
...
+        if (*fmt1 == ':') {
+            *err_funcname = fmt1 + 1;
+            break;
+        }
+        if (*fmt1 == ';') {
+            *err_message = fmt1 + 1;
+            break;
+        }
```
thedrow, line 139:
```suggestion
        if (*fmt1 == ':' || *fmt1 == ';') {
            *err_funcname = fmt1 + 1;
            break;
        }
```

This is cleaner in my opinion and it's likely that the compiler does this anyway.

hpy/devel/src/runtime/argparse.c
```diff
@@ -34,54 +100,108 @@ int _HPyArg_ParseItem(HPyContext ctx, HPy current_arg, const char **fmt, va_list
...
+    return fmt1;
+}
+
+static void
+set_error(HPyContext ctx, HPy exc, const char *err_funcname, const char *err_message, const char *msg) {
+    char err_buf[_MAX_ERR_STRING_LENGTH];
+    if (err_message != NULL) {
+        snprintf(err_buf, _MAX_ERR_STRING_LENGTH, "%s", err_message);
+    }
+    if (err_message == NULL) {
```
thedrow, line 150:
```suggestion
    else {
```

I'm not sure if this is the right logic but it might be. I haven't really followed through the code.

hpy/devel/src/runtime/argparse.c
```diff
@@ -94,75 +214,95 @@ HPyAPI_RUNTIME_FUNC(int)
...
+    int optional = 0;
+    int keyword_only = 0;
+    HPy_ssize_t i = 0;
+    HPy_ssize_t nkw = 0;
+    HPy current_arg;
+
+    fmt_end = parse_errstring(fmt, &err_funcname, &err_message);
+
+    // first count positional only arguments
+    while (keywords[nkw] != NULL && !*keywords[nkw]) nkw++;
```
thedrow, line 231: Do we enforce braces for blocks?

hpy/devel/src/runtime/argparse.c
```diff
@@ -94,75 +214,95 @@ HPyAPI_RUNTIME_FUNC(int)
...
-        }
-      }
-      else {
-        HPyErr_SetString(ctx, ctx->h_TypeError, "XXX: no value for required argument");
+
+    va_list vl;
+    va_start(vl, keywords);
+
+    while (fmt1 != fmt_end) {
+        if (*fmt1 == '|') {
```
thedrow, line 246: Personally, I'd extract each condition to a named helper function as it makes the code clearer and requires fewer comments to understand.
They'll be inlined anyway.

I don't feel strongly about this one though.

thedrow (changes requested):
Just a few nitpicks.

hpy/devel/src/runtime/argparse.c
antocuni: we don't have style guidelines. Maybe we should

hpy/devel/src/runtime/argparse.c
hodgestar: The first block assigns to `err_funcname` and the second one to `err_message`, so I'm not clear on how they can be combined.

hpy/devel/src/runtime/argparse.c
hodgestar: Braces added!

I want to discuss style guidelines / linters on the HPy call on Thursday morning (I'd like to have them, but we need to agree as a group on what to have exactly).

hpy/devel/src/runtime/argparse.c
hodgestar: Thanks! This was leftover from refactoring `set_error` a few times.

hpy/devel/src/runtime/argparse.c
hodgestar: I'm happy for these couple of special cases to stay, but I do want to split up the pieces of `parse_item`.

hodgestar:
@thedrow Thank you very much for the review and suggestions!

hpy/devel/src/runtime/argparse.c
thedrow: My bad. I missed it.

thedrow (approved):
LGTM.

hodgestar:
Merging & continuing with follow-up PRs. Tx again for the review @thedrow.

PR #113, opened 2020-11-03
Rename HPyTracker_Free to _Close and _RemoveAll to _ForgetAll

hodgestar:
After some discussion on IRC we decided that HPyTracker_Close communicates the more important part of what the _Free method does (i.e. closes the handles) and that HPyTracker_ForgetAll better communicates that the handles are "forgotten" (i.e. no longer tracked).


antocuni (approved):
LGTM

Issue #17, opened 2020-02-29
Enough mature for a HPy port of PicoNumpy? 


paugier:
I would like to port to HPy a tiny extension providing a (very very limited) Numpy-like array class: https://github.com/paugier/piconumpy

The project is explained in the README, which also contains a list of the functions of the CPython C-API used in PicoNumpy.

Do you think HPy will soon be mature enough for HPy ports of such tiny extensions?

hodgestar:
@paugier PicoNumpy looks like an excellent idea to try with HPy at this stage. A few of us are at an HPy sprint this week where we hope to implement support for creating custom Python types with HPy. Once that is in place, you should be able to mostly implement PicoNumpy on HPy. If you find any other blockers please let us know.

Once it's working we'd love to see the performance benchmarks you get!

paugier:
> an HPy sprint this week where we hope to implement support for creating custom Python types with HPy.

Exactly what I need for PicoNumpy! Have a fun and productive week then!

hodgestar:
@paugier We're working on adding types in https://github.com/pyhandle/hpy/pull/22.

paugier:
Just to say that I rewrote the C extension in PicoNumpy using `PyType_Slot`, `PyType_Spec` and `PyType_FromSpec` so that the extensions written with both APIs will be very similar.

I can test that as soon as #22 is ready.

The benchmark results will be interesting. Without HPy, it gives:

- With CPython

```raw
Julia                      :     1 * norm = 0.00534 s
Transonic-Pythran          : 0.564 * norm
Numpy                      :  18.2 * norm
PicoNumpy (purepy)         :  23.7 * norm
PicoNumpy (purepy_array)   :  21.7 * norm
PicoNumpy (Cython)         :  15.6 * norm
PicoNumpy (CPython C-API)  :   4.7 * norm
```

- With PyPy3

```raw
Julia                      :     1 * norm = 0.00534 s
Transonic-Pythran          : 0.603 * norm
Numpy                      :   183 * norm
PicoNumpy (purepy)         :   2.4 * norm
PicoNumpy (purepy_array)   :  3.73 * norm
PicoNumpy (Cython)         :   135 * norm
PicoNumpy (CPython C-API)  :  17.8 * norm
```

antocuni:
An update on this issue: I started to create an hpy version of piconumpy, incrementally porting the C/API version one function/method at a time.

[...]

PR #114, opened 2020-11-03
fix build on pypy

antocuni:
As reported by #17 , the current behavior of `setup.py` is confusing: if you run it on top of PyPy it also tries to build `hpy.universal`: this is doubly wrong because:
1. The point of `hpy/universal/src` is to run on CPython only
2. the built module is ignored on PyPy anyway, because it is shadowed by `lib_pypy/hpy/universal.py` (which is shipped with PyPy)

This PR makes sure that:

1. you error out if you try to compile `hpy.universal` on PyPy
2. `hpy.universal` is built&installed only on CPython

### DIFF
--- a/hpy/universal/src/hpymodule.c
+++ b/hpy/universal/src/hpymodule.c
@@ -12,6 +12,10 @@
 #include "handles.h"
 #include "common/version.h"
 
+#ifdef PYPY_VERSION
+#  error "Cannot build hpy.univeral on top of PyPy. PyPy comes with its own version of it"
+#endif
+
 typedef HPy (*InitFuncPtr)(HPyContext ctx);
 
 static PyObject *set_debug(PyObject *self, PyObject *args)
--- a/setup.py
+++ b/setup.py
@@ -1,3 +1,4 @@
+import sys
 import os
 from setuptools import setup, Extension
 
@@ -39,19 +40,9 @@ def get_scm_config():
 
     return {} # use the default config
 
-dev_requirements = [
-    "pytest",
-    "pytest-xdist",
-]
-
-setup(
-    name="hpy.devel",
-    packages = ['hpy.devel'],
-    include_package_data=True,
-    extras_require={
-        "dev": dev_requirements,
-    },
-    ext_modules = [
+EXT_MODULES = []
+if sys.implementation.name == 'cpython':
+    EXT_MODULES += [
         Extension('hpy.universal',
                   ['hpy/universal/src/hpymodule.c',
                    'hpy/universal/src/handles.c',
@@ -73,8 +64,24 @@ setup(
                   extra_compile_args=[
                       '-DHPY_UNIVERSAL_ABI',
                   ] + EXTRA_COMPILE_ARGS
-        )],
-      entry_points={
+                  )
+        ]
+
+
+DEV_REQUIREMENTS = [
+    "pytest",
+    "pytest-xdist",
+]
+
+setup(
+    name="hpy.devel",
+    packages = ['hpy.devel'],
+    include_package_data=True,
+    extras_require={
+        "dev": DEV_REQUIREMENTS,
+    },
+    ext_modules = EXT_MODULES,
+    entry_points={
           "distutils.setup_keywords": [
               "hpy_ext_modules = hpy.devel:handle_hpy_ext_modules",
           ],


hpy/universal/src/hpymodule.c
```diff
@@ -12,6 +12,10 @@
 #include "handles.h"
 #include "common/version.h"
 
+#ifdef PYPY_VERSION
```
hodgestar, line 15: I'm wondering what we should do about other cases -- e.g. this should likely also error on GraalPython, but probably not error on Pyston.

setup.py
```diff
@@ -39,19 +40,9 @@ def get_scm_config():
...
-setup(
-    name="hpy.devel",
-    packages = ['hpy.devel'],
-    include_package_data=True,
-    extras_require={
-        "dev": dev_requirements,
-    },
-    ext_modules = [
+EXT_MODULES = []
+if sys.implementation.name == 'cpython':
```
hodgestar, line 44: This is a bit strange because currently this repo is two things:

1) a toolkit of headers, etc for developing hpy extensions and compiling them either as cpython-only or universal
2) an implementation of hpy.universal for cpython.

Perhaps at some point hpy.universal for cpython should be split out into its own section more properly and then it is only a dependency of the C extension that uses the HPy API and not of hpy itself.

hodgestar (approved):
Approved with some comments for the future. Tx for fixing this.

setup.py
paugier: Do you think there will be 2 packages on PyPI ? A build time dependency (hpy.devel) and a runtime dependency for CPython (hpy.universal) ? I don't know if it is possible (and good) to produce two distinct PyPI packages with one setup.py / pyproject.toml. Also I don't know how to specify a runtime dependency only for one Python implementation.

mattip: > Also I don't know how to specify a runtime dependency only for one Python implementation.

In a requirements.txt and pyproject.toml you can do

```
hpy.devel ; platform_python_implementation != "PyPy"
```

mattip: > I don't know if it is possible (and good) to produce two distinct PyPI package

It is non-standard, but has been done in other projects. This was recently discussed on [discuss.python.org](https://discuss.python.org/t/use-case-supported-2-namespace-packages-in-single-source-repo-with-1-setup-py/5440).

hpy/universal/src/hpymodule.c
antocuni: I fear the only option is to add it case-by-case.
Note that for PyPy it's just a sanity check: if you compile it it works, but it's clearly not what you want.

setup.py
antocuni: > This is a bit strange because currently this repo is two things:
[cut]

> Do you think there will be 2 packages on PyPI ? A build time dependency (hpy.devel) and a runtime dependency for CPython (hpy.universal) ? 

yes, the plan since the beginning was to have one repo and two distinct pypi packages (I already pre-registered them on PyPI :)). We need to have them in the same repo because there are some files which are shared between the two so it's much easier to manage this way.

- `hpy.devel` is a build-time dependency. It is supposed to be put in `setup_requires`

- `hpy.universal` is a runtime dependency. For CPython, you will download it from PyPI, but e.g. PyPy will ship its own version (as it does with cffi and vmprof).

> I don't know if it is possible (and good) to produce two distinct PyPI packages with one setup.py / pyproject.toml.

I don't know either. At worst, we can have two different files: `setup_hpy_devel.py` and `setup_hpy_universal.py`: it's a bit nonstandard but I don't see why it shouldn't work.

> Also I don't know how to specify a runtime dependency only for one Python implementation.

PyPy will also ship `hpy.universal.egg-info`, as it [already does](https://foss.heptapod.net/pypy/pypy/-/tree/branch/py3.6/lib_pypy) for `cffi`, `readline`, `greenlet` etc. This way if you do `install_requires=['hpy.universal']`, `pip` and `setuptools` will download the package from PyPI **only** on CPython.

However, there is ANOTHER problem, which might be impossible to solve with the current tools. The dependency on `hpy.universal` is not a property of the package itself, but of the way it is built. If you are on CPython and you build your package with the CPython ABI, the wheel that you produce has no runtime dependency at all. But if you build it with the universal ABI, then you suddenly depends on `hpy.universal`.

I suppose that an easy way to solve is to always depend on `hpy.universal` and if you don't need it... too bad.

mattip: The last point sounds like [this long discussion](https://discuss.python.org/t/options-to-build-the-same-package-in-different-ways-with-different-build-dependencies/4458) about building the same package in different ways. @paugier it seems you were an active participant there: any more insights after that discussion?

paugier: I can maybe add that h5py does that to build a standard package built against sequential hdf5 and a MPI package built against hdf5 MPI (see https://github.com/h5py/h5py/pull/1619).

setuptools supports specifying build dependencies both in pyproject.toml and in setup.py, and the later can be controlled with environment variables. It works fine for now with isolated build (we use that in fluiddyn packages) but I guess people working on improving Python packaging don't like that because (from what I understood) (i) it's better to be able to know the build dependencies without Python execution (without setup.py), (ii) there is no mechanism to change the name of wheels depending on the build and to specify that you want to install a specific build (similar to `conda install "blas-devel[build=*openblas]"`).

PR #115, opened 2020-11-04
Add handle tracking to HPyArg_Parse and HPyArg_ParseKeywords

hodgestar:
Adds an optional `HPyTracker *ht` argument to both `HPyArg_Parse` and `HPyArg_ParseKeywords`. One can pass `NULL` if the tracker is not needed.

* Both `HPyArg_Parse` and `HPyArg_ParseKeywords` create the tracker if `ht` is not `NULL`.
* `HPyArg_Parse` currently doesn't use the tracker itself, but it still creates one if requested. The idea is that we will need tracking for other format types that create temporary storage in the future.
* `HPyArg_ParseKeywords` doesn't require a tracker unless `O` is used as a format, in which case it raises an error and complains that a tracker is required in this case.
* `O+` is gone, and `O` now calls `HPy_Dup(current_arg)` and `HPyTracker_Add(ctx, ht, output)` when `current_arg` needs to be dup'ed (i.e. when called by `HPyArg_ParseKeywords`).

This replaces the argument parsing portions of #75. #116 removes `HPyDict_GetItem` and `HPyDict_SetItem`.

### DIFF
--- a/hpy/devel/src/runtime/argparse.c
+++ b/hpy/devel/src/runtime/argparse.c
@@ -27,9 +27,10 @@
  * -------
  *
  * O (object) [HPy *]
- *     Returns an existing handle. Not supported by HPyArg_ParseKeywords as
- *     retrieving an item from the keywords dictionary would require creating
- *     a new handled. Use O+ instead for this case.
+ *     Returns an handle. When using HPyArg_ParseKeywords, this might retrieve
+ *     a new handle from the keywords dictionary, so an HPyTracker pointer
+ *     must be passed and on successful return HPyTracker_Close must be called
+ *     to close any handles that were created.
  *
  * O+ (object) [HPy *]
  *     Returns a new handle. The new handle must be closed if the argument
@@ -94,14 +95,14 @@ set_error(HPyContext ctx, HPy exc, const char *err_fmt, const char *msg) {
         snprintf(err_buf, _ERR_STRING_MAX_LENGTH, "%.200s() %.256s", err_fmt + 1, msg);
     }
     else {
-          snprintf(err_buf, _ERR_STRING_MAX_LENGTH, "%s", err_fmt + 1);
+        snprintf(err_buf, _ERR_STRING_MAX_LENGTH, "%s", err_fmt + 1);
     }
     HPyErr_SetString(ctx, exc, err_buf);
 }
 
 
 static int
-parse_item(HPyContext ctx, HPy current_arg, const char **fmt, va_list *vl, const char *err_fmt)
+parse_item(HPyContext ctx, HPyTracker *ht, HPy current_arg, int current_arg_tmp, const char **fmt, va_list *vl, const char *err_fmt)
 {
     switch (*(*fmt)++) {
     case 'i': {
@@ -132,18 +133,13 @@ parse_item(HPyContext ctx, HPy current_arg, const char **fmt, va_list *vl, const
         break;
     }
     case 'O': {
-        if (**fmt == '+') {
-            (*fmt)++;
-            HPy *output = va_arg(*vl, HPy *);
-            _BREAK_IF_OPTIONAL(current_arg);
+        HPy *output = va_arg(*vl, HPy *);
+        _BREAK_IF_OPTIONAL(current_arg);
+        if (current_arg_tmp) {
             *output = HPy_Dup(ctx, current_arg);
-            if (HPy_IsNull(*output))
-                return 0;
-            break;
+            HPyTracker_Add(ctx, *ht, *output);
         }
         else {
-            HPy *output = va_arg(*vl, HPy *);
-            _BREAK_IF_OPTIONAL(current_arg);
             *output = current_arg;
         }
         break;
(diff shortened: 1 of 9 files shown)


test/test_tracker.py
```diff
@@ -10,15 +10,15 @@
 
 
 class TestHPyTracker(HPyTest):
-    def hpytracker_module(self, ops, new="HPyTracker_New(ctx, 0)"):
+    def hpytracker_module(self, ops, size=0):
```
hodgestar, line 13: The `size` change is unrelated to the argument parsing, but @antocuni complained that passing more code here made finding the code that was creating the tracker complicated, so I improved this while I was in the file.

antocuni:
I didn't review the PR yet, but I would prefer to think a bit more about what is the best API to implement. Let me paste what I wrote in [this comment](https://github.com/hpyproject/hpy/pull/75#discussion_r515905754).
I think we have three options w.r.t `HPyArg_Parse*` and `HPyTracker`:

1. HPyArg_Parse returns int and HPyArg_ParseKeywords returns HPyTracker
2. they both return HPyTracker (but it will be unused for HPyArg_Parse)
3. they both return int and we return the tracker via an output arg. This has also the advantage to avoid allocating the tracker if we happen not to need it.

It seems that here you have implemented option number 3, is this correct? Any reason to prefer it over 1 and 2?
One pros of solution 3 that I didn't think about is that you don't need a tracker unless you use `O`, so this greatly simplifies simple cases. 

I'm deeply undecided about it: basically, I don't like any of these options for one reason or another, so I wonder whether there is a better solution that we didn't think about yet.

hodgestar:
@antocuni Correct, I have implemented option 3.

I did try options 1 and 2 initially, but there are a few issues:

* Returning HPyTracker instead of `int` makes the check for success require doing some version of `HPyTracker ht; ht = HPyArg_Parse(...); if (HPyIsNull(ht)) {...}` and later `HPyTracker_Close(ctx, ht)`. One has to do these always even if one doesn't need the HPyTracker at all (e,g. one doesn't use `O` or are calling `HPyArg_Parse`).

* One always has to create the HPyTracker even if it isn't used (which it isn't in the majority of cases).

* At some point in the future we likely will need a tracker for HPyArg_Parse when we want support returning buffers that later need to be freed.

The first reason was probably what I disliked the most.

What I liked about where this ended up:

* For simple cases one only needs to pass `NULL` as an extra argument to HPyArg_Parse or HPyArg_ParseKeywords and not have to deal with the HPyTracker at all.

* The HPyTracker is optional, but available easily if one uses a format code that requires it.

I 100% agree that we should think carefully about this API -- pretty much every method function is going to use it.

hpy/devel/src/runtime/argparse.c
```diff
@@ -27,9 +27,10 @@
  * -------
  *
  * O (object) [HPy *]
- *     Returns an existing handle. Not supported by HPyArg_ParseKeywords as
- *     retrieving an item from the keywords dictionary would require creating
- *     a new handled. Use O+ instead for this case.
+ *     Returns an handle. When using HPyArg_ParseKeywords, this might retrieve
```
antocuni, line 30: what about this:
```
Store a handle pointing to a generic Python object. 
To use O, you must provide an HPyTracker. All the handles returned by O are
automatically closed when you call HPyTracker_Close.
```

hpy/devel/src/runtime/argparse.c
```diff
@@ -27,9 +27,10 @@
...
+ *     a new handle from the keywords dictionary, so an HPyTracker pointer
+ *     must be passed and on successful return HPyTracker_Close must be called
+ *     to close any handles that were created.
  *
  * O+ (object) [HPy *]
```
antocuni, line 35: do we want to keep `O+` now that we have a nice way to deal with `O`?

I see that it might be convenient in some cases, but OTOH it makes it harder to read the code ("should I close this handle or not? Oh wait, let's look 50 lines above whether I used O+ or O")

antocuni (commented):
Ok, you convinced me that passing `HPyTracker` as a first (optional) argument is a good idea.
My only remaining doubt is about `O+`. I would like to hear opinions about others before making a call

hpy/devel/src/runtime/argparse.c
hodgestar: It sounds good, but one only needs to supply an HPyTracker when using `O` with `HPyArg_ParseKeywords`. It's not required when for using `O` with `HPyArg_Parse`. One also only supplies a pointer to an HPyTracker -- not an actual HPyTracker (i.e. it is created for you).

Let me see about improving this bit of documentation along the lines you've suggested though.

antocuni: I wonder whether we should require to pass an `HPyTracker` (which will be unused) also in the `HPyArg_Parse` case, for consistency

hodgestar: I thought about that but I'm a little bit -1 on it because it introduces an HPyTracker (and allocating and freeing it) in a case where it's not needed. For the me the current rule is "if an HPyTracker might be required, it will raise an error regardless of the arguments passed if no HPyTracker pointer is given".

hodgestar: New documentation added.

This isn't perfect for now -- I intend to add documentation for `HPyArg_Parse` and `HPyArg_ParseKeywords` themselves in a bit (as part of the "documenting the API" project) and hopefully we can polish things more there (it makes more sense to describe how the parameter ht works than to describe the particular case of how `ht` and `O` interact, I think).

antocuni (approved):
LGTM

PR #116, opened 2020-11-05
Remove HPyDict_GetItem and HPyDict_SetItem.

hodgestar:
As discussed in #75 and in the TODOs around these two functions:

1) We have HPy_GetItem and HPy_SetItem already which perform very similar operations.
2) PyDict_GetItem returns a borrowed reference so we have to change how it works anyway.

Given the above, it seems a net win to remove the extra methods from the HPy API.

One consequence of this change is that PyDict_GetItem silenced any lookup errors, so one has to use HPyErr_Clear to clear any exception raised if one wants to get the same behaviour. Perhaps having to explicitly silencing errors is an advantage too, but it is an extra change people porting their code might have to make.


antocuni:
I am +0.5 about this:

- we should definitely remove these functions

- OTOH, it might complicate porting existing extensions: I fear that many people would just replace `PyDict_GetItem` with `HPy_GetItem` without noticing the subtle change in semantics

- IIRC, during the Leysin sprint we thought about having a header called `hpy/compat.h` (suggestions for a better name are welcome) which contains functions which are not part of the official HPy API but might help porting extensions: e.g. we could implement `HPyDict_GetItem_NewHandle` there.

- I don't know whether we want to go down that route though. I am also happy to merge this PR now and open an issue about `hpy/compat.h` where to discuss whether it's a good idea or not

test/test_hpydict.py
```diff
@@ -33,7 +33,7 @@ def test_New(self):
         """)
         assert mod.f() == {}
 
-    def test_SetItem(self):
+    def test_set_item(self):
```
antocuni, line 36: do we actually need these tests? At a very quick glace it seems to me that they are not testing anything more than the equivalent tests in `test_object.py`.
If the don't, we can just kill :)
If they **do** test more, we should improve `test_object.py` and kill these.

antocuni (approved):
+0.5, as detailed in my other comment

hodgestar:
Re +0.5: Another option is to re-implement HPyDict_GetItem and HPyDict_SetItem in terms of HPy_GetItem and HPy_SetItem and do the renaming that was suggested in the comments I removed.

If we do that, should they go in the runtime? Or somewhere else?

test/test_hpydict.py
hodgestar: They're here to test that one can re-implement the equivalent of PyDict_GetItem and PyDict_SetItem easily. :)

hodgestar:
I'm going to merge this -- I think it's good to keep the API small & it'll be easy enough to add helper functions if there's demand for them later.

antocuni:
> Re +0.5: Another option is to re-implement HPyDict_GetItem and HPyDict_SetItem in terms of HPy_GetItem and HPy_SetItem and do the renaming that was suggested in the comments I removed.

isn't this exactly what I suggested, except that you propose to put them in the main `hpy.h` instead of `hpy/compat.h`?

> If we do that, should they go in the runtime? Or somewhere else?

I think they should be `static inline` functions, so they can stay in the header.

PR #117, opened 2020-11-06
Add microbenchmarks AND speedup HPy types

antocuni:
This PR does three things:

1. Introduce a very simple benchmark runner based on `pytest`: each benchmark is a test using the `timer` fixture. This has many advantages:

      * you can easily run only a subset of benchmarks using `-k` or `-m`

     * it automatically handles exceptions

     * you can easily run the same benchmark against multiple implementations using fixtures (as we do with the `simple` module)

2. Add a number of microbench to measure method calls on modules and types. This is a direct port of my [cpyext-benchmarks](https://github.com/antocuni/cpyext-benchmarks)

3. An important performance fix: the early run of microbenchmarks showed that on CPython, method calls on hpy types were ~20% slower than method calls on Python/C types. This has been fixed by commit 6865973, by enabling `Py_TPFLAGS_HAVE_VERSION_TAG` by default (thanks @hodgestar !)


thedrow:
Why aren't you using pytest-benchmark?

paugier:
piconumpy bench without this PR (with CPython of course):
```
PicoNumpy (CPython C-API)  :  8.75 * norm
PicoNumpy (hpy)            :  9.49 * norm
```
(hpy 8.0 % slower)

And with this PR:
```
PicoNumpy (CPython C-API)  :  8.74 * norm
PicoNumpy (hpy)            :  9.01 * norm
```
(hpy 3.0 % slower)

microbench/test_microbench.py
```diff
@@ -0,0 +1,129 @@
...
+    def test_method_lookup(self, simple, timer, N):
+        obj = simple.Foo()
+        with timer:
+            for i in range(N):
+                # note: here we are NOT calling it, we want to measure just
+                # the lookup
+                obj.noargs
+
+    def test_method_lookup_ht(self, api, simple, timer, N):
+        obj = simple.HTFoo()
```
hodgestar, line 88: Should we perhaps run all benchmarks against both `HTFoo` and `Foo`? Of course in the HPy case these are the same thing, but it would be good to compare with both static and heap types in the old C API.

Simplest option is to copy the tests and have `TestType` and `TestHeapType` . Other option would be to add another parameter to the test case.

hodgestar (approved):
We should clean up the "HTFoo" stuff a bit (i.e. make a complete set of microbenchmarks and explain somewhere what "HT" is and that "HTFoo" and "Foo" are the same for HPy).

pytest-benchmark looks interesting and might save a lot of code, but I guess someone needs to try and see how it works out.

Also happy for this to be merged and I can sort out the details in a follow-up PR.

thedrow:
> pytest-benchmark looks interesting and might save a lot of code, but I guess someone needs to try and see how it works out.

I've used pytest-benchmark before and it works wonderfully.

microbench/test_microbench.py
antocuni: > Simplest option is to copy the tests and have TestType and TestHeapType . Other option would be to add another parameter to the test case.

or to substitute `simple` with a fixture `Foo` which yields both `simple. Foo` and `simple.HTFoo`.

antocuni:
> Why aren't you using pytest-benchmark?

I have used it in the past: it works well but I was not sure it is the best for our use case:

* writing benchmarks is more verbose, since you need to create a function for the test and a callable for the benchmark itself, with more complicated logic if you want to do a setup step which is not measured (you need to write yet another callable which returns something which is passed to the benchmarked function itself, etc.).  I like the `timer` approach: it's very straightforward and it makes it easy to move setup code outside the loop.

* pytest-benchmark has logic to automatically calculate the number of iterations: I see why it's useful but I don't think we want it here

* pytest-benchmark creates a **very** verbose report which I find it hard to read: so even if we use it, we would end up writing our own reporting I think

* pytest-benchmark computes advanced stats like mean, min, stddev, etc. This is actually useful and it's the only advantage I can see. But IMHO not enough to overcome the disadvantages I explained above. If we decide we need  them, they are very easy to compute anyway.

microbench/test_microbench.py
hodgestar: Further parameterizing fixtures is what I did with pytest-benchmark, but it requires rewriting the display logic when using the code in this PR, so I just added TestHeapType for now.

hodgestar (approved):

hodgestar:
@antocuni I pushed a minimal change to add the extra HeapType tests and explain what they are for.

I think if we're going to make this more complicated, we'll want to rethink a bit our solution. pytest-benchmark is more flexible and perhaps better structured in some ways, but I agree it has its limitations. On the other hand, our current solution is hacked together quickly and will likely generate a bunch of work for us if we try do more complicated things with it.

Either way, the microbenchmarks are an awesome addition and will help us avoid silly mistakes, so +5 on this PR as it is.

antocuni:
> I think if we're going to make this more complicated, we'll want to rethink a bit our solution. pytest-benchmark is more flexible and perhaps better structured in some ways, but I agree it has its limitations. On the other hand, our current solution is hacked together quickly and will likely generate a bunch of work for us if we try do more complicated things with it.

I am up to being disproved of course, but my experience tells me that trying to tweak pytest-benchmark to do what I want ends up in *more* work.

> Either way, the microbenchmarks are an awesome addition and will help us avoid silly mistakes, so +5 on this PR as it is.

thanks, merging!

PR #118, opened 2020-11-07
Move HPyErr_ tests into test_hpyerr.

hodgestar:
Seems only logical. :)

This PR also removes an argparse exception message test raised when trying to parse a non-double argument as a double. The error message is different on PyPy and possibly other implementations, which makes testing there harder.


antocuni (approved):
thanks :)

PR #120, opened 2020-11-10
Add remaining numeric types to arg parsing.

hodgestar:
This PR adds support for:

* b (unsigned char)
* B (unsigned char without overflow checking)
* h (short)
* H (unsigned short)
* I (unsigned int)
* l (long)
* L (unsigned long)
* k (long long)
* K (unsigned long long)
* n (HPy_ssize_t)
* f (float)
* p (bool)

It also adds overflow checking to:

* i

To support the code for the new formats it adds the following extra conversion API functions:

* `HPyLong_AsUnsignedLong`
* `HPyLong_AsUnsignedLongMask`
* `HPyLong_AsLongLong`
* `HPyLong_AsUnsignedLongLong`
* `HPyLong_AsUnsignedLongLongMask`
* `HPyLong_AsSize_t`
* `HPyLong_AsSsize_t`

The list of `HPyLong_As` functions should now match the list of `HPyLong_From` functions.

It also adds `OverflowError` which is raised by the integer bounds checking of some formats.


hpy/devel/src/runtime/argparse.c
```diff
@@ -105,15 +144,100 @@ static int
...
+            return 0;
+        *output = (unsigned char) value;
+        break;
+    }
+
+    case 'h': { /* signed short int */
+        short *output = va_arg(*vl, short *);
+        _BREAK_IF_OPTIONAL(current_arg);
+        long value = HPyLong_AsLong(ctx, current_arg);
+        if (value == -1 && HPyErr_Occurred(ctx))
```
thedrow, line 183: Missing braces?

hpy/devel/include/cpython/hpy.h
```diff
@@ -49,10 +49,11 @@ typedef struct _HPyContext_s {
     HPy h_True;
     HPy h_False;
     HPy h_Exception;
-    HPy h_ValueError;
-    HPy h_TypeError;
     HPy h_IndexError;
+    HPy h_OverflowError;
     HPy h_SystemError;
+    HPy h_TypeError;
+    HPy h_ValueError;
```
antocuni, line 56: at some point we should just make a branch to add all exceptions at once.
Currently it is very fragile both here and on PyPy because you need to manually initialize these in the very same order, so whenever we do such a change the risk of breaking things is high.

test/test_argparse.py
```diff
@@ -12,6 +12,10 @@
 class TestParseItem(HPyTest):
     def make_parse_item(self, fmt, type, hpy_converter):
         mod = self.make_module("""
+            #define CHAR_TO_HPYBYTES(ctx, a) ( \
```
antocuni, line 15: just nitpick, but what about `static inline` instead of `#define`?

test/test_hpyerr.py
```diff
@@ -69,3 +69,36 @@ def test_HPyErr_Cleared(self):
         """)
         assert mod.f() is None
         assert sys.exc_info() == (None, None, None)
+
+    def check_exception_constant(self, cls):
```
antocuni, line 73: thank you for adding these tests, we really needed them!
I wonder whether it's better to just have a single test for all of them to save compilation time, but your choice

test/test_hpylong.py
```diff
@@ -1,8 +1,22 @@
 from .support import HPyTest
 
+
 class TestLong(HPyTest):
 
-    def test_long_aslong(self):
+    def test_Long_FromLong(self):
```
antocuni, line 6: I know that you are simply following the existing tests and might be a chance which is worth its own PR, but I think that at some point we should start to test these API more: e.g., what happens if you pass something of the wrong type, an `int` subclass, a class which implements `__int__` and/or `__index__`, etc.

antocuni (approved):
LGTM. I added a few remarks but they are more about future work and/or suggestions for other PRs than changes which are relevant to this.

Bonus point if you implement the new `HPyLong_*` functions also in PyPy :)

test/test_hpyerr.py
hodgestar: I initially did try a single test, but I couldn't think of a good way to do `ctx->h_XXX` (where `XXX` can change) without recompiling.

hpy/devel/src/runtime/argparse.c
hodgestar: If you're referring to the missing `{}` then HPy already uses the pattern without the braces quite a lot for returning errors. I'm happy to change it, but I think it would be good to agree on a standard first.

hpy/devel/include/cpython/hpy.h
hodgestar: Let me do that next and get it out of the way and then it's done.

test/test_argparse.py
hodgestar: Done!

test/test_hpylong.py
hodgestar: For `HPyLong_From...` one can't really pass the wrong thing (because it won't compile) but let me at least add a simple check to each of the `HPyLong_As...` so that we're sure it does the right thing for non-number types.

test/test_hpyerr.py
antocuni: what about using a dict? E.g.:
```c
HPy h_dict = HPyDict_New(ctx);
HPy_SetAttr_s(ctx, h_dict, "ValueError", ctx->h_ValueError);
...
```

Then you can just call `mod.f('ValueError')` and check that it raises. The `HPy_SetAttr_s` lines can be either written manually or autogenerated; my personal preference would be to write them manually, I like to have tests as linear and easy to read as possible.

test/test_hpylong.py
```diff
@@ -91,18 +101,23 @@ def test_Long_FromLongLong(self):
...
+        import pytest
         mod = self.make_module("""
             HPyDef_METH(f, "f", f_impl, HPyFunc_O)
             static HPy f_impl(HPyContext ctx, HPy self, HPy arg)
             {
                 long long a = HPyLong_AsLongLong(ctx, arg);
-                return HPyLong_FromLongLong(ctx, a);
+                if ((a == (long long) -1) && HPyErr_Occurred(ctx))
+                    return HPy_NULL;
+                                return HPyLong_FromLongLong(ctx, a);
```
antocuni, line 112: wrong indentation?

test/test_hpyerr.py
hodgestar: Ah, so type them all into the same module. Let me do that in the PR with all the exceptions I am working on now.

test/test_hpylong.py
hodgestar: Whoops. Will fix.

hodgestar: Fixed in 4a0d4ffa000319.

PR #121, opened 2020-11-11
Add exception and warning constants

hodgestar:
This adds all the exceptions and warnings from Python's pyerrors.h, except:

* EnvironmentError (an alias for OSError)
* IOError (as alias for OSError)
* WindowsError (only defined on Windows and just an alias for OSError)

The EnvironmentError and IOError aliases are tested from Python code.

It also adds:

* HPyErr_SetObject (for raising the created unicode exception instances)

It not longer adds HPy_Call and marks the unicode exception tests as expected failures until HPy_Call or similar exists.


TeamSpen210:
According to [the docs](https://docs.python.org/3/library/exceptions.html#WindowsError), `WindowsError` is just a compatibility alias for `OSError`, so it can probably just be omitted. Maybe `IOError` and `EnvironmentError` can be omitted as well with an explanatory comment.

hpy/tools/autogen/public_api.h
```diff
@@ -140,6 +208,8 @@ int HPy_RichCompareBool(HPyContext ctx, HPy v, HPy w, int op);
 
 HPy_hash_t HPy_Hash(HPyContext ctx, HPy obj);
 
+HPy HPy_Call(HPyContext ctx, HPy callable, HPy args, HPy kwargs);
```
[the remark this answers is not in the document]
hodgestar: Agreed it should be discussed. I like the idea of a signature a bit like our KEYWORDS method signature -- i.e. `*args, nargs, kw`.

Happy to remove from here and start a new PR with what we decide -- perhaps I will just disable the test for the moment and we can enable it when Py_Call arrives.

hodgestar:
@antocuni HPy_Call removed.
@TeamSpen210 EnvironmentError and IOError removed -- thanks for the suggestion!

test/test_hpyerr.py
```diff
@@ -200,6 +200,10 @@ def check_exception(cls):
 
     def test_h_unicode_exceptions(self):
         import pytest
+        pytest.xfail(
```
antocuni, line 236: pypy's apptest don't support xfail. Can we just skip it instead?

antocuni (approved):
LGTM, apart the xfail vs skip

test/test_hpyerr.py
hodgestar: What happens with the `pytest.mark.xfail` in test_importing.py?

antocuni: a completely different thing, in the best pypy tradition :man_facepalming: 
`pytest.mark.*` are handled by pytest at collection time, so xfail works properly.
However, once you are inside the test it's a different story: since it's an apptest, the `pytest` module which is imported inside the test is a dummy one which implements only `skip` and `raises`: see `pypy/tool/pytest/objspace.py:maketestobjspace()` to see where the (dark) magic happens.

So to summarize: I withdraw my suggestion above, you can just use `@pytest.mark.xfail`

hodgestar: Woot. `@pytest.mark.xfail` used instead.

hpy/tools/autogen/public_api.h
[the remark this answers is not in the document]
hodgestar: `HPy_Call` being discussed in https://github.com/hpyproject/hpy/issues/122.

PR #123, opened 2020-11-21
Add API reference documentation

hodgestar:
Preview on readthedocs: https://hpy.readthedocs.io/en/feature-add-api-documentation/api-reference/argument-parsing.html

Here is a screenshot of the results:

image


hodgestar:
That was way too many commits for a tiny fix, but it's working now.

hodgestar:
Ready for review.

docs/conf.py
```diff
@@ -40,6 +46,32 @@
...
+
+
+def setup(app):
+    app.connect("c-autodoc-pre-process", pre_process)
+
+
+if 'READTHEDOCS' in os.environ:
+    from clang import cindex
+    cindex.Config.set_library_file(
+        "/usr/lib/x86_64-linux-gnu/libclang-6.0.so.1"
```
antocuni, line 73: this looks a bit fragile, but I suppose it's the only way to have it working?

hpy/devel/src/runtime/argparse.c
```diff
@@ -53,10 +54,10 @@
  * d (float) [double]
  *     Convert a Python floating point number to a C double.
  *
- * Handles
- * -------
+ * Handles (Python Objects)
+ * ~~~~~~~~~~~~~~~~~~~~~~~~
  *
- * O (object) [HPy *]
+ * O (object) [HPy]
```
antocuni, line 60: is it possible to use `fixed width fonts` for things like the character codes and the types? It is especially important for the character codes because with the default font things like `I` risk to be misinterpreted as `l`

hpy/devel/src/runtime/argparse.c
```diff
@@ -339,6 +346,36 @@ parse_item(HPyContext ctx, HPyTracker *ht, HPy current_arg, int current_arg_tmp,
...
+ *
+ * :returns: 0 on failure, 1 on success.
+ *
+ * If a `NULL` pointer is passed to `ht` and an `HPyTracker` is required by
+ * the format string, an exception will be raised.
+ *
+ * If a pointer is provided to `ht`, the `HPyTracker` will always be created
+ * and must be closed with `HPyTracker_Close` if parsing succeeds (after all
+ * handles returned are no longer needed). If parsing fails, this function
+ * will close the `HPyTracker` automatically.
```
antocuni, line 377: would be nice to have also an "Examples" section

antocuni (approved):
I added a few nitpick comments but in general this looks wonderful, thank you!

docs/conf.py
hodgestar: The full story is a bit involved, but yes.

Right now readthedocs uses Ubuntu 18.04 and only includes llvm 6. sphinx-c-autodoc requires python-clang >= 7 and I've make pip install clang==10.0.1. This looks for libclang-10.so, which it can't find. However, python-clang 10.0.1 seems to work fine with the libclang-6.0.so if one points it directly to the .so like we do here.

The good news is that readthedocs is busy trialing Ubuntu 20.04 for their builds and once that becomes public, we should be able to just delete these lines (or at least it currently works without these lines on my Ubuntu 20.04 machine).

hpy/devel/src/runtime/argparse.c
hodgestar: Good idea & turned out to be quite easy. Fixed.

hpy/devel/src/runtime/argparse.c
hodgestar: Also a good idea. I'll make an attempt at an Examples section in a bit.

hodgestar: Examples added. Adding the example also made me realize we didn't have a test for passing an HPyTracker to HPyArg_Parse.

docs/conf.py
thedrow: In that case, please add a TODO comment so we know when to remove this code.

antocuni (approved):
LGTM

docs/conf.py
hodgestar: TODO added.

Issue #85, opened 2020-09-29
Make universal-ABI modules installable


rlamy:
ATM, there is no way to properly install a module that uses the universal ABI. 
At first glance, `python setup.py --hpy-abi=universal install` looks like it might work. But while it does create a valid `pof.hpy.so` and sticks it in `site-packages/`, the `pof.py` shim that setuptools creates isn't HPy-aware.
In any case, that solution uses the deprecated `.egg` format, and we should rather find something that works with pip.

rlamy:
BTW, replacing setuptools' shim with something like below does work correctly, but I don't know how to tell setuptools to produce it.

```python
class Spec:
    def __init__(self, name, origin):
        self.name = name
        self.origin = origin

def __bootstrap__():
    global __loader__, __file__
    import sys, pkg_resources
    from hpy.universal import load_from_spec
    __file__ = pkg_resources.resource_filename(__name__, 'piconumpy_hpy.hpy.so')
    m = load_from_spec(Spec('piconumpy_hpy', __file__))
    sys.modules[__name__] = m
__bootstrap__()
```

antocuni:
I did a tiny improvement about this in PR #105 : instead of installing something which doesn't work, it complains clearly.

About a more proper solution: automatically generating a shim similar to what `bdist_egg` does might work, but we need to ensure that it works also with `bdist_wheel` (which I think it's what it's used by pip under the hood).

Moreover, we need to think about how this will interact with the debug mode: to load a module in debug mode, we always need the universal binary. So we might want to support an installation/packaging mode in which we build/distribute BOTH versions (CPython and universal) on CPython, but only universal on pypy, and possibly other combinations on other implementations.

So, whatever we decide to do it needs to be well designed, and I think it's better to wait until we have a debug mode to have our ideas cleaner.

hodgestar:
The universal ABI modules have been installable for awhile now (since #124 was merged :).

PR #124, opened 2020-11-26
Make universal extensions installable

hodgestar:
This is an attempt to address #85 and remove many of the caveats in the old hpy.devel distutils hacking by separating the hpy extension building into its own build_hpy_ext command.

What's in the PR:

* Universal modules now get a stub file written (using the example from @rlamy in #85).
* Universal modules now build correctly with `bdist_wheel`.
* `pip install --hpy-abi=universal ...` now works correctly.
* Tests now load universal modules and cpython modules the same way.
* A test for FatalError has been added that loads the test module in a Python subprocess.
* The proof-of-concept test harness now installs cpython and universal modules the same way (with `--hpy-abi=...`)

The pypy MR for these changes is at https://foss.heptapod.net/pypy/pypy/-/merge_requests/783.

### DIFF
--- a/hpy/devel/__init__.py
+++ b/hpy/devel/__init__.py
@@ -94,9 +94,9 @@ class Spec:
 
 
 def __bootstrap__():
-    import os, sys
+    import sys, pkg_resources
     from hpy.universal import load_from_spec
-    ext_filepath = os.path.join(os.path.dirname(__file__), {ext_file!r})
+    ext_filepath = pkg_resources.resource_filename(__name__, {ext_file!r})
     m = load_from_spec(Spec({module_name!r}, ext_filepath))
     m.__file__ = ext_filepath
     sys.modules[__name__] = m
--- a/hpy/devel/__init__.py
+++ b/hpy/devel/__init__.py
@@ -40,6 +40,34 @@ class HPyDevel:
         """
         return list(map(str, self.src_dir.glob('ctx_*.c')))
 
+    def fix_distribution(self, dist):
+        """ Override build_ext to support hpy modules.
+
+            Used from both setup.py and hpy/test.
+        """
+        dist.hpydevel = self
+
+        base_build = dist.cmdclass.get("build", build)
+        base_build_ext = dist.cmdclass.get("build_ext", build_ext)
+
+        class build_hpy_ext(build_hpy_ext_mixin, base_build_ext):
+            _base_build_ext = base_build_ext
+
+        def dist_has_ext_modules(self):
+            if self.ext_modules or self.hpy_ext_modules:
+                return True
+            return False
+
+        def build_has_ext_modules(self):
+            return self.distribution.has_ext_modules()
+
+        # replace build_ext subcommand
+        dist.cmdclass['build_ext'] = build_hpy_ext
+        dist.__class__.has_ext_modules = dist_has_ext_modules
+        base_build.has_ext_modules = build_has_ext_modules
+        idx = [sub[0] for sub in base_build.sub_commands].index("build_ext")
+        base_build.sub_commands[idx] = ("build_ext", build_has_ext_modules)
+
 
 def handle_hpy_ext_modules(dist, attr, hpy_ext_modules):
     """ Distuils hpy_ext_module setup(...) argument and --hpy-abi option.
@@ -49,31 +77,13 @@ def handle_hpy_ext_modules(dist, attr, hpy_ext_modules):
     """
     assert attr == 'hpy_ext_modules'
 
-    base_build = dist.cmdclass.get("build", build)
-    base_build_ext = dist.cmdclass.get("build_ext", build_ext)
-
-    class build_hpy_ext(build_hpy_ext_mixin, base_build_ext):
-        _base_build_ext = base_build_ext
-
-    def dist_has_ext_modules(self):
-        if self.ext_modules or self.hpy_ext_modules:
-            return True
-        return False
-
-    def build_has_ext_modules(self):
-        return self.distribution.has_ext_modules()
-
     # add a global option --hpy-abi to setup.py
     dist.__class__.hpy_abi = 'cpython'
     dist.__class__.global_options += [
         ('hpy-abi=', None, 'Specify the HPy ABI mode (default: cpython)')
     ]
-    # replace build_ext subcommand
-    dist.cmdclass['build_ext'] = build_hpy_ext
-    dist.__class__.has_ext_modules = dist_has_ext_modules
-    base_build.has_ext_modules = build_has_ext_modules
-    idx = [sub[0] for sub in base_build.sub_commands].index("build_ext")
-    base_build.sub_commands[idx] = ("build_ext", build_has_ext_modules)
+    hpydevel = HPyDevel()
+    hpydevel.fix_distribution(dist)
 
 
 _HPY_UNIVERSAL_MODULE_STUB_TEMPLATE = """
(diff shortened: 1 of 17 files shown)


antocuni:
I didn't look at it in details, but just wanted to note that `hpy.devel` is also used by pypy to compile its own hpy tests, so before merging this please check that it doesn't break pypy after you do `./update_vendored.sh` (or, if it breaks, that you have a fix available :))

hpy/devel/__init__.py
```diff
@@ -1,164 +1,210 @@
...
+        HPy extensions.
+    """
+
+    _extensions = None
+
+    @property
+    def extensions(self):
+        return self._extensions
+
+    @extensions.setter
```
antocuni, line 134: obscure. Add a comment explaining why it's needed

hpy/devel/__init__.py
```diff
@@ -1,164 +1,210 @@
...
+    ext_filepath = pkg_resources.resource_filename(__name__, {ext_file!r})
+    m = load_from_spec(Spec({module_name!r}, ext_filepath))
+    m.__file__ = ext_filepath
+    sys.modules[__name__] = m
+
+__bootstrap__()
+"""
+
+
+class HPyExtensionName(str):
```
antocuni, line 117: nice hack ;)
I remember that I and @arigo considered using something like this when we originally wrote this code, but it looked too obscure even for us. But I agree that nowadays it's a better solution than the old `collect_hpy_ext_names`.

It is worth a comment though, especially explaining why you can't just stick an attribute `is_hpy_extension` on the `ext` objects.

Moreover, you might want to consider to add a function `is_hpy_extension` which does `isinstance(HPyExtensionName)`, it looks cleaner to me

antocuni (changes requested):
wow, nice job!
I didn't try things by myself but it looks like this is a BIG improvement over the old situation.
I have added a couple of remarks; most of them are "this code is too obscure, please explain what the hell you are doing",  because it's a bit hard to follow all the jumps if you don't know distutils deeply.

It might be worth adding also a comment describing the "general strategy", e.g.: "the first hook to be called is `handle_hpy_ext_modules`, which fixes the distribution, then at stage X you use `build_hpy_ext` instead of `build_ext`, etc. etc.

test/test_hpyerr.py
```diff
@@ -33,8 +36,15 @@ def test_FatalError(self):
             @EXPORT(f)
             @INIT
         """)
-        # Calling mod.f() gives a fatal error, ending in abort().
-        # How to check that?  For now we just check that the above compiles
+        result = subprocess.run([
```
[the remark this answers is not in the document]
hodgestar: Inside app-level tests the `sys` module has no `executable` attribute, so that provides a convenient way to skip the remainder of the test.

I don't understand app-level tests well enough to know if a better plan can be made -- a separate process probably needs to be involved somehow since although `SIGABRT` can be "caught", restoring the process back to a sane place seems likely to require crazy crazy hacks.

hodgestar:
@antocuni Thank you for the initial review of this obscure hacking. I *think* I've now addressed everything in one way or another. Would you mind looking at it again?

I am still going to make sure it can be made to work on PyPy before merging (now that PyPy is up to date with hpy and this PR is cleaned up a bit).

hodgestar:
@antocuni Ready for re-review.

test/support.py
```diff
@@ -205,29 +205,23 @@ def make_module(self, ExtensionTemplate, main_src, name, extra_sources):
...
-            spec.loader.exec_module(module)
+            sys.path.insert(0, os.path.dirname(mod_filename))
+            module = importlib.import_module(name)
         finally:
-            # the universal module stub .py file places the newly loaded HPy
-            # extension module in sys.modules so we retrieve it from there
-            # (for cpython modules, the module was already placed in
-            # sys.modules earlier in this function)
-            module = sys.modules.pop(name)
+            del sys.path[0]
```
antocuni, line 222: nitpick: if the imported module modifies `sys.path`, this is wrong.
I know that in practice it will never happen because we have control on the test module that we generate, but who knows what someone will write in the future? It is worth at least an `assert sys.path[0] == os.path.dirname(mod_filename)`.

Basically, the thing which I dislike the most is that before this branch importing a test module was "stateless" and self-contained, while now we are depending on global state (either `sys.modules` or importlib's cache), which is often problematic in tests. But I suppose that we can't do much to avoid it if we want to use the `.py` stub :(

antocuni (approved):
@hodgestar LGTM, thanks for incorporating my suggestions.

I added a couple of remarks in the comments about pkg_resources and global state, but feel free to consider them just a rant :)

hodgestar:
@antocuni Thank you for the explanation about app-level tests -- that was useful.

Some responses to other questions (apologies for these not being inline, but GitHub UI didn't make that easy):

* Re `pkg_resources`: I *think* the point of `pkg_resources.resource_filename` is largely to support extension that have been stored in `.zip` files, and perhaps other more obscure use cases I don't know about. The point of this particular stub is that it closely follows the one for `bdist_egg` inside setuptools itself, and I think following a similar approach to that is worthwhile because it's a least a set of weird edge cases that people are used to.

* Re `del sys.path[0]`: Let me add the assert, because it might be very annoying to debug later if the assumption is violated. Another options would be to save and restore all of sys.path (it's a bit hard to see why a compiled test module should expect to be able to modify sys.path at all).

antocuni:
> Re pkg_resources: I think the point of pkg_resources.resource_filename is largely to support extension that have been stored in .zip files, and perhaps other more obscure use cases I don't know about. The point of this particular stub is that it closely follows the one for bdist_egg inside setuptools itself, and I think following a similar approach to that is worthwhile because it's a least a set of weird edge cases that people are used to.

yes, I think the same. But I think that nowadays eggs and zips are mostly unused/unsupported? I think that wheels explicitly requires the installer to unzip the content (e.g., this is the first result for googling [python zipimport wheel](https://mail.python.org/pipermail/distutils-sig/2014-January/023558.html)).

Moreover, so far the only installation method that we test/support is wheels, I think? I we decide we wanto to explicitly support eggs, it might be a good idea to add a test for that. I'm not saying that we should drop the call to pkg_resources or deliberately break eggs. I just want to make sure that either choice it's a deliberate design decision and not an accident.

That said, I am fine to merge this as is and think about this issue in another PR

> Re del sys.path[0]: Let me add the assert, because it might be very annoying to debug later if the assumption is violated. 

thank you, +1

> Another options would be to save and restore all of sys.path (it's a bit hard to see why a compiled test module should expect to be able to modify sys.path at all).

lesson learned in 14 years of pypy: people's code will always be weirder than you think, even if you take this lesson into account ;)

hodgestar:
@antocuni Thank you for the reviews (again)!

FFY00:
@hodgestar is there any specific reason you moved back to `pkg_resources` (https://github.com/hpyproject/hpy/pull/124/commits/7f83e39d1e5b192ba395d554d31f723f545720c0) in this PR? This adds a runtime dependency on `setuptools`. `setuptools` used to be virtually present on all Python environments during package building, and after as a byproduct, but this is not the case anymore with PEP 517. So, I was wondering if there was any technical reason for the change, I can't find anything.

hodgestar:
@FFY00 -- my reasoning is documented in a comment earlier in this PR -- https://github.com/hpyproject/hpy/pull/124#issuecomment-741716495

If we only want to support Python 3.7+, then we could use `importlib.resources.path` instead. Ideally we will eventually add an import hook that allows us to avoid this stub loader entirely.

PR #125, opened 2020-11-29
start to support windows

mattip:
~it passes one test, but fails to link on others.~ Solved. 
- ~I think a libraries argument is needed on windows to make the linking happy~ Solved
- ~There is a problem with the `%TEMP%` path used in pytest, see numpy/numpy#17658~ Solved

Support windows in build and tests


hpy/universal/src/ctx_meth.c
```diff
@@ -23,33 +23,36 @@ ctx_CallRealFunctionFromTrampoline(HPyContext ctx, HPyFunc_Signature sig,
         HPyFunc_varargs f = (HPyFunc_varargs)func;
         _HPyFunc_args_VARARGS *a = (_HPyFunc_args_VARARGS*)args;
         Py_ssize_t nargs = PyTuple_GET_SIZE(a->args);
-        HPy h_args[nargs * sizeof(HPy)];
+        HPy *h_args = (HPy *)malloc(nargs * sizeof(HPy));
```
antocuni, line 26: this is "wrong": we are using variable-length arrays to allocate the space on the stack, which is faster than `malloc`. It's a C99 feature, so I am surprised that VC++ doesn't support it: maybe you need to pass some flag?

If we can't use it, we need to go back using `alloca`. See also PR #76, which removed calls to `alloca` in favor of variable-length arrays

test/test_argparse.py
```diff
@@ -12,7 +12,10 @@
 class TestParseItem(HPyTest):
     def make_parse_item(self, fmt, type, hpy_converter):
         mod = self.make_module("""
-            __attribute__((unused)) static inline
```
antocuni, line 15: maybe we should hide this difference behind a macro and put it in `hpy.h`. It looks like a generally useful feature.

hpy/universal/src/ctx_meth.c
PMeira: As a random person watching this repo, please allow me to share some info on this.

From [this post](https://devblogs.microsoft.com/cppblog/c11-and-c17-standard-support-arriving-in-msvc/#variable-length-arrays), the latest minor versions of Visual Studio added support for C11 and C17 but VLAs are not supported, intentionally. Here's an excerpt:

> ### Variable Length Arrays
> Astute readers will note that VLAs are also not supported. Variable length arrays are generally less efficient than comparable fixed sized arrays, and generally inefficient compared to equivalent `malloc()`, when safely and securely implemented. VLAs provide attack vectors comparable to those of the infamous `gets()` — deprecated and destined to removal — for opportunities of “shifting the stack” and other exploits. For these reasons we intend not to support VLAs as an optional feature in C11.

Note that it now supports C89 + some extensions, C11 and C17, but full C99 support was never there. Some more info here: https://docs.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=msvc-160#c-standards-support-1

As someone that has built some C libraries with MSVC in the past decade, lack of VLAs, headers like `stdint.h`, and variables declared in the middle of a scope are common issues. As long as you don't need to support Python 2.7 on Windows (MSVC 2008), things should be smoother nowadays, VLAs being a notable exception.

arigo: Well, then I stand by my comment in PR #76.  Let's stick to `alloca`.  It seems to be present on every platform where `cffi` compiles (there's some careful includes needed, tough).

antocuni: thank you for the explanation @pmeira!

> Well, then I stand by my comment in PR #76. Let's stick to alloca. It seems to be present on every platform where cffi compiles (there's some careful includes needed, tough).

+1 for reverting PR #76 and go back using `alloca`.

thedrow: I was completely unaware of that problem.
We'll need to suppress that error in cppcheck as well.

Ping me when you revert, I'll try to do it.

thedrow: It turns out that `__STDC_NO_VLA__` should be set to 1 if the C compiler does not support VLA.
We could fallback to alloca if we want to 🤷‍♂️.

antocuni: I think it would just make the code unnecessarily more complicated. If we need to use `alloca` on windows, better to just use it everywhere and forget VLAs

mattip:
I wonder what construct MSVC will support instead of `HPy items[] = {};`, which it errors out on.
```
// args is null, but kw is not, so we need to create an empty args tuple
// for CPython's PyObject_Call
HPy items[] = {};
HPy empty_tuple = HPyTuple_FromArray(ctx, items, 0);
```

test/test_argparse.py
mattip: but hpy.h is not imported where the macro is needed

mattip:
Happy for help if someone sees how to fix the int/long/long long failures. The current proof-of-concept failure is due to trying to set up a virtualenv where one already exists.

mattip:
It turns out the proof-of-concept run was confusing various virtual environments since it used `VENV=...` without declaring that a local variable. 

This is evident when running the `wheel()` function: it first sets `VENV`, then calls `_build_wheel`. This sets `VENV` (overriding the value previously set), creates a virtualenv, and installs hpy and builds the poc wheel. Then `_build_wheel` exits. Now the wheel function creates a virtualenv, activates it, installs the poc wheel, and run tests. Note it did **not** install hpy into the virtualenv. Before this PR, that worked because the value of `VENV` set in `_build_wheel` changed the value set before the call to `_build_wheel`. Now that the `VENV` is local, running tests failed because `hpy` was not installed into the last virtualenv. 

In 26ef55b I added a call to _install_hpy to fix this.

test/test_argparse.py
antocuni: yes, it's included automatically by `make_module`:
https://github.com/hpyproject/hpy/blob/d5c8442beae23c85d606f00e0f60ce816b1447dc/test/support.py#L56-L61

azure-pipelines.yml
```diff
@@ -57,24 +60,27 @@ jobs:
...
+        imageName: 'windows-latest'
   steps:
   - template: azure-templates/ccache.yml
     parameters:
       pythonVersion: $(python.version)
   - template: azure-templates/python.yml
     parameters:
       pythonVersion: $(python.version)
-  - script: proof-of-concept/test_pof.sh wheel cpython
+  - bash: proof-of-concept/test_pof.sh wheel cpython
```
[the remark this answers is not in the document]
mattip: they are interchangeable on linux. On windows, bash is required as far as I know.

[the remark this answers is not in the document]
mattip: otherwise I wouldn't have done it in d758015

stonebig:
can it be part of upcoming PyPy-7.3.5 64bit for Windows ?

hodgestar:
> Can it be part of upcoming PyPy-7.3.5 64bit for Windows ? It could help for your May 11th [PyCon Langage Summit ](https://us.pycon.org/2021/summits/language/)presentation

PyPy implements its own `hpy.universal` support so Windows support for HPy on PyPy is separate to the Windows support for HPy on CPython in this PR. I don't know offhand what the plans are for PyPy 7.3.5.

hpy/debug/src/debug_ctx.c
```diff
@@ -55,20 +58,31 @@ HPyContext * hpy_debug_get_ctx(HPyContext *uctx)
 // this function is supposed to be called from gdb: it tries to determine
 // whether a handle is universal or debug by looking at the last bit
 extern struct _HPyContext_s g_universal_ctx;
-__attribute__((unused)) static void hpy_magic_dump(HPy h)
+#ifndef _MSC_VER
+__attribute__((unused))
+#endif
```
hodgestar, line 63: We should define `HPy_UNUSED` in `macros.h` and then use that everywhere.

hpy/debug/src/include/hpy_debug.h
```diff
@@ -29,7 +29,18 @@ HPyContext * hpy_debug_get_ctx(HPyContext *uctx);
 // this is the HPy init function created by HPy_MODINIT. In CPython's version
 // of hpy.universal the code is embedded inside the extension, so we can call
 // this function directly instead of dlopen it. This is similar to what
-// CPython does for its own built-in modules
+// CPython does for its own built-in modules. But we must use the same
+// signature as HPy_MODINIT
+
+// Copied from Python's exports.h
+#ifndef Py_EXPORTED_SYMBOL
```
hodgestar, line 36: Should we call this `HPy_EXPORTED_SYMBOL` and add it to `macros.h`?

hpy/devel/include/common/hpytype.h
```diff
@@ -2,6 +2,11 @@
 #define HPY_UNIVERSAL_HPYTYPE_H
 
 #include <stdbool.h>
+#ifdef __GNUC__
+#define HPyAPI_UNUSED __attribute__((unused)) static inline
```
hodgestar, line 6: Perhaps this should also be moved to `macros.h`.

hodgestar (commented):
These are the issues that I think are still outstanding:

hodgestar:
@mattip @antocuni I'm also happy to merge this PR and address the minor issues in a follow up PR.

mattip:
Merging now and fixing up the other issues in subsequent PRs sounds good to me. This PR is already quite large.

hodgestar (approved):

hodgestar:
Merged!

stonebig:
There is no Windows wheel on pypi for hpy-0.9.0
Is it to happen only for 1.0 ?

mattip:
The windows wheels are [explicity excluded](https://github.com/hpyproject/hpy/blob/f6114734b4a5d03cfb9d24d3e8c71f77e8803881/.github/workflows/release-pypi.yml#L59). Will this github workflow run without a tag, on a PR to debug the windows builds?

fangerer:
Sorry for my late answer. When we've migrated to GitHub Workflows, building wheels for Windows using `cibuildwheel` did not work. Since then, we actually never tested. However, since we run tests on Windows as well, it should in theory work. If that's still something people are interested in, I can give it a try.

stonebig:
> Sorry for my late answer. When we've migrated to GitHub Workflows, building wheels for Windows using `cibuildwheel` did not work. Since then, we actually never tested. However, since we run tests on Windows as well, it should in theory work. If that's still something people are interested in, I can give it a try.

if hpy is to become a PEP, it would make sense to at least check it can generate a Windows wheel

PR #126, opened 2020-12-01
More tests about exceptions

antocuni:
Add two tests which were failing on PyPy, testing that the functions return the correct error_value in case of error


test/test_hpyerr.py
```diff
@@ -341,3 +341,49 @@ def check_warning(cls):
...
+                res = HPy_TrueDivide(ctx, a, b);
+
+                // the point of the test is to check that in case of error
+                // HPy_Div returns HPy_NULL
+                if (HPy_IsNull(res)) {
+                    HPyErr_Clear(ctx);
+                    return HPyLong_FromLong(ctx, -42);
+                }
+
+                HPy_Close(ctx, b);
```
hodgestar, line 364: `b` should not be closed here.

hodgestar:
Otherwise looks good!

test/test_hpyerr.py
antocuni: uh, true! We need the debug mode :facepalm:

antocuni: fixed

hodgestar (approved):

PR #128, opened 2020-12-03
Add tests for converting objects with __int__ and __index__ to integers.

hodgestar:
Note that support for `__int__` in these methods is deprecated since Python 3.8. Since deprecated is not the same as gone, I've kept the tests for them, even though they display the deprecation warnings.

We could suppress the warnings, but I'm not sure PyPy's pytest supports that feature.

### DIFF
--- a/test/test_hpylong.py
+++ b/test/test_hpylong.py
@@ -21,6 +21,15 @@ class TestLong(HPyTest):
                 return v
         return MagicIndex()
 
+    def python_supports_magic_index(self):
+        """ Return True if the Python version is 3.8 or later and thus
+            should support calling __index__ in the various HPyLong_As...
+            methods.
+        """
+        import sys
+        vi = sys.version_info
+        return (vi.major > 3 or (vi.major == 3 and vi.minor >= 8))
+
     def test_Long_FromLong(self):
         mod = self.make_module("""
             HPyDef_METH(f, "f", f_impl, HPyFunc_NOARGS)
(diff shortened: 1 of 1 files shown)


hodgestar:
Update: Support for `__index__` was only added in Python 3.8. Perhaps I should guard the test with a Python version check? Or just not test it for now?

test/test_hpylong.py
```diff
@@ -3,6 +3,33 @@
 
 class TestLong(HPyTest):
 
+    def magic_int(self, v):
```
antocuni, line 6: I fear that defining classes here (outside the test functions) won't work for pypy apptests. It might be easier either to:

1. copy&paste the magic classes inside each test

2. add a single test which tests the behavior for all the various `HPyLong_As*` functions

antocuni (commented):
> Update: Support for __index__ was only added in Python 3.8. Perhaps I should guard the test with a Python version check? Or just not test it for now?

Good question. I have no idea how HPy should handle cases like this in which CPython changes behavior over time. Maybe just checking the python version as you did is the best thing to do, but I'd like to hear opinions from others

test/test_hpylong.py
hodgestar: Are these really outside of the test functions? These methods are only called from within test functions.

antocuni: yes, but they are defined outside. PyPy's applevel tests only see the content of the single method, and the "self" that they receive is not an instance of `TestLong`, it's some fake object. And they don't get the same `globals()` also, which is the reason why you need to put `import pytest` inside the method.

There are ways to stick arbitrary objects into the app-level version of `self` (see e.g. `self.w_make_module = self.space.wrap(interp2app(descr_make_module))` inside `support.py`) but they are complex and very hard to integrate with hpy's test suite, which is supposed to run also outside pypy.

hodgestar: Ah, I did not know that `self` was a fake object (despite having recently edited the bit of `support.py` which sets up the fake object!). I will make a plan.

hodgestar: Good news -- I ran these in pypy as app-level tests and they just work?

hodgestar:
@antocuni Since this is working, I am going to merge this and continue getting PyPy working on the latest master. If there are more issues I'm happy to address them later. Hoping that is okay by you!

PR #131, opened 2020-12-08
Use apt-get to install cppcheck 1.9 -- the apt package is old, but works.

hodgestar:
The cppcheck snap package was recently updated and it is broken. It feels better to just use the apt package which is more stable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ cppcheck-build-dir:
 	mkdir -p $(or ${CPPCHECK_BUILD_DIR}, .cppcheck)
 
 cppcheck: cppcheck-build-dir
-	cppcheck --error-exitcode=1 --cppcheck-build-dir=$(or ${CPPCHECK_BUILD_DIR}, .cppcheck) --enable=warning,performance,portability,information,unusedFunction,missingInclude --report-progress -I hpy/devel/include/ -I hpy/devel/include/common/ -I hpy/devel/include/cpython/ -I hpy/devel/include/universal/ -I hpy/universal/src/ --suppress=nullPointerRedundantCheck --force -D NULL=0 .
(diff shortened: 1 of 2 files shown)


thedrow:
You're going to lose a lot of the newer checks which is kind of a bummer.
I'm not sure why this fails as cppcheck is working just fine for me.

hodgestar:
> You're going to lose a lot of the newer checks which is kind of a bummer.
> I'm not sure why this fails as cppcheck is working just fine for me.

The snap package failed to install `std.cfg` in the right place. It is a bummer to lose the new checks but I don't have a better idea. Any suggestions?

thedrow:
An edge version was released today. It'll be promoted soon, I believe.
Keep this PR open for a few days. Let's see what happens.

thedrow:
I think @ColinIanKing is the snap package's maintainer.
I hope he's aware of the issue.

hodgestar:
I a bit -1 on going back to the snap because the version changes all the time and we have no way to pin the version, so we will keep having to debug issues as new versions are released.

hodgestar:
I'm trying out the new snap release in #132 to see what happens.

hodgestar:
Closing this in favour of #132. We can always resurrect this PR if the cppcheck snap continues to annoy us.

hodgestar:
The cppcheck snap has disappeared from snapcraft.io, so I'm going to bring this PR back.

It's a shame to lose the new tests, but the tests also have to work reliably.

ColinIanKing:
Sorry about this, but I don't have the time to keep the snap fresh.

hodgestar:
> Sorry about this, but I don't have the time to keep the snap fresh.

No worries! Thank you for maintaining it for awhile. It was awesome to get new packages for free while it lasted. Keep well -- it's a crazy world out there.

PR #133, opened 2020-12-10
Preserve the process environment in the subprocess Python call in the HPy_FatalError test.

hodgestar:
The test works without this in simple cases (e.g. current HPy tests on CPython) but fails in more complex cases (e.g. pypy translated tests). Preserving the environment as part of the call is probably a good idea regardless.


hodgestar:
Merging this minor fix.

PR #134, opened 2020-12-10
Add a .translate(...) wrapper to HPyExtensionName.

hodgestar:
The `.translate(...)` method is used by the version of distutils shipped with Python 2.7, and PyPy's app-level tests run the compilation under 2.7, so we need to support that here.

See https://foss.heptapod.net/pypy/pypy/-/merge_requests/783#note_149515 for where this was discovered.

### DIFF
--- a/hpy/devel/__init__.py
+++ b/hpy/devel/__init__.py
@@ -134,6 +134,10 @@ class HPyExtensionName(str):
         result = str.split(self, *args, **kw)
         return [self.__class__(s) for s in result]
 
+    def translate(self, *args, **kw):
+        result = str.translate(self, *args, **kw)
+        return self.__class__(result)
+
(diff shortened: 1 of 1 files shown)


hodgestar:
Merging this small fix so that I can update pypy.

antocuni:
this precise PR is good, but I am a bit worried that this approach is a bit fragile: what happens if a newer version of the py3.x distutils start using yet another method?
What about putting as `assert` somewhere which would have failed without this change? At least we will get a clean error instead of some obscure behavior

hodgestar:
Perhaps if there are *only* HPy extensions in the list we could assert that every call to one of the extension name methods should receive an HPyExtensionName? This would at least raise errors during tests.

PR #135, opened 2020-12-10
Add simple detection of missed HPy extension names for the common case when there are only HPy extensions

hodgestar:
We'd like to catch issues in the HPy extension name detection and raise good error messages rather than end up with obscure bugs.

See #134 for where `.translate(...)` was added. We almost missed spotting that case in PyPy and there might be others lurking out there in other versions of setuptools / distutils.

### DIFF
--- a/hpy/devel/__init__.py
+++ b/hpy/devel/__init__.py
@@ -150,6 +150,14 @@ def remember_hpy_extension(f):
     """
     @functools.wraps(f)
     def wrapper(self, ext_name):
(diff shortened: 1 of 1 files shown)


antocuni (commented):
it's a hack, but I fear that we can't do much better than this.
Maybe put a comment to explain is `_only_hpy_extension` is used only as a sanity check, and that we need it because if we also have non-hpy extensions we are basically lost.

hodgestar:
Comment added. Merging.

PR #136, opened 2020-12-11
Reactivate the importing test.

hodgestar:


hpy/devel/__init__.py
```diff
@@ -112,6 +112,9 @@ def __bootstrap__():
     ext_filepath = pkg_resources.resource_filename(__name__, {ext_file!r})
     m = load_from_spec(Spec({module_name!r}, ext_filepath))
     m.__file__ = ext_filepath
+    m.__loader__ = __loader__
```
hodgestar, line 115: I'm not 100% sure that copy the loader and spec are the perfect thing to do, but:

1. These are the loader and spec that were used to import the module, so they're correct in that sense.
2. There isn't another loader and spec than can be used to import an hpy.universal extension.

test/test_importing.py
```diff
@@ -1,31 +1,20 @@
...
 
 class TestImporting(HPyTest):
 
-    @pytest_collecting.mark.xfail
     def test_importing_attributes(self):
+        import pytest
         import sys
-        modname = 'mytest'
-        so_filename = self.compile_module("""
+        if not hasattr(sys, "executable"):
```
hodgestar, line 9: Is is to check whether we are inside app-level tests or not. I should probably factor this out into a small helper function but I'm not particular sure what to call it. `is_app_level` is one idea, but perhaps that's too PyPy specific?

test/test_importing.py
```diff
@@ -1,31 +1,20 @@
...
+            pytest.skip()
+        mod = self.make_module("""
```
hodgestar, line 11: `make_module` now does a proper import, so we don't need to do anything different in this test to import the module normally.

test/test_importing.py
antocuni: ideally, the HPy tests should be independent of the implementation. But practically speaking, being able to do `if is_pypy_apptest` will probably be very useful.
What about introducing a generic way to check the implementation? E.g., we could have `self.implementation` which by default is `sys.implementation.name` (so, `'cpython'`, `'pypy'`, etc.), but pypy's `test/support.py` could override it and set to `'pypy_apptest'`. And we could also add a `self.is_pypy` which does `return self.implementation in ('pypy', 'pypy_apptest')` maybe?

hodgestar: I added some more methods that describe the possible capabilities of the test runner that the hpy tests currently care about. I dabbled with making something more generic (e.g. a `.capabilities(...)` method that returns a set of supported behaviours) but that was actually more code and less clear than the simple methods, so I went back to those.

hodgestar:
@antocuni Ready for a next review, I think.

test/test_importing.py
antocuni: I fear that we are stumbling across each other feet in this PR and #142, sorry for that :(.

I strongly believe that `make_module` should NOT do a proper import. Basically, the biggest difference between loading a module in memory and "properly importing" it is that the latter modifies some global state (e.g. `sys.modules`) plus it has interactions with importlib caches, etc.. And for tests, the least we mess with global state, the better it is. 
E.g. for testing the debug mode, we need to load the same module twice (with and without the debug mode support), so if we do "proper imports" we need to un-import the module between tests etc. It's a mess. That's why `self.make_module` was originally written like that, it even had a comment to make this behavior explicit ;)

See also commit ba33924: I believe that the code written this way is much cleaner and doesn't need all the hacks to sys.path, sys.modules, etc.

That said, we surely need a way to properly import modules of course, but it's probably better/easier to leave it clearly separate from the low-level stuff used by tests. So I propose the following:

1. `hpy.universal.load` is a very low-level function which just loads the module in memory, without "importing" it
2. we write a helper (maybe called `hpy.importlib` or something like that) which uses `hpy.universal.load` under the hood
3. `hpy.importlib` can have its own tests of course: these can be written by calling `self.compile_module` + `hpy.importlib.do_proper_import` instead of relying on `self.make_module` to do everything
4. the .py stubs can call `hpy.importlib` to do whatever is appropriate

What do you think about this solution?

hodgestar:
@antocuni I am happy for us to go back to `test_support` doing a lower-level import that avoids updating sys.modules, etc, but as a practical matter could we perhaps merge this and then maybe sort things out in your PR since this one is about re-activating the import tests.

Probably we need the current import code in `support.py` to move to `test_import.py` and be replaced with the new code in your debug branch.

antocuni:
> @antocuni I am happy for us to go back to test_support doing a lower-level import that avoids updating sys.modules, etc, but as a practical matter could we perhaps merge this and then maybe sort things out in your PR since this one is about re-activating the import tests.

sounds like a good idea, +1

hodgestar:
Thanks!

PR #139, opened 2020-12-14
Run test_basic before all the others

antocuni:
I find it useful because you can quickly catch dumb regressions.


hodgestar:
I'm a bit -1 on including this tiny hack / weirdness when one can just run `pytest test/test_basic.py`.

antocuni:
> I'm a bit -1 on including this tiny hack / weirdness when one can just run pytest test/test_basic.py.

it happens often to me that after doing a change I want to run all the tests to ensure I didn't break anything, and it `pytest` starts with `test_argparse`, which takes ~40 seconds on my machine. Then it run `test_basic` and it quickly discover a regression.
My point is that there is no point in running all the other tests if `test_basic` fails, and as such it should run first. We use this hack often in pypy, see e.g. `rpython/jit/metainterp/test/test_ajit.py` or all the various `test_ztranslation`.

hodgestar:
@antocuni I thought about this while walking this evening and I think I would be happy if we renamed `test_basic.py` in a way that made it a bit clearer that it's a set of tests that attempts to ensure that things are sane enough for other tests to run. What about something like `test_00_sanity_check.py` ?

hodgestar:
Maybe we could also add a docstring at the top of the file to explain what the tests in it are for.

antocuni:
> @antocuni I thought about this while walking this evening and I think I would be happy if we renamed `test_basic.py` in a way that made it a bit clearer that it's a set of tests that attempts to ensure that things are sane enough for other tests to run. What about something like `test_00_sanity_check.py` ?

personally I think that `test_basic` already conveys that meaning :)

```
ba•sic /ˈbeɪsɪk/  
adj. 
[before a noun]of, relating to, or forming a base or basis;
fundamental:
a basic principle.
```

If you don't like `test_basic`, what about `test_fundamental`?
That said, it's not worth spending too much time discussing on such a tiny issue, so I'm fine to just close this PR if there isn't consensus

hodgestar:
I think let's merge it. We can make a better name later if it confuses people other than me. :)

PR #141, opened 2020-12-16
Include a reference to the file which generates autogen files

antocuni:
This makes it much easier to understand what to edit if you want to change an autogen file, especially for newcomers


hodgestar (approved):
Nice!

PR #142, opened 2020-12-16
Add the debug mode

### LINKED ISSUE #155: hpy.universal leak handles and it's unnecessary slow
The title mentions two different problems but I am grouping them in the same issue since I think they can be solved together.
I am already working on a fix in PR #142 but I thought it was useful to report my findings here, for the future.

First problem: consider `ctx_CallRealFunctionFromTrampoline`:
https://github.com/hpyproject/hpy/blob/fdc6047e6db755ecaeded1f805b91823eb11f08d/hpy/universal/src/ctx_meth.c#L6-L15

in CPython-ABI mode this is not a problem because `_py2h` and `_h2py` are just no-op casts, but in in universal mode, `_py2h` allocates new handles which are never freed:
https://github.com/hpyproject/hpy/blob/fdc6047e6db755ecaeded1f805b91823eb11f08d/hpy/universal/src/handles.c#L137-L152

The proper solution is something like this:
```c
    switch (sig) {
    case HPyFunc_NOARGS: {
        HPyFunc_noargs f = (HPyFunc_noargs)func;
        _HPyFunc_args_NOARGS *a = (_HPyFunc_args_NOARGS*)args;
        HPy h0 = _py2h(a->self);
        a->result = _h2py(f(ctx, h0));
        _hclose_nodecref(h0);
        return;
    }
```

However, there is a better solution: the second problem is that the current implementation of handles is unnecessarily slow. I tried the ujson and piconumpy benchmarks:
- ujson: the universal ABI is ~15% slower than the CPython ABI
- piconumpy: the universal ABI is ~35% slower than the CPython ABI

In PR #142 I am experimenting with a different approach for `hpy.universal`. In particular, `_py2h` and `_h2py` are implemented like this:
```c
// The main reason for +1/-1 is to make sure that if people casts HPy to
// PyObject* directly, things explode.

static inline HPy _py2h(PyObject *obj) {
    if (obj == NULL)
        return HPy_NULL;
    return (HPy){(HPy_ssize_t)(obj + 1)};
}

static inline PyObject *_h2py(HPy h) {
    if HPy_IsNull(h)
        return NULL;
    return (PyObject *)(h._i - 1);
}
```

So, they are basically no-op casts again, and the benchmarks are much faster:
- ujson: the universal ABI is ~0.9% slower than the CPython ABI
- piconumpy: the universal ABI is ~6% slower than the CPython ABI

**Historical note**: why do we represent hpy.universal handles as indexes in a list? The original ideas was to support the debug mode, so that we could easily store extra debugging infos for each handles.

The idea that I am trying in PR #142 is different, i.e. to wrap a generic universal ctx into a debug ctx: so, debug handles are wrapper around generic opaque universal handles and the extra infos can be attached directly on the wrappers. Moreover, by doing that we pay the overhead of "heavy" handles only for the modules for which the debug mode is enabled.

**EDIT**: fixed the PR number

antocuni:
This PR introduces the debug mode. It contains many different pieces which are all connected together, so let me try to summarize:

- the debug mode is implemented as a wrapper around a generic `HPyContext`: handles are represented by `struct DebugHandle` which at the moment contains just a reference to the underlying `h` and a couple of extra fields. In the future, we can put more stuff in it, for example a C stack trace to remember who created the handle

- I have killed `hpy/universal/src/handles.c`. `hpy.universal` now no longer uses small integers to represent handles, but it uses directly `PyObject*`, although with a bit of masking/unmasking (see the comments in the code). The original reason to represent handles as indexes into an array was to support debug mode, but now that the bug mode is a fully wrapper it's no longer needed. Moreover, it makes the non-debug universal mode slightly faster and fixes #155 

- the code which implements the debug wrapper is in `hpy/debug/src/debug_ctx_*.c`: these files are compiled together with `hpy.universal` for simplicity, but in theory are fully generic. I expect that e.g. PyPy should be able to vendor these files as they are

- to load a module in debug mode, you use `hpy.universal.load(..., debug=True)`. I am perfectly aware that this is not user friendly nor the API that we want at the end. The idea is that PR introduces a low-level interface, and then later we will build a high-level, user-friendly interface and/or importing mechanism on top of it. For example,I can imagine having an env variable `HPY_DEBUG=foo,bar` to signal that we want to use the debug mode for those two modules, or maybe a configuration file, etc.

- `hpy/debug/src/_debugmod.c` implements the app-level interface to the debug mode; in particular, it exports an applevel `DebugHandle` type a functions like `get_open_handles()`. The nice thing is that `_debugmod.c` is implemented in HPy itself :). For simplicity it is *also* bundled and compiled together with `hpy.universal`, and it is exposed as `hpy.universal._debug`. Again, I expect it to work unmodified on PyPy. 

- `hpy/debug/*.py`: this is a higher level API implemented on top of `_debug`. In particular, it implements the context manager `LeakDetector`.

- all tests are now run three times: `[cpython]`, `[universal]` and `[debug]`.

- I addedd `ctx->h_NotImplemented` (because I needed it) and `ctx->h_Ellipsis` (because it was the only remaining one)

There are tons of stuff to make the debug mode better, but I think it's best to do them in subsequent PRs. Incomplete list of next steps:

1. add a way to check whether we are using a closed handle
2. implement additional sanity check for many functions
3. make it possible to enable the debug mode without using `hpy.universal.load` and integrate this mechanism with the stub .py files
4. improve the pytest fixtures so that it automatically checks for leaks when we execute a `[debug]` test, and make sure that all the existing tests pass


Makefile
```diff
@@ -15,7 +19,7 @@ cppcheck: cppcheck-build-dir
 
 infer:
 	python3 setup.py build_ext -if -U NDEBUG | compiledb
-	@infer --fail-on-issue --compilation-database compile_commands.json
+	@infer --fail-on-issue --compilation-database compile_commands.json --report-blacklist-path-regex "hpy/debug/src/debug_ctx.c"
```
hodgestar, line 22: Should we perhaps leave a comment on why `debug_ctx.c` had to be blacklisted?

hodgestar:
Question for when we implement `HPY_DEBUG=1` or similar: We should probably make sure not to load `hpy.universal._debugmod` with a debug context, and maybe other third party libraries would like the option to not be loaded with debug either?

hpy/debug/src/_debugmod.c
```diff
@@ -0,0 +1,212 @@
...
+#include "debug_internal.h"
+
+static UHPy new_DebugHandleObj(HPyContext uctx, UHPy u_DebugHandleType,
+                               DebugHandle *handle);
+
+
+HPyDef_METH(new_generation, "new_generation", new_generation_impl, HPyFunc_NOARGS)
+static UHPy new_generation_impl(HPyContext uctx, UHPy self)
+{
+    HPyContext dctx = hpy_debug_get_ctx(uctx);
```
hodgestar, line 17: Obtaining a new debug ctx from the `uctx` requires that the `uctx` be the same in this module as in the module being debugged, yes? Which means there essentially needs to be only one `ctx` per interpreter? This is already true in the CPython and PyPy (I think) implementations, but it's not currently *required* by HPy.

I'm not quite sure how to remove this constraint right now -- the ctx is fairly invisible from Python code, and that likely needs to stay that way.

hpy/debug/src/debug_internal.h
```diff
@@ -0,0 +1,113 @@
...
+
+typedef struct {
+    long magic_number; // used just for sanity checks
+    HPyContext uctx;
+    long current_generation;
+    DebugHandle *open_handles;   // linked list
+    //DebugHandle *closed_handles; // linked list
+} HPyDebugInfo;
+
+static inline HPyDebugInfo *get_info(HPyContext dctx)
```
hodgestar, line 105: Perhaps this function should be called DHPy_get_info?

fangerer (approved):
Looks good to me. Thank you!

hpy/debug/src/_debugmod.c
```diff
@@ -0,0 +1,212 @@
...
+    .flags = HPy_TPFLAGS_DEFAULT,
+    .defines = DebugHandleType_defs,
+};
+
+
+static UHPy new_DebugHandleObj(HPyContext uctx, UHPy u_DebugHandleType,
+                               DebugHandle *handle)
+{
+    DebugHandleObject *dhobj;
+    UHPy u_result = HPy_New(uctx, u_DebugHandleType, &dhobj);
```
hodgestar, line 176: Would it make sense to put the DebugHandle type on the context like we do for other build-in types like List?

hpy/debug/src/_debugmod.c
```diff
@@ -0,0 +1,212 @@
...
+   call get_open_handles. This way, we save 16 bytes per DebugHandle.
+
+   This means that you can have different DebugHandleObjects wrapping the same
+   DebugHandle. To make it easier to compare them, they expose the .id
+   attribute, which is the address of the wrapped DebugHandle. Also,
+   DebugHandleObjects compare equal if their .id is equal.
+*/
+
+typedef struct {
+    HPyObject_HEAD
```
hodgestar, line 85: Could we make this a pure type once the cast support branch lands?

hpy/debug/src/_debugmod.c
hodgestar: I guess this is also a good illustration of the slight hopping one has to do at the moment since global type objects are gone -- and the hopping is actually not too bad in simple cases.

hpy/debug/src/_debugmod.c
```diff
@@ -0,0 +1,212 @@
...
+HPyDef_SLOT(DebugHandle_repr, DebugHandle_repr_impl, HPy_tp_repr)
+static UHPy DebugHandle_repr_impl(HPyContext uctx, UHPy self)
+{
+    DebugHandleObject *dh = HPy_CAST(uctx, DebugHandleObject, self);
+    UHPy uh_fmt = HPy_NULL;
+    UHPy uh_id = HPy_NULL;
+    UHPy uh_args = HPy_NULL;
+    UHPy uh_result = HPy_NULL;
+
+    uh_fmt = HPyUnicode_FromString(uctx, "<DebugHandle 0x%x for %r>");
```
hodgestar, line 133: We could do with `PyUnicode_FromFormat` here.

hpy/debug/src/debug_ctx.c
```diff
@@ -0,0 +1,121 @@
...
+    HPyDebugInfo *info = malloc(sizeof(HPyDebugInfo));
+    if (info == NULL) {
+        HPyErr_NoMemory(uctx);
+        return -1;
+    }
+    info->magic_number = HPY_DEBUG_MAGIC;
+    info->uctx = uctx;
+    info->current_generation = 0;
+    info->open_handles = NULL;
+    //info->closed_handles = NULL;
```
hodgestar, line 34: Is there a plan to used `closed_handles` later?

hpy/debug/src/debug_ctx_cpython.c
```diff
@@ -0,0 +1,121 @@
...
+        DHPy_close(dctx, dh_arg);
+        DHPy_close(dctx, dh_result);
+        return;
+    }
+    case HPyFunc_VARARGS: {
+        HPyFunc_varargs f = (HPyFunc_varargs)func;
+        _HPyFunc_args_VARARGS *a = (_HPyFunc_args_VARARGS*)args;
+        DHPy dh_self = _py2dh(dctx, a->self);
+        Py_ssize_t nargs = PyTuple_GET_SIZE(a->args);
+        DHPy dh_args[nargs * sizeof(DHPy)];
```
hodgestar, line 70: Should we add a comment here about removing the use VLAs later?

hodgestar: Maybe it doesn't matter in this file?

hpy/tools/autogen/debug.py
```diff
@@ -0,0 +1,164 @@
...
+    def gen_debug_wrapper(self, func):
+        if func.name in self.NO_WRAPPER:
+            return
+        #
+        assert not func.is_varargs()
+        node = get_debug_wrapper_node(func)
+        signature = toC(node)
+        rettype = toC(node.type.type)
+        #
+        def get_params():
```
hodgestar, line 89: Is there a reason `get_params` is a nested function?

hpy/universal/src/hpymodule.c
```diff
@@ -115,49 +115,174 @@ static PyObject *load_from_spec(PyObject *self, PyObject *spec)
...
+static void init_universal_ctx(HPyContext ctx)
+{
+    if (!HPy_IsNull(ctx->h_None))
+        // already initialized
+        return;
+
+    // XXX this code is basically the same as found in cpython/hpy.h. We
+    // should probably share and/or autogenerate both versions
+    /* Constants */
+    ctx->h_None = _py2h(Py_None);
```
hodgestar, line 201: I'm not sure I understand why the ctx initialization is here inside hpymodule. Could you explain a bit?

hodgestar (approved):
This looks great. Woot.

I left a few comments, although they were largely me coming to grips with the code. Maybe they will result in some small things we want to change now, or some issues we want to write for the future, but I'm also happy for this PR to be merged and for us to work on things in smaller PRs from here.

Makefile
antocuni: good idea. The full explanation is in the log message of cd8cd6e, I added a comment which redirects to it

hpy/debug/src/_debugmod.c
antocuni: I tried to explain more what are the implicit and explicit assumptions in commit 6bad6c3. If it's still unclear, please comment/review/reword directly on PR #164

hpy/debug/src/debug_internal.h
antocuni: maybe, but note that it's a `static inline` in a private header, so it's not exposed anywhere else. I preferred to keep it short because it's used a bit everywhere

hpy/debug/src/_debugmod.c
antocuni: > Would it make sense to put the DebugHandle type on the context like we do for other build-in types like List?

no, because it's not a builtin. Why should it be different than, say, `ndarray`?

> I guess this is also a good illustration of the slight hopping one has to do at the moment since global type objects are gone -- and the hopping is actually not too bad in simple cases.

exactly, that's the point. We need a way to have per-module "globals" at the C level, which we still don't have :(

hpy/debug/src/_debugmod.c
antocuni: yes!

hpy/debug/src/_debugmod.c
antocuni: good point, although we don't have `HPyUnicode_FromFormat` here. And we don't want to use the `Py_*` version because this is supposed to be a pure-hpy module, remember it will be used also by pypy and hopefully graalpython.
I addedd an XXX in e2b548a

hpy/debug/src/debug_ctx.c
antocuni: yes. It will be used to detect when you try to use a handle which has already been closed

hpy/debug/src/debug_ctx_cpython.c
antocuni: I think it matters, because VLAs seems to be unsupported by VC++.
Added comments in 3756927

hpy/debug/src/_debugmod.c
hodgestar: Correct. HPyObject_HEAD is already gone in the cast support branch.

hpy/tools/autogen/debug.py
antocuni: I just liked it that way. It's basically an alternative way to group a bit of logic together instead of putting a comment `compute the params`.
And putting it outside seems overkill

hpy/universal/src/hpymodule.c
antocuni: before this PR the fields were statically initialized to things like `CONSTANT_H_NONE`, `CONSTANT_H_TRUE` etc, which were small integers defined in an enum. And the old `handles.c` made sure to allocate the corresponding handle at the "right place". So this code is the equivalent of what was inside the old `allocate_more_handles`.

It would be nice to statically initialize things like this:
```c
struct _HPyContext_s g_universal_ctx = {
    ...,
    h_None = _py2h(Py_None),
```

but I don't think it is legal to call `_py2h` in a static initializer

antocuni:
This PR is already merged, so I am addressing the comments in #164.

> Question for when we implement HPY_DEBUG=1 or similar: We should probably make sure not to load hpy.universal._debugmod with a debug context, and maybe other third party libraries would like the option to not be loaded with debug either?

remind that `_debugmod` is "statically loaded" by `hpy.universal` itself, so it is its responsibility to ensure that it's initialized with the "correct" context. But add a sanity check looks like a good idea, I added one in commit 8c492e8.

Issue #140, opened 2020-12-14
setup.py --hpy-abi=universal build_ext --inplace doesn't work


antocuni:
As the title says:

```
$ cd proof-of-concept
$ rm build -rf
$ python3 setup.py --hpy-abi=universal build_ext --inplace
...
writing hpy universal stub loader for pofpackage.foo to build/lib
copying build/lib/pof.hpy.so -> 
writing hpy universal stub loader for pof to build/lib
error: build/lib/pof.py already exists! Please delete.
```

cc @hodgestar

antocuni:
I also tried to comment out the "please delete" check in `hpy/devel/__init__.py`, but things still don't work:

```diff
diff --git a/hpy/devel/__init__.py b/hpy/devel/__init__.py
index 33be1a0..c232029 100644
--- a/hpy/devel/__init__.py
+++ b/hpy/devel/__init__.py
@@ -240,8 +240,8 @@ class build_hpy_ext_mixin:
             ext._full_name, output_dir)
         stub_file = (os.path.join(output_dir, *ext._full_name.split('.')) +
                      '.py')
-        if compile and os.path.exists(stub_file):
-            raise DistutilsError(stub_file + " already exists! Please delete.")
+        ## if compile and os.path.exists(stub_file):
+        ##     raise DistutilsError(stub_file + " already exists! Please delete.")
         ext_file = os.path.basename(ext._file_name)
         module_name = ext_file.split(".")[0]
         if not self.dry_run:
```

```
$ python3 setup.py --hpy-abi=universal build_ext --inplace
...
copying build/lib/pof.hpy.so -> 
writing hpy universal stub loader for pof to build/lib
copying build/lib/pofpackage/foo.hpy.so -> pofpackage
writing hpy universal stub loader for pofpackage.foo to build/lib

$ python3 -c 'import pof'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'pof'
```

It seems that the stub `pof.py` is not copied to the inplace directory. If I copy it manually, it works.

hodgestar:
Fixed in #143.

PR #143, opened 2020-12-18
Correctly handle the case where .write_stub is called during build_ext --inplace.

hodgestar:
This addresses #140.

### DIFF
--- a/test/support.py
+++ b/test/support.py
@@ -273,13 +273,15 @@ def _build(tmpdir, ext, hpy_devel, hpy_abi, compiler_verbose=0, debug=None):
     #
     dist = Distribution()
     dist.parse_config_files()
-    options = dist.get_option_dict('build_ext')
     if debug is None:
         debug = sys.flags.debug
-    options['debug'] = ('ffiplatform', debug)
-    options['force'] = ('ffiplatform', True)
-    options['build_lib'] = ('ffiplatform', tmpdir)
-    options['build_temp'] = ('ffiplatform', tmpdir)
+    options_build_ext = dist.get_option_dict('build_ext')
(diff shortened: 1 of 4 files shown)


test/support.py
```diff
@@ -273,13 +273,15 @@ def _build(tmpdir, ext, hpy_devel, hpy_abi, compiler_verbose=0, debug=None):
...
-    options['force'] = ('ffiplatform', True)
-    options['build_lib'] = ('ffiplatform', tmpdir)
-    options['build_temp'] = ('ffiplatform', tmpdir)
+    options_build_ext = dist.get_option_dict('build_ext')
+    options_build_ext['debug'] = ('ffiplatform', debug)
+    options_build_ext['force'] = ('ffiplatform', True)
+    options_build_ext['build_lib'] = ('ffiplatform', tmpdir)
+    options_build_ext['build_temp'] = ('ffiplatform', tmpdir)
+    options_build_py = dist.get_option_dict('build_py')
+    options_build_py['build_lib'] = ('ffiplatform', tmpdir)
```
hodgestar, line 284: So the build_ext command reads the .build_lib value from the build_py command when calling its own .write_stub method. Why this is the case I have no idea.

This was a pre-existing bug that existed because we used to overwrite the output_dir in .write_stub to be .build_lib but now we no longer do that.

antocuni: I think that this code was originally copied&pasted directly from cffi (which is why it says `ffiplatform`). I don't know the details of that it does, maybe @arigo remembers better.

antocuni (approved):
thank you for doing this! I confirm that it works for me, so +1 for merging.

Issue #122, opened 2020-11-20
Calling Python functions from HPy


hodgestar:
We need to pick some APIs to provide for calling Python functions from HPy.

The existing C API has a large zoo of methods for making functions calls.  The zoo is documented at https://docs.python.org/3/c-api/call.html#object-calling-api.

Under the hood there are two main calling conventions  -- `tp_call` and `vectorcall`. I propose that we largely ignore these and instead focus on what fits well with the Python language syntax and what is convenient to call from C.

## Most Pythonic

Of the zoo, `PyObject_Call` most clearly matches a generic `f(*args, **kw)` in Python, and I propose that we adapt is as follows for HPy:

```c
/* Call a callable Python object, with arguments given by the tuple args, and named arguments
    given by the dictionary kwargs.

    ctx: The HPy execution context.
    args: May be HPy_NULL if no positional arguments need to be passed. Otherwise it should be
             a handle to a tuple.
    kwargs: May be HPy_NULL if no keywords arguments need to be passed. Otherwise it should be
                 a handle to a dictionary of keyword arguments.

    Return the result of the call on success, or raise an exception and return HPy_NULL on failure.

    This is the equivalent of the Python expression: callable(*args, **kwargs).
 */
HPy HPy_Call(HPyContext ctx, HPy callable, HPy args, HPy kwargs);    
```

Note that this differs from `PyObject_Call` in that `args` may be `NULL`. I'm not sure why this was a requirement in the existing API.

A some point in the future we might want to implement Py_BuildValue to allow tuples of C values to be constructed easily (and maybe even something similar for easily constructing dictionaries from C values).

## Most C-like (Most Scenic? :)

`PyObject_VectorcallDict` closely matches the signature we chose for `HPyFunc_KEYWORDS` methods, but while this is convenient for receiving arguments from Python, I'm not convinced it's a great fit for calling Python functions from C because one has to construct a dictionary of keyword arguments.

`PyObject_Vectorcall` takes only the names of the keyword arguments (as a tuple), which seems slightly more convenient.

All of the vectorcall methods have the strange behaviour that `nargs` also functions as a flag selecting whether `args[-1]` (or `args[0]` for `PyObject_VectorcallMethod`) may temporarily be overwritten by the called function. I propose that we ensure we do NOT copy this behaviour.

The best suggestion I have at the moment for a convenient C-like calling convention is:

```c
/* Call a callable Python object with positional and keyword arguments given by a C arrays.

    ctx: The HPy execution context.
[...]

PR #147, opened 2020-12-21
Add HPy_CallTupleDict,  HPyCallable_Check and HPyTuple_Check.

hodgestar:
Implements the simpler pythonic calling API for #122.


hpy/tools/autogen/public_api.h
```diff
@@ -232,9 +237,6 @@ int HPyList_Append(HPyContext ctx, HPy h_list, HPy h_item);
 int HPyDict_Check(HPyContext ctx, HPy h);
 HPy HPyDict_New(HPyContext ctx);
 
-/* pyerrors.h */
```
hodgestar, line 235: The pyerrors.h functions were accidentally split into two sections, so I combined those sections.

hodgestar:
Ready for review.

test/test_call.py
```diff
@@ -0,0 +1,93 @@
...
+                return HPy_Call(ctx, f, f_args, f_kw);
+            }
+            @EXPORT(call)
+            @INIT
+        """)
+
+        def f(a, b):
+            return a + b
+
+        assert mod.call(f, (1, 2)) == 3
```
antocuni, line 27: maybe we should also test what happens if you pass something which is not a tuple and/or a dict?

antocuni (commented):
the PR looks good.
I'm not 100% sure that we can to call it `HPy_Call`, though. On CPython it is called that way because it matches the signature of `tp_call`, but in HPy we will probably end up with a different signature, so it will be a bit weird.

One option could be to call it `HPy_LegacyCall` maybe? This would leave `HPy_Call` free for whatever signature we think it's more appropriate later.

test/test_call.py
hodgestar: Good idea. I'll add some.

hodgestar:
@antocuni I don't mind a different name, but:
* I'm against "Legacy" because I don't know if it really is legacy or not. My guess is that it will continue to be fine to use it for a long time.
* I like "HPy_Call" because it maps obviously to "PyObject_Call" which it replaces, so it's for people porting code to know what to use and how it works. A later "HPy_Call" that does things differently to "PyObject_Call" would add to the confusion.
I don't have another good name to suggest really, so I'm a bit stuck on this point. Any other suggestions? If it's not going to be the same name as "_Call", I think the name should try describe the method.

antocuni:
I agree that no solutin is perfect, it's the usual tradeoff between keeping compatibility with the old API and designing a better one.
Let me try to explain why I think this case is "more special" than others; we have a general guideline than HPy_Foo should map precisely to Py_Foo so I agree we should do the same for HPy_Call. However, there are at least two use cases which comes to my mind:

1. call something with the same arguments that we just received from the caller
2. call something with a tuple which we just created

Your version of `HPy_Call` addresses point 2, but doesn't help with point 1. So people who are porting code doing 1 will be annoyed/confused anyway because simply adding the `H` in front of the call doesn't work. OTOH, if we make `HPy_Call` the equivalent of `PyObject_VectorCall`, point 1 will work out of the box, because this is what they receive from the caller.

As for point 2, I claim we should discourage it :). In most cases creating a tuple just to call a function is inefficient and unnecessary: on PyPy, we just don't need the tuple at all and CPython is migrating toward vector call and thus more native functions support the faster calling convention. I don't know about GraalPython, maybe @fangerer or @timfel have opinions?
Moreover, creating the tuple itself is different in HPy since we have TupleBuilder now: so people might end up doing *more* work to use "your version" of HPy_Call, just because it was named so conveniently :)

Of course we should support a way to call things "the old way", that's why I proposed `HPy_LegacyCall`. But it should be a conscious decision, not something which is done "by chance" only because it was very convenient to add the H in front of it.

I tried to do a quick grep on the numpy source code to see how it is used in the wild. Excluding the cython-generated sources, there are 13 usages of `PyObject_Call`: 7 of them are doing what I described in point 1, and 6 are creating new tuples. So, whatever decision we take we are going to break ~50% of the usages anyway.

hodgestar:
For clarity, are you proposing to have the names be `HPy_CallLegacy` and `HPy_CallObjectLegacy`? Or something else?

TeamSpen210:
What about `HPy_CallTuple()` and `HPy_CallTupleDict()`? That makes the requirement for a tuple/dict explicit, and leaves `HPy_Call()` open for an optimised/preferred calling convention later.

antocuni:
> For clarity, are you proposing to have the names be HPy_CallLegacy and HPy_CallObjectLegacy? Or something else?

CPython implements `PyObject_CallObject` as `PyObject_Call(..., NULL)`, so it might be fine to only add the second as a "native" HPy API function. If we really want, we can provide a small helper to implement the equivalent of `PyObject_CallObject` which lives just in a .h as a static inline function (IIRC we have already discussed the idea of having a header which contains implementation of various functions which are not part of the official API but that we want to provide to make porting easier).

So, let's implement only the equivalent of `PyObject_Call`. Possible names:
1. `HPy_LegacyCall`
2. `HPy_CallLegacy`
3. `HPy_CallTupleDict`

I don't have any strong preference. `HPy_CallTupleDict` is kind of nice because it's very explicit.

hodgestar:
> So, let's implement only the equivalent of `PyObject_Call`.

Note that in the old C API, `PyObject_CallObject` allows `args`) to be `NULL`, but `PyObject_Call` does not. I could work around that in `HPy_CallTupleDict` by creating an empty tuple if needed?

timfel:
> HPy_CallTupleDict is kind of nice because it's very explicit.

I like explicit, too. In general we would of course like to avoid creating the tuple for calls if we can. When not running in ABI mode (i.e., everything is compiled and run as bitcode and jitted on Graal) we can sometimes remove the tuple allocation if we're lucky, but a) that's more work for the optimizer and b) it's better if ABI mode doesn't incur a performance hit on GraalPython that would push users towards compiling per VM yet again.

test/test_call.py
hodgestar: The answer is that C Python segfaults in some cases and works fine in others (e.g. a list of args). I've added tests and changed the HPy implementation to check the types of handles rather than segfault.

hodgestar:
I've unified things into a single `HPy_CallTupleDict`. The API seems nice now, but I'm not sure I'm wild about the amount of custom C code in `HPy_CallTupleDict` or that it calls back to multiple different C Python API methods.

 Thoughts?

test/test_call.py
antocuni: I wonder if doing these extra check costs any performance. Do you feel like adding microbenchmarks?
If they are costly, we could do the check only in debug mode.

hodgestar: I'm guessing not a whole lot, but I'll add the microbenchmarks.

antocuni:
> I've unified things into a single `HPy_CallTupleDict`. The API seems nice now, but I'm not sure I'm wild about the amount of custom C code in `HPy_CallTupleDict` or that it calls back to multiple different C Python API methods.

I *think* that having CallTupleDict which maps to either Call or CallObject is fine, provided that it doesn't cost any performance on CPython.
It looks like an improvement from the API point of view and it should be very easy for people to port their code. Maybe add also a note to `docs/porting-guide.rst`? The document is very rough at the moment and we should do it properly, but it's good to start collecting notes while we implement things.

hpy/devel/src/runtime/ctx_call.c
```diff
@@ -0,0 +1,38 @@
...
+    }
+    if (HPy_IsNull(kw)) {
+        obj = PyObject_CallObject(_h2py(callable), _h2py(args));
+    }
+    else if (!HPy_IsNull(args)){
+        obj = PyObject_Call(_h2py(callable), _h2py(args), _h2py(kw));
+    }
+    else {
+        // args is null, but kw is not, so we need to create an empty args tuple
+        // for CPython's PyObject_Call
```
antocuni, line 31: ah, nice catch. I didn't know that CPython doesn't have a way to pass only kwargs. I don't know if it will be actually useful but it doesn't cost anything to add support for it, so it's probably a good idea to be more complete, +1

test/test_call.py
```diff
@@ -0,0 +1,109 @@
+from .support import HPyTest
+
+
+class TestCall(HPyTest):
+    def argument_combinations(self, *items):
```
antocuni, line 5: good idea! Maybe write it to take `**items` instead of `*items`? It would make the calls much nicer to read, and and I think that on modern Python the order of items in the dict is guaranteed, isn't it?

antocuni (commented):
I like it :), see also my comments above

test/test_call.py
hodgestar: I think it is. Let me try that.

hodgestar: Woot. Implemented. Thank you for the suggestion.

test/test_call.py
hodgestar: Microbenchmarks added.

hodgestar:
Porting guide updated.

hodgestar:
@antocuni Ready for a (hopefully) final review.

antocuni (approved):
Thank you, I think it's ready to merge!
For the records, I tried to run the microbench on bencher7, here are the results or a single run (I ran it few times and didn't see any noticeable difference):
```
                                                     cpy                    hpy
                                        ----------------    -------------------
TestModule::test_call_with_tuple              1144.95 us      1146.30 us [1.00]
TestModule::test_call_with_tuple_and_dict     1909.51 us      1884.72 us [0.99]
```

hodgestar:
Thanks for reviewing! Your benchmark numbers look very similar to those on my laptop, where there is also no difference visible across several runs.

PR #132, opened 2020-12-08
Install cppcheck from latest/edge.

hodgestar:
Try out the new cppcheck release.

PR #151, opened 2021-01-07
Install cppcheck via apt.

hodgestar:
Unfortunately the cppcheck snap has broken again. This time it is just completely gone from snapcraft.io. Given that we have limited time to spend fixing build issues, I think it's an okay trade-off to get older less comprehensive cppcheck tests in favour of not having to fix the installation of cppcheck every month. Maybe in awhile the Ubuntu repositories will get a newer version of cppcheck.

See #131 and #132 for previous issues.


hodgestar:
Merging to make builds work again. Happy to implement a different solution later if a good one is suggested.

Issue #160, opened 2021-01-28
Equivalent of PyObject_TypeCheck


antocuni:
This is the signature of `PyObject_TypeCheck`:
```c
int PyObject_TypeCheck(PyObject *o, PyTypeObject *type)
// Return true if the object o is of type type or a subtype of type. Both parameters must be non-NULL.
```

The obvious HPy equivalent would be:
```c
int HPy_TypeCheck(HPy o, HPy type)
```

But note that `type` is now a generic `HPy` instead of a specific `PyTypeObject *`. However, this poses some design issues:
1. Should we do an extra check to ensure that `type` is actually a type object?
2. If yes, how do we signal the failure? Returning -1 would work, but it's too easy to forget to check if people are used to the existing semantics. Moreover, if you forget to check for errors, it's very likely that you will interpret the result as `true`, with all kinds of funny results.
3. Moreover, the extra check could introduce an unnecessary slowdown compared to CPython.

Note that there is also `PyObject_IsInstance` which has the "right" signature but has a much more complex logic.

Possible solutions:
1. don't do the check. It's user's responsibility to ensure that what you pass is a `type`, similar to how they already need to ensure that you don't pass `NULL`. We can introduce the extra checks in debug mode, though.
2. do the check, but don't raise an exception and use `HPy_FatalError` instead
3. change the name to something else: `HPy_CheckType`? `HPy_TryTypeCheck`? `HPy_TypeCheckMaybeFail`? 

I think my preferred solution is (1), especially considered what is the typical usage of the API. I suspect that the in the vast majority of cases it is used to define `PyXXX_Check` for custom types. For example, numpy does the following:
```c
#define PyArray_Check(op) PyObject_TypeCheck(op, &PyArray_Type)
```

So from this POV, the probability of passing something which is not a type is very low, and a fatal error in debug mode could be enough.

Moreover, there is also an additional alternative which is a much bigger redesign, linked to [this comment on #83](https://github.com/hpyproject/hpy/issues/83#issuecomment-698469987): when you define a custom type, you almost always need helper functions such as `XXX_Check`, `XXX_Cast`, `XXX_TryCast`, etc.
If we go in that direction we can introduce a macro which automatically defines all the`XXX_*` helpers: `XXX_Check` will use `_HPy_TypeCheck` which can be made semi-private because it will not be needed anywhere else.

hodgestar:
In HPy one can't do something like:
```
```

antocuni:
[...]

PR #161, opened 2021-01-28
introduce HPy_Type and tentatively HPy_TypeCheck

antocuni:
`HPy_Type` is smooth, but the precise semantics of `HPy_TypeCheck` are still to be decided, see #160:
this PR tentatively implements solution (2): do the check but raise a fatal error instead of a proper exception.

I need both functions to progress with PR #142, that's why I opened a PR before we reached consensus on how to solve the issue. I propose to merge it as is, and revisit the semantics and/or change the name later when we make a decision.


hodgestar (approved):
Looks good to me!

hodgestar:
@antocuni The infer test is failing here too. I am checking why.

hodgestar:
Ah:

PR #163, opened 2021-01-29
Implement buffer protocol

rlamy:
Add slots HPy_bf_getbuffer and HPy_bf_releasebuffer to allow HPy types to implement the buffer protocol. Note that before 3.9, this isn't supported by PyType_FromSpec, so we need to manually create a PyBufferProcs struct and stick it on the type object, after calling PyType_FromSpecWithBases.

Since the Py_buffer struct contains a PyObject* member, we need our own version, with the PyObject* replaced by an HPy (which is wrong BTW, but is the best we can do until #9 is resolved). That unfortunately requires a lot of mostly pointless copying back and forth between Py_buffer and HPy_buffer in the universal ABI. We can avoid the copies in the cpython ABI thanks to a rather dodgy cast...


hpy/devel/src/runtime/ctx_type.c
```diff
@@ -310,6 +317,36 @@ create_slot_defs(HPyType_Spec *hpyspec)
...
+    PyBufferProcs *buffer_procs = NULL;
+    if (hpyspec->defines != NULL) {
+        for (int i = 0; hpyspec->defines[i] != NULL; i++) {
+            HPyDef *src = hpyspec->defines[i];
+            if (src->kind != HPyDef_Kind_Slot)
+                continue;
+            switch (src->slot.slot) {
+                case HPy_bf_getbuffer:
+                    if (buffer_procs == NULL) {
+                        buffer_procs = PyMem_Calloc(1, sizeof(PyBufferProcs));
```
antocuni, line 332: you should check that `PyMem_Calloc` didn't fail

antocuni: moreover, it's a bit sad that you need to duplicate the code for allocating `PyBufferProcs`. Maybe you can detect in advance whether it's needed by tweaking `HPyDef_count`?

hpy/devel/src/runtime/ctx_type.c
```diff
@@ -310,6 +317,36 @@ create_slot_defs(HPyType_Spec *hpyspec)
...
+                    }
+                    buffer_procs->bf_releasebuffer = src->slot.cpy_trampoline;
+                    break;
+                default:
+                    break;
+            }
+        }
+    }
+    return buffer_procs;
+}
```
antocuni, line 348: I seem to understand that this function is only needed on py<=3.8. It is worth a comment explaining why, and a note saying that we can kill it once we support only >=3.9

hpy/devel/include/cpython/hpyfunc_trampolines.h
```diff
@@ -69,4 +69,17 @@
         return _h2py(IMPL(_HPyGetContext(), _py2h(self), _py2h(obj), op)); \
     }
 
+/* With the cpython ABI, the implicit casts from Py_buffer* to HPy_buffer*
+ * happen to Just Work, which makes things easier. */
+#define _HPyFunc_TRAMPOLINE_HPyFunc_GETBUFFERPROC(SYM, IMPL) \
+    static int SYM(cpy_PyObject *arg0, HPy_buffer *arg1, int arg2) \
+    { \
+        return (IMPL(_HPyGetContext(), _py2h(arg0), arg1, arg2)); \
+    }
```
antocuni, line 78: I fear that casting `Py_Buffer*` to `HPy_Buffer*` is undefined behavior :(, although I'm not sure how to fix.
Also, I think that the signature of `SYM` should take a `Py_Buffer*`?
And probably it's worth putting an explicit case to `HPy_Buffer*` in the call to `IMPL`.

hpy/universal/src/ctx_meth.c
```diff
@@ -52,6 +83,25 @@ ctx_CallRealFunctionFromTrampoline(HPyContext ctx, HPyFunc_Signature sig,
...
        return;
     }
+    case HPyFunc_GETBUFFERPROC: {
+        HPyFunc_getbufferproc f = (HPyFunc_getbufferproc)func;
+        _HPyFunc_args_GETBUFFERPROC *a = (_HPyFunc_args_GETBUFFERPROC*)args;
+        HPy_buffer hbuf;
+        a->result = f(ctx, _py2h(a->arg0), &hbuf, a->arg2);
+        _buffer_h2py(ctx, &hbuf, a->arg1);
+        HPy_Close(ctx, hbuf.obj);
+        return;
```
antocuni, line 97: it's a bit annoying that you need to copy all these fields every time you call the impl. We should remind to check whether it's a performance problem.
A possible hack would be to just cast `Py_Buffer` to `HPy_Buffer` and fix the `obj` before and after the call. This is probably undefined behavior again, but we are already relying on it (see my previous commnet)

antocuni (commented):
LGTM overall. However, I'm not so sure about `obj` field.
The CPython docs for Py_Buffer say that obj is supposed to be a `void *`. I propose to:
1. make `HPy_Buffer.obj` a `void *`
2. introduce `HPyBuffer_GetObject` (or similar name) which returns a `HPy`.

This looks saner and it has also the nice property to hide under the cover the fact that for now we use a `HPy` to store obj, while we wait for #9 .

hpy/devel/include/cpython/hpyfunc_trampolines.h
rlamy: Yes, it's UB, but I think it should work everywhere.
Good point about making the cast explicit.

antocuni: I think it's worth putting a comment explaining that we know it's UB

hodgestar:
Could we drop `obj` entirely (or make it a dummy value) and rely on `internal` to opaquely store a reference to whatever handles or other things an implementation needs? This will complicate passing things back and forth a bit, but will be a good idea in the long run, I think.

hodgestar:
> Could we drop `obj` entirely (or make it a dummy value) and rely on `internal` to opaquely store a reference to whatever handles or other things an implementation needs? This will complicate passing things back and forth a bit, but will be a good idea in the long run, I think.

Storing the CPython buffer inside `internal` might also avoid having to update and fix the HPy buffer on each call since one could pass `view->internal->cpython_view` directly to the underlying CPython API methods.

hpy/devel/include/common/hpyfunc.h
```diff
@@ -70,6 +72,19 @@ typedef enum {
     enum { SYM##_sig = SIG };              \
     _HPyFunc_TRAMPOLINE_##SIG(SYM, IMPL)
 
+typedef struct {
+    void *buf;
+    HPy obj;        /* owned reference */
```
hodgestar, line 77: I think it would be better to remove `obj` from `HPy_buffer` and store it in `internal` for the CPython HPy implementations. This avoids the interface needing to be fixed when we figure out how to deal with long-lived references to objects and hides an implementation detail.

I'm not actually sure why `obj` exists on CPython -- I suspect it is implementation detail that was exposed for no good reason and is there only to allow the reference to the object to be DECREF'ed when the buffer is released.

hpy/devel/include/cpython/hpyfunc_trampolines.h
```diff
@@ -69,4 +69,17 @@
         return _h2py(IMPL(_HPyGetContext(), _py2h(self), _py2h(obj), op)); \
     }
 
+/* With the cpython ABI, the casts from Py_buffer* to HPy_buffer*
+ * happen to Just Work, which makes things easier. */
+#define _HPyFunc_TRAMPOLINE_HPyFunc_GETBUFFERPROC(SYM, IMPL) \
+    static int SYM(PyObject *arg0, Py_buffer *arg1, int arg2) \
+    { \
+        return (IMPL(_HPyGetContext(), _py2h(arg0), (HPy_buffer*)arg1, arg2)); \
```
hodgestar, line 77: Rather than casting the HPy_buffer directly to a CPython one, I would rather we kept the CPython buffer separate and used `internal` to map between the two. Otherwise there is a risk of accidentally exposing things we don't want to expose (e.g. the CPython `obj`).

hodgestar:
An issue with my suggestion is that `internal` is meant to be for the code implementing `HPy_bf_getbuffer` and `HPy_bf_releasebuffer` and not for the implementation itself. Perhaps we can get around that by keeping `obj` but renaming it to `_obj` and making it even more opaque -- i.e. declaring it a `void *` that each Python implementation may use as it wishes.

Maintaining the mapping between the CPython and HPy buffers might require a small list of objects like we have for handles currently (or more likely for the long lived object references we plan to implement in future).

hodgestar:
I don't want my comments to necessarily hold up merging this PR, so if the PR is merged with them unaddressed, I'm happy to write them up in a GitHub issue. Perhaps just renaming `obj` to `_obj` or something different to `obj` is enough to allow us to fix things later.

rlamy:
AFAICT, the `obj` field in `Py_buffer` is mostly used by `PyBuffer_Release()` to find which bf_releasebuffer slot to call and to decref the object. We don't necessarily need to stick it in `hpy_buf->obj`, but `getbuffer_impl` must somehow return it to the interpreter, because it's up to the slot implementation to decide which Python object the buffer is attached to (it's usually `self` but not always).

I don't think we can use `internal` for that because it's supposed to be under the full control of the extension. However, we could probably use the return value (so we'd have `typedef HPy (*HPyFunc_getbufferproc)(HPyContext, HPy, HPy_buffer *, int);`). That would be a bit different from CPython though, and it would make it impossible to create a Py_buffer with .obj == NULL (the docs advise against that, but it's technically possible).

hpy/devel/src/runtime/ctx_type.c
rlamy: Error checking added.
I tried to find a way to factor out the allocation of PyBufferProcs, but everything I thought of just made things more complicated. Considering that there will only ever be 2 copies of that code, I'd rather leave it as is.

rlamy:
After some discussions in the dev call yesterday, we concluded that, at least in the context of the object buffer protocol, `HPy_buffer` can be seen as a short-lived struct that bundles together various return values from bf_getbuffer, so that it's not an issue to have an HPy member in it. We may have to revisit this when we implement more of the buffer API.

hpy/debug/src/debug_ctx_cpython.c
```diff
@@ -35,6 +35,36 @@ static inline PyObject *_dh2py(DHPy dh)
     return _h2py(DHPy_unwrap(dh));
 }
 
+static void _buffer_h2py(HPyContext ctx, const HPy_buffer *src, Py_buffer *dest)
```
antocuni, line 38: maybe call it `dctx`?

hpy/devel/include/universal/hpyfunc_trampolines.h
```diff
@@ -118,4 +118,34 @@ typedef struct {
         return a.result;                                                \
     }
 
+typedef struct {
+    cpy_PyObject *arg0;
+    cpy_Py_buffer *arg1;
+    int arg2;
```
hodgestar, line 124: Should we give these arguments more readable names rather than `arg0`. `arg1`? Maybe `self`, `buffer`, `flags`?

test/test_slots.py
```diff
@@ -264,6 +264,79 @@ def test_nb_ops_power(self):
...
+        mod = self.make_module("""
+            typedef struct {
+                HPyObject_HEAD
+                int exports;
+            } FakeArrayObject;
+
+            static char static_mem[12] = {0,1,2,3,4,5,6,7,8,9,10,11};
+            static HPy_ssize_t _shape[1] = {12};
+            static HPy_ssize_t _strides[1] = {1};
+            static HPy_buffer fakebuf = {
```
hodgestar, line 279: If `HPy_buffers` are meant to be short lived objects, should we perhaps fill in the buffer in the test, instead of reusing this static one?

hpy/debug/src/debug_ctx_cpython.c
hodgestar, line 38: It would be good to avoid duplicating these helper functions, but I'm not quite sure where we would put them.

hpy/devel/include/cpython/hpyfunc_trampolines.h
hodgestar, line 77: I *think* now this cast to `HPy_buffer` is wrong? `arg1` should be a `cpy_Py_buffer` which is just `Py_buffer` right? Maybe now the cast can just be removed?

hodgestar:
@rlamy Done with my second review!

hpy/debug/src/debug_ctx_cpython.c
antocuni: yes and no. I agree that code duplication is bad, but OTOH this file and `ctx_meth.c` (where the "other" functions live) are supposed to be independent. The mess comes from the fact that the universal mode should be interpreter-agnostic, the debug mode too and both should be independent from each other.
But in practice, we need to put CPython-specific code in both (which is `ctx_meth.c` for the universal mode and `debug_ctx_cpython.c` for the debug mode). Maybe we *could* try to share code between the CPython-specific parts of the two modes, but as long as the code duplication is short&straightforward, I think it's simpler to just live with it.

hpy/devel/include/cpython/hpyfunc_trampolines.h
antocuni: yes but `IMPL` takes an `HPy_buffer`, so the cast is needed, I think. The actual problem is that casting `Py_Buffer*` to `HPy_Buffer*` is probably undefined behavior, although it will work in practice.  We should put a comment and explain the issue.

hodgestar: Does this mean that a type that registers a getbufferproc slot using HPy but is called with `PyObject_GetBuffer` will return a buffer object where `obj` is an `HPy` and will possibly break when `PyBuffer_Release` is called on and attempts to `Py_XDECREF` `obj`?

hpy/devel/include/universal/hpyfunc_trampolines.h
rlamy: Done, using `self`, `view`, `flags`.

hpy/debug/src/debug_ctx_cpython.c
rlamy: done.

hpy/devel/include/cpython/hpyfunc_trampolines.h
rlamy: Unless we change the implementation of `HPy`, the likely issue is that `Py_buffer` and `HPy_buffer` could have different layouts, so any member could be wrong and we'd quickly end up with a segfault as soon as the code does anything with the buffer.

hpy/devel/src/runtime/ctx_type.c
rlamy: Yes, this hack isn't needed on 3.9+. And I think it'd be better to already implement things correctly on 3.9, using a few `#if PY_VERSION_HEX < 0x03090000`, but that would require us testing on 3.9 first...

rlamy:
I guess this is ready to merge now.

hodgestar (approved):

PR #164, opened 2021-02-09
Use the debug mode in tests

antocuni:
The main feature of this PR is the `hpy_debug` pytest fixture, which checks that you don't have memory leaks in your hpy tests (when you run in debug mode). This fixture is automatically used by our tests so now we get leak detection out of the box.

Among the other things:
1. improve the formatting of `HPyLeakError`
2. add support for HPyTracker to the debug mode: before this PR, it was simply wrong and handles closed through HPyTracker_Close were reported as leaks
3. fix some actual memory leaks which were found when running the tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,7 @@ cppcheck: cppcheck-build-dir
 		--cppcheck-build-dir=$(or ${CPPCHECK_BUILD_DIR}, .cppcheck) \
 		--output-file=$(or ${CPPCHECK_BUILD_DIR}, .cppcheck)/output.txt \
 		--enable=warning,performance,portability,information,missingInclude \
+		--inline-suppr \
 		-I hpy/devel/include/ \
 		-I hpy/devel/include/common/ \
 		-I hpy/devel/include/cpython/ \
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ cppcheck-build-dir:
 	mkdir -p $(or ${CPPCHECK_BUILD_DIR}, .cppcheck)
 
 cppcheck: cppcheck-build-dir
-	cppcheck --error-exitcode=1 --cppcheck-build-dir=$(or ${CPPCHECK_BUILD_DIR}, .cppcheck) --enable=warning,performance,portability,information,missingInclude --report-progress -I hpy/devel/include/ -I hpy/devel/include/common/ -I hpy/devel/include/cpython/ -I hpy/devel/include/universal/ -I hpy/universal/src/ --force -D NULL=0 .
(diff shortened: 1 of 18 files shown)


hpy/debug/leakdetector.py
```diff
@@ -1,16 +1,18 @@
...
+class HPyLeakError(HPyDebugError):
     def __init__(self, leaks):
         super().__init__()
         self.leaks = leaks
 
     def __str__(self):
         lines = []
-        lines.append('%s handles have not been closed properly:' % len(self.leaks))
+        n = len(self.leaks)
+        s = 's' if n > 1 else ''
```
hodgestar, line 14: This should be `'s' if n != 1 ...` since 0 handles should display `0 unclosed handles`.
```suggestion
        s = 's' if n != 1 else ''
```

hpy/debug/src/include/hpy_debug.h
```diff
@@ -3,7 +3,28 @@
...
+  by hpy.universal implementations (including but not limited to the
+  CPython's version of hpy.universal which is included in this repo).
+
+  The idea is that for every uctx there is a corresponding unique dctx which
+  wraps it.
+
+  If you call hpy_debug_get_ctx twice on the same uctx, you get the same
+  result.
+
+  IMPLEMENTATION NOTE: at the moment of writing, the only knonw user of the
```
hodgestar, line 17:
```suggestion
  IMPLEMENTATION NOTE: at the moment of writing, the only known user of the
```

hodgestar (approved):
Left a few small suggestions, but looks good to me.

Makefile
```diff
@@ -15,10 +15,25 @@ cppcheck-build-dir:
...
+		--enable=warning,performance,portability,information,missingInclude \
+		--inline-suppr \
+		-I hpy/devel/include/ \
+		-I hpy/devel/include/common/ \
+		-I hpy/devel/include/cpython/ \
+		-I hpy/devel/include/universal/ \
+		-I hpy/universal/src/ \
+		--force \
+		-D NULL=0 \
+		. || (cat $(or ${CPPCHECK_BUILD_DIR}, .cppcheck)/output.txt && false)
```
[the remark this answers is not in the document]
antocuni: I tried but it doesn't work :(
See [this azure log](https://dev.azure.com/pyhandle/hpy/_build/results?buildId=1162&view=logs&j=71490d96-a242-5f1c-bb38-546ccf5abc39&t=31ff8806-cfa1-5f20-3b24-e9367d546093) for example: you can see that I ran it with `2>&1` but the error is not shown, and I have no clue why

hpy/debug/leakdetector.py
antocuni: maybe, although you should never get this if you have 0 unclosed handles 😅. But I suppose it doesn't hurt

PR #165, opened 2021-02-10
Remove Travis CI.

hodgestar:
The Travis CI completely:

* The IRC notifications are noisy and don't contain particularly interesting information, so let's turn them off.
* The Travis CI runs are uninteresting, because all our tests happen in Azure Pipelines.
* Travis seems to be slowly dropping support for its free tier, so we're unlikely to add more to it in the foreseeable future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,3 @@
 language: python
 
 script : true
-
-notifications:
-    irc: 
-        channels: "chat.freenode.net#hpy"
-        template:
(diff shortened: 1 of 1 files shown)


antocuni (commented):
I think we should just remove the file completely

hodgestar:
> I think we should just remove the file completely

Agreed. Done.

antocuni:
LGTM

hodgestar:
I've also removed the Travis CI integration from the GitHub settings.

PR #166, opened 2021-02-11
Implement HPyDef_DOCSTRING

hodgestar:
So that we can add docstrings to types.

Note: Because of how we implement defines with `HPyDef`, some of the `HPy_tp_XXX` like `HPy_tp_doc` are not actually used anywhere. I'm not sure if we should keep the definitions or mark them as unused somehow.

### DIFF
--- a/hpy/devel/include/common/hpydef.h
+++ b/hpy/devel/include/common/hpydef.h
@@ -70,11 +70,16 @@ typedef struct {
     void *closure;
 } HPyGetSet;
 
+typedef struct {
+    const char *docstring;  // the null-terminated UTF-8 docstring
+} HPyDocString;
+
 typedef enum {
     HPyDef_Kind_Slot = 1,
     HPyDef_Kind_Meth = 2,
     HPyDef_Kind_Member = 3,
     HPyDef_Kind_GetSet = 4,
+    HPyDef_Kind_DocString = 5,
 } HPyDef_Kind;
 
 typedef struct {
@@ -84,6 +89,7 @@ typedef struct {
         HPyMeth meth;
         HPyMember member;
         HPyGetSet getset;
+        HPyDocString docstring;
     };
 } HPyDef;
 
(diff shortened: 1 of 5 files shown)


antocuni:
If we decide to keep `HPyDef_DOCSTRING`, the PR looks good.
However, I wonder whether we could make it simpler an just add a `.doc` fields to `HPyType_Spec`. For CPython's `PyType_Spec` is less necessary because you need to write something like this:
```c
static PyType_Slot Xxo_Type_slots[] = {
    {Py_tp_doc, "The Xxo type"},
   ...
}
```

for for HPy, `HPyDef_DOCSTRING` sounds a bit heavyweight.

> Note: Because of how we implement defines with HPyDef, some of the HPy_tp_XXX like HPy_tp_doc are not actually used anywhere. I'm not sure if we should keep the definitions or mark them as unused somehow.

Ideally, it would be nice if we find a way to tell the compiler to emit an error if you use it. Something like:
`ERROR: HPy_tp_doc doesn't exist, please use HPyType_Spec.doc = ... instead`

Not sure if it's possible though

hpy/devel/include/common/hpydef.h
```diff
@@ -70,11 +70,16 @@ typedef struct {
     void *closure;
 } HPyGetSet;
 
+typedef struct {
+    const char *docstring;  // the null-terminated UTF-8 docstring
+} HPyDocString;
```
fangerer, line 75: Why do you create a separate definition kind?
Since it's just a slot in CPython, we could also just store it in `HPySlot`. This, of course, has the penalty of two more words (`HPySlot.cpy_trampoline` and the slot kind).

fangerer (approved):
It's still fine for me to have a new definition kind.

hpy/devel/include/common/hpydef.h
antocuni: well, in CPython everything is a slot :). In HPy we are trying to be a more type safe so if we want to introduce `HPy_tp_doc`, I think this is the way to go.
(I still think I prefer `.doc`, though)

hodgestar: HPy works a bit differently. The macros generate `HPyDef` objects, which are then examined when the type is created. I could stuff the docstring into the `void * impl` filed of HPySlot, but then all the logic that looks at the slots would have to special case the docstring because it's not a function.

hodgestar:
I like the idea of using `.doc`, so I'll reimplement type docstrings that way. That also matches what happens for modules docstrings nicely.

hodgestar:
Closing in favour of the `.doc` `HPyType_Spec` field @antocuni suggested and which I have implemented in #167.

PR #171, opened 2021-02-21
revert to using alloca

### LINKED ISSUE #157: Don't use Variable Length Arrays (VLA)
VLA are a feature of C99 but they are optional so not all compilers use it.
At first, we used `alloca` to allocate arrays on the stack, then we introduced VLA in #76 , but this causes problems on windows, see #125 .

We should revert #76 .

mattip:
Reverts #76
Closes #157 

Use alloca instead of VLA to support MSVC

### DIFF
--- a/hpy/debug/src/debug_ctx_cpython.c
+++ b/hpy/debug/src/debug_ctx_cpython.c
@@ -97,7 +97,7 @@ void debug_ctx_CallRealFunctionFromTrampoline(HPyContext dctx,
         _HPyFunc_args_VARARGS *a = (_HPyFunc_args_VARARGS*)args;
         DHPy dh_self = _py2dh(dctx, a->self);
         Py_ssize_t nargs = PyTuple_GET_SIZE(a->args);
-        DHPy dh_args[nargs * sizeof(DHPy)]; // VLA, should be killed by #157
+        DHPy *dh_args = (DHPy *)alloca(nargs * sizeof(DHPy));
         for (Py_ssize_t i = 0; i < nargs; i++) {
             dh_args[i] = _py2dh(dctx, PyTuple_GET_ITEM(a->args, i));
         }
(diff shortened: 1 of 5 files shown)


mattip:
I am not sure how to do error checking, but these allocations should be small

hpy/debug/src/debug_ctx_cpython.c
```diff
@@ -97,7 +97,7 @@ void debug_ctx_CallRealFunctionFromTrampoline(HPyContext dctx,
         _HPyFunc_args_VARARGS *a = (_HPyFunc_args_VARARGS*)args;
         DHPy dh_self = _py2dh(dctx, a->self);
         Py_ssize_t nargs = PyTuple_GET_SIZE(a->args);
-        DHPy dh_args[nargs * sizeof(DHPy)]; // VLA, should be killed by #157
+        DHPy *dh_args = (DHPy *)alloca(nargs * sizeof(DHPy));
```
mattip, line 100: I think this was wrong in #76 - it allocated too many values?

antocuni: I think you are right, thanks for fixing it

antocuni (approved):
LGTM

hodgestar (approved):

hodgestar:
`cppcheck` is complaining about the use of `alloca`. We can turn that off by either:

* Adding `--suppress=alloca` to the Makefile call to cppcheck, or
* Adding `// cppcheck-suppress alloca` to each line that calls alloca

The former feels neater, since we'll likely need to use alloca a bit all over the place.

mattip:
> Adding `--suppress=alloca` to the Makefile call to cppcheck

Done

hodgestar:
> > Adding `--suppress=alloca` to the Makefile call to cppcheck

> Done

Apologies. I think it should be `-suppress=allocaCalled`. :/

hodgestar (approved):
Eep. Good catch cppcheck!

thedrow:
> 
> 
> Eep. Good catch cppcheck!

Yay! I'm glad it helped.
I learned something new about VLAs through this experience. Sorry it took you guys through this advanture as well.

hodgestar:
> > Eep. Good catch cppcheck!
> 
> Yay! I'm glad it helped.

:D

> I learned something new about VLAs through this experience. Sorry it took you guys through this advanture as well.

No apologies needed -- I think we all learned things along the way.

Issue #158, opened 2021-01-25
Make it possible to put HPyType_Spec in HPyModuleDef.defines


antocuni:
This is an idea which just came to my mind, let me write it down not to forget.
To create a type, we currently need to define an `HPyType_Spec` and then manually call `HPyType_FromSpec` in the module init, as you do on CPython.
However, for the vast majority of cases this is unnecessarily complicate. Let's support something like this:
```c
HPyType_Spec PointType_spec = { ... };
HPyDef_TYPE(Point_def, "Point", PointType_spec);

static HPyDef *module_defines[] = {
    &Point_def,
    NULL
};

static HPyModuleDef moduledef = {
    HPyModuleDef_HEAD_INIT,
    .m_name = "foo",
    .defines = module_defines
};
```

The semantics would be to call `HPyType_FromSpec` and put the result inside the module dictionary, of course

hodgestar:
I like this idea & I suggested something similar myself in IRC.

The upsides: removes the only non-declarative part of defining types in modules; provides one obvious way to set up a type; helps people avoid accidentally doing slightly odd things during type setup.

Downsides: adds a layer of abstraction; makes it a bit harder to customize type setup; I'm not sure how many types will need to doing something different during type setup.

antocuni:
> I like this idea & I suggested something similar myself in IRC.

uh, maybe I read it on IRC and that's why it came to my mind: sorry for the missing attribution 😅.
OTOH, no cookie for you because you didn't open an issue first 😛
 
> The upsides: removes the only non-declarative part of defining types in modules; provides one obvious way to set up a type; helps people avoid accidentally doing slightly odd things during type setup.
> 
> Downsides: adds a layer of abstraction; makes it a bit harder to customize type setup; I'm not sure how many types will need to doing something different during type setup.

To be clear, I am not proposing to kill `HPyType_FromSpec`: it will still be possible to call it manually, if you want.

Also, note that maybe `HPyDef_TYPE` is not needed. We could write directly this:
```c
HPyDef PointTypeDef = {
    .kind = HPyDef_Kind_TypeSpec,
    .typespec = {
        .name = "foo",
        .defines = ...
    }
};
```

hodgestar:
I started implementing this in https://github.com/hpyproject/hpy/pull/173, but passing a base type or metaclass requires access to the handle for that type, which won't be available statically. This made me wonder whether implementing this is a good idea? Perhaps there is something we can do instead?

[...]

PR #173, opened 2021-02-24
WIP: Add support for listing type specs in module defines

hodgestar:
This should make it significantly easier to add types to modules.

Addresses #158.

Still to do:
* Clean up error handling.
* Add support for specparams / bases (might be hard)
* Think about whether we need support for tweaking types after they've been created.

### DIFF
--- a/hpy/devel/include/common/hpydef.h
+++ b/hpy/devel/include/common/hpydef.h
@@ -3,7 +3,10 @@
 
 #include <stddef.h> /* to make sure "offsetof" is available for our users */
 
+typedef struct HPyDef HPyDef;
+
 #include "common/hpyfunc.h"
+#include "common/hpytype.h"
 #include "common/autogen_hpyslot.h"
 
 typedef struct {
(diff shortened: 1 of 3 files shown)


test/test_00_basic.py
```diff
@@ -34,6 +34,38 @@ def test_different_name(self):
         """, name="foo")
         assert mod.__name__ == "foo"
 
+    def test_module_typespec(self):
+        mod = self.make_module("""
+            static const HPyType_Spec Foo_spec = {
+                .name = "mytest.Foo",
+            };
+            HPyDef_TYPE(Foo_def, "Foo", Foo_spec)
```
hodgestar, line 42: Clang complains that `Foo_spec` is not a compile time constant. GCC is happy with the construct. I'm not sure how to make Clang happy. I could pass the actual spec definition into the macro, but that seems likely to end up being a bit ugly.

hpy/devel/include/common/hpydef.h
```diff
@@ -70,22 +73,29 @@ typedef struct {
     void *closure;
 } HPyGetSet;
 
+typedef struct {
+    const char *name;
+    HPyType_Spec typespec;
+} HPyTypeSpecDef;
+
```
antocuni, line 80: This indirection looks weird. Why do we need both `HPyType_Spec` **and** `HPyTypeSpecDef`?
And moreover, both have a `.name` which will make things even more confusing, especially if you don't use the same name in both places.

Note, this is a rhetorical question: I *know* why you had to introduce this, but I think it causes more problems than it solves, and the clang complaint is one of those. I propose one of the following:

```c
struct HPyDef {
    HPyDef_Kind kind;
    union {
        ...
        HPyType_Spec *typespec;
    };
} HPyDef;

struct HPyDef {
    HPyDef_Kind kind;
    union {
        ...
        HPyType_Spec typespec;
    };
} HPyDef;
```

The first option implies that you need to create a typespec separately, and that it can also be in another translation unit. The `HPyDef` can be generated using `HPyDef_TYPE`.

The second option removes the need to create a typespec explicitly. The user will need to write an `HPyDef` by hand, like this:

```c
HPyDef MyType {
    .kind = HPyDef_Kind_Type,
    .typespec = {
        .name = "MyType",
        ...
    }
}
```

Both have pros and cons, I have no strong preference for any of them

hodgestar:
Looking at `HPyType_SpecParam` has made me wonder if this approach is a good idea. Passing a base type or metaclass requires access to the handle for that type, which won't be available statically.

Perhaps we should just let people write the bit of C code for adding types to modules themselves?

Alternatively we could write a small macro or function that takes a `HPyType_Spec` and `HPyType_SpecParam` and adds the new type to the module and returns the type (in case more is needed).

hodgestar:
Closing this in favour of #180.

PR #176, opened 2021-02-26
Detect usage of closed handles

antocuni:
This PR enhances the debug mode so that it is able to detect usage of already-closed handles.

Detecting closed handles is trickier than detecting leaks. The problem is that as soon as we `free()` a `DebugHandle`, the libc will probably reuse the released memory to `malloc()` one of the subsequent handles, with the consequence that our invalid handles will look like a valid one, but to a different object!

To solve the issue, we introduce the `info->closed_handles` queue, where we put closed `DebugHandles` instead of `free()`ing them immediately. Moreover, once the queue reaches its maximum capacity, we reuse handles from there instead of `malloc()`ing new ones: by doing this, we are basically using the `closed_handles` queue as a free list, which should also improve performances (but I didn't do any benchmark yet).

To explain what happens when we detect the usage of a closed handle, see #51c3212. I'll copy&paste the commit message here for convenience:
```
    Ideally, we would like to raise a clean exception when this happens, but it is
    impossible to do in a general way because some of the API functions can't
    fail. I tried to raise an exception in some cases and abort in others, but
    it's too complicate and probably not so useful: using a closed handle is more
    or less equivalent to dereference a NULL pointer, so a clean abort is probably
    a reasonable thing to do anyway.
    
    However, I also added the possibility of adding a custom hook, and in that
    case the execution does NOT abort. This is useful:
    
    1. in tests
    
    2. on CPython, when porting existing modules to HPy. On CPython "closed
       handles" are still valid in practice as long as the refcount > 0: it should
       make it easier to port extensions to HPy, e.g. by printing a warning inside
       the callback and let the execution to continue, so that people can fix the
       warnings one by one.
```

Misc stuff which are part of this PR:
- I added the possibility to write tests for C code, and the corresponding azure step
- I added `HPyErr_NewException`, `HPyErr_NewExceptionWithDoc` and `HPyImport_ImportModule`: I introduced them because I thought I would need. At the end I didn't need, but I'm including them in this PR anyway since it's a very small change.


antocuni:
I am merging this as-is, since it looks like nobody is going to review it :)

test/test_hpyimport.py
```diff
@@ -0,0 +1,24 @@
...
+                    return HPy_NULL;
+                return HPyImport_ImportModule(ctx, name);
+            }
+            @EXPORT(f)
+            @INIT
+        """)
+        sys2 = mod.f(b'sys')
+        assert sys is sys2
+        with pytest.raises(ImportError):
+            mod.f(b'This is the name of a module which does not exist, hopefully')
```
hodgestar, line 24: :joy_cat:

hpy/debug/src/debug_handles.c
```diff
@@ -1,25 +1,98 @@
...
+    // XXX: so far the only call API that we have is CallTupleDict, we
+    // should use a more C-friendly variant as soon as we have it.
+    UHPy uh_args = HPy_NULL;
+    UHPy uh_res = HPy_NULL;
+    uh_args = HPyTuple_Pack(uctx, 0);
+    if (HPy_IsNull(uh_args)) {
+        print_error(uctx, "Error when preparing args to call the on_invalid_handle callback");
+        goto exit;
+    }
+    uh_res = HPy_CallTupleDict(uctx, info->uh_on_invalid_handle, uh_args, HPy_NULL);
```
hodgestar, line 86: If `uh_args` is intended to be empty, then you should be able to pass `HPy_NULL` instead of making a tuple. At least that was my intention when I wrote `HPy_CallTupleDict`.

hodgestar (commented):
In addition to the smaller comments, is there a way to get something similar to the LeakDetector but for closed handles? E.g. but registering a callback that will keep track of issues and report them when a context handler exits? Maybe that is already a future plan?

It would also be nice to be able to have a single context manager that does both the leak detecting and the closed handle use detecting.

antocuni:
> In addition to the smaller comments, is there a way to get something similar to the LeakDetector but for closed handles? E.g. but registering a callback that will keep track of issues and report them when a context handler exits? Maybe that is already a future plan?

I think that in the best case it's very hard to do and the worst case it might be impossible to do for the general case. As soon as you try to use a closed handle, your program is invalid: what do you propose we should do in that case?
1. Raise an exception? This would be nice but it's the opposite of reporting issues in the LeakDetector as you suggest. Moreover, it's hard (see below)
2. Let the program continue and report the problems at the end. This is also problematic because the user program is in an undefined state now, especially in case of API calls which cannot fail. Imagine the following:
```c
if (HPyBytes_Check(ctx, h)) {
    HPy_Close(ctx, h); // mistake!
    const char *buf = HPyBytes_AsString(ctx, h);
    ...
}
```
The user doesn't expect the call to fail (since it has already called `HPyBytes_Check`), the program will likely segfault and/or have undefined behavior and the mistake which we recorded will never be reported because we will never arrive at that point. I think it's much better to report the mistake immediately. We could argue that it's the user's fault not to check for errors after `HPyBytes_Check`, but the debugging mode is supposed to catch user's mistakes, not to fail because of them :).

What we *could* do is to raise regular exceptions whenever it's possible, and abort() only in the few cases in which we can't signal an error. However, this has a bunch of problems:
1. it wouldn't solve the problem above. `HPyBytes_AsString` CAN raise, it's just that we don't expect the user to do the check in that case. 
2. it's a mess to implement: I tried and failed. Basically it means that you need to check for errors whenever you call `h2py`, but you also need to write code in a way which doesn't introduce slowdowns in non-debug mode. I ended up with a lot of code duplication, manually written code which needs to be carefully checked, etc.
3. Even if you implement (2), you still need to abort() when you call cannot-fail APIs. And I suspect that the majority of real-world mistakes will come from calling `HPy_Close` twice.

> It would also be nice to be able to have a single context manager that does both the leak detecting and the closed handle use detecting.

I kind of agree in principle but on the other hand, leaks and double-close are so fundamentally different that you can't handle them in the same way

hodgestar:
@antocuni Hmm. Maybe a bigger question is then how should people use the closed handle detector in their code? Run in debug mode and look for error messages in the console?

antocuni:
> @antocuni Hmm. Maybe a bigger question is then how should people use the closed handle detector in their code? Run in debug mode and look for error messages in the console?

well, the default behavior is that if you use a closed handle in debug mode, it calls `HPy_FatalError()` so it aborts. I know it's not very nice but if you are testing C code, you should be prepared to get segfaults too. From my POV, using a closed handle is an error similar to dereferencing a dangling pointer.

hpy/debug/src/dhqueue.c
```diff
@@ -0,0 +1,117 @@
+#include "debug_internal.h"
+
+void DHQueue_init(DHQueue *q) {
```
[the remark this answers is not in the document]
antocuni: good idea, I'll add a note

hpy/debug/src/debug_handles.c
antocuni: good catch, thank you. This simplifies the code a lot

Issue #153, opened 2021-01-13
Setting __name__ for extensions inside packages


rlamy:
When using the universal ABI, extension modules inside packages get their `__name__` directly from the value set in `moduledef.m_name`. However, according to CPython, `__name__` should look like `package.extension_name` in such a case, even if the C code only has `.m_name = "extension_name",`. IOW, applying the following diff, the test should pass (and does indeed pass with the cpython ABI).

```
diff --git a/proof-of-concept/pofpackage/foo.c b/proof-of-concept/pofpackage/foo.c
index 1480bc7..eea2134 100644
--- a/proof-of-concept/pofpackage/foo.c
+++ b/proof-of-concept/pofpackage/foo.c
@@ -13,7 +13,7 @@ static HPyDef *module_defines[] = {
 };
 static HPyModuleDef moduledef = {
     HPyModuleDef_HEAD_INIT,
-    .m_name = "pofpackage.foo",
+    .m_name = "foo",
     .m_doc = "HPy Proof of Concept",
     .m_size = -1,
     .defines = module_defines
diff --git a/proof-of-concept/test_pof.py b/proof-of-concept/test_pof.py
index e0f0c53..8fe3307 100644
--- a/proof-of-concept/test_pof.py
+++ b/proof-of-concept/test_pof.py
@@ -19,4 +19,5 @@ def test_point():
     assert repr(p) == 'Point(?, ?)' # fixme when we have HPyFloat_FromDouble
 
 def test_pofpackage():
+    assert pofpackage.foo.__name__ == 'pofpackage.foo'
     assert pofpackage.foo.hello() == 'hello from pofpackage.foo'
```

hodgestar:
Addressed by #177.

PR #177, opened 2021-03-03
Correctly set the __name__ of universal modules inside packages.

hodgestar:
Addresses #153.

### DIFF
--- a/hpy/devel/__init__.py
+++ b/hpy/devel/__init__.py
@@ -107,6 +107,7 @@ def __bootstrap__():
     m = load({module_name!r}, ext_filepath)
     m.__file__ = ext_filepath
     m.__loader__ = __loader__
+    m.__name__ = __name__
     m.__package__ = __package__
     m.__spec__ = __spec__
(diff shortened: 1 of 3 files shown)


auvipy (approved):

rlamy (approved):
Well, that was easy! The name will still be wrong during module initialisation though, but I guess that's fine.

hodgestar:
> Well, that was easy! The name will still be wrong during module initialisation though, but I guess that's fine.

It will be wrong during initialisation, but that is true on CPython too, I think, because it's the import machinery that sets `__name__`.

rlamy:
> It will be wrong during initialisation, but that is true on CPython too, I think, because it's the import machinery that sets `__name__`.

Actually, CPython does have [some weird code](https://github.com/python/cpython/blob/e161ec5dd7ba9355eb06757b9304019ac53cdf69/Objects/moduleobject.c#L207) to set the correct name early. And cpyext does reimplement the hack, so I guess it could matter to someone somewhere...

Issue #154, opened 2021-01-14
stub file error when --hpy-abi=universal


antocuni:
I got this error when trying to install the [antocuni/update-to-latest-hpy](https://github.com/hpyproject/ultrajson-hpy/tree/antocuni/update-to-latest-hpy) branch of ultrajson-hpy (hpy version fdc6047):
```
$ python setup.py --hpy-abi=universal install
[... build goes smoothly ...]
creating stub loader for ujson_hpy.hpy.so
stub file already created for ujson_hpy.hpy.so
error: file '/home/antocuni/pypy/misc/ultrajson-hpy/build/bdist.linux-x86_64/egg/ujson_hpy.py' does not exist
```
The same command succeeds for `proof-of-concept`, so I think it's something inside ujson's `setup.py`.
I **think** it might be related to `cmdclass` but I didn't investigate:
```
setup(
    name = 'ujson-hpy',
    ...
    cmdclass = {'build_ext': build_ext, 'build_clib': build_clib_without_warnings},
)
```

/cc @hodgestar as he's our setuptools expert by now :)

hodgestar:
I pushed a small fix to https://github.com/hpyproject/ultrajson-hpy/tree/antocuni/update-to-latest-hpy.

The high-level issue is that ultrajson-hpy's setup.py was passing the *distutils* version of `build_ext` rather than the setuptools one.

The question now is what to do about it on the HPy side? It's likely a common and confusing issue, so we should do something. 

One option is to try detect the issue and either raise an error or warning. One way to detect the issue is to check whether the setuptools build_ext is somewhere in the inheritance hierarchy of the current build_ext. If it isn't, the build will probably fail anyway.

Another option is to somehow handle the case where only distutils is being used, but I don't really know what that should look like, other than that it would probably be even more of a mess.

paugier:
My 2 cents: there is a discussion to depreciate distutils (https://www.python.org/dev/peps/pep-0632/). Now that we have pep517 and isolated builds with pyproject.toml, supporting builds using hpy with only distutils doesn't seem to be very useful.

antocuni:
> One option is to try detect the issue and either raise an error or warning. One way to detect the issue is to check whether the setuptools build_ext is somewhere in the inheritance hierarchy of the current build_ext. If it isn't, the build will probably fail anyway.
> Another option is to somehow handle the case where only distutils is being used, but I don't really know what that should look like, other than that it would probably be even more of a mess.

[...]

PR #178, opened 2021-03-03
Add assert that the setuptools build_ext is present.

hodgestar:
Addresses #154.

I'm not sure of a good way to add a test for this.

### DIFF
--- a/hpy/devel/__init__.py
+++ b/hpy/devel/__init__.py
@@ -53,6 +53,15 @@ class HPyDevel:
         base_build_ext = dist.cmdclass.get("build_ext", build_ext)
         orig_bdist_egg_write_stub = bdist_egg_mod.write_stub
 
+        assert ('setuptools.command.build_ext', 'build_ext') in [
+            (c.__module__, c.__name__) for c in base_build_ext.__mro__
+        ], (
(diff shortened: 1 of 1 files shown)


rlamy (approved):
A bit ugly, but that's probably the best we can do!

hpy/devel/__init__.py
```diff
@@ -53,6 +53,15 @@ def fix_distribution(self, dist):
...
 
+        assert ('setuptools.command.build_ext', 'build_ext') in [
+            (c.__module__, c.__name__) for c in base_build_ext.__mro__
+        ], (
+            "dist.cmdclass['build_ext'] does not inherit from"
+            " setuptools.command.build_ext.build_ext. The HPy build"
+            " system does not currently support any other build_ext"
+            " classes."
+        )
+
```
antocuni, line 64: sorry for the late review.
This is ugly but I agree it's the best we can do. 
We could improve the error message, though. Something like `if you are using distutils.commands.build_ext, please use setuptools.commands.build_ext`

hodgestar: Good idea. I push the improved error message straight to master in @06bde571321218d8f6466f05257b680ba1d65fd1.

PR #180, opened 2021-03-04
Add HPyModule_AddType.

hodgestar:
This is another attempt to address #158 after the stumbling blocks around statically accessing base type handles in #173.


test/test_module_helpers.py
```diff
@@ -0,0 +1,69 @@
...
+                }
+                return HPy_Dup(ctx, ctx->h_None);
+            }
+
+            @EXPORT(f)
+            @INIT
+        """)
+        assert getattr(mod, "Dummy", None) is None
+        mod.f()
+        print(dir(mod))
```
antocuni, line 34: this looks a leftover?

test/test_module_helpers.py
```diff
@@ -0,0 +1,69 @@
...
+                {
+                    return HPy_NULL;
+                }
+                return HPy_Dup(ctx, ctx->h_None);
+            }
+
+            @EXPORT(f)
+            @INIT
+        """)
+        assert getattr(mod, "Dummy", None) is None
```
antocuni, line 32: what about `assert not hasattr(mod, 'Dummy')`? Looks more direct

antocuni (approved):
LGTM overall.
I am not 100% sure about the name though.
From some POV it's a good name because it's clearly useful for modules. But OTOH, it's not module-specific: you can pass any object to it, since it uses `HPy_SetAttr` internally, so calling it `HPyModule_*` is a bit weird.

I don't have suggestions for a better name though, maybe someone else does?

hodgestar:
> I don't have suggestions for a better name though, maybe someone else does?

What about `HPyType_CreateAndSetAttr`?

hodgestar:
If we change the name we should probably also change the name of the file from `module_helpers.h` to something else. I almost created `helpers.h` but felt it was too generic. Maybe `runtime_helpers.h`?

antocuni:
> What about HPyType_CreateAndSetAttr?

I agree that it conveys the actual meaning, but this is ugly.
Maybe `HPyModule_AddType` is not so bad, even though it is imprecise.

Some more random ideas (some of which I don't like at all, but maybe they trigger more ideas):
- `HPyType_Add`
- `HPyHelpers_AddType`
- `HPyType_FromSpecEx`
- `HPyType_FromSpec_SetAttr`
- `HPyType_FromSpecAndSetAttr`

pinging @rlamy @fangerer @arigo @mattip @timfel to see whether they have ideas :)

> If we change the name we should probably also change the name of the file from module_helpers.h to something else. I almost created helpers.h but felt it was too generic. Maybe runtime_helpers.h?

I suggest to create a generic file which we can use to put "random" helpers in the future. If/when it grows too much, we can split it. `helpers.[ch]` looks fine to me

hodgestar:
I renamed `module_helpers.[ch]` to `helpers.[ch]`. Will await some thoughts on the name.

test/test_module_helpers.py
hodgestar: Yep! Removed. Tx.

test/test_module_helpers.py
hodgestar: Done!

hodgestar:
@antocuni If you and the CI robots are happy, I will probably just merge this as is. I can always adjust the name later if needed.

antocuni:
it's late, but LGTM :)

PR #184, opened 2021-03-17
WIP: Add multi-phase module initialisation.

hodgestar:


hodgestar:
I would leave it open as a starting point until either multi-phase module init is implemented or it's decided that the approach here is superseded by a better idea.

mattip:
This is obsoleted by #353, I think we should close it. @hodgestar thoughts?

hodgestar:
> This is obsoleted by #353, I think we should close it. @hodgestar thoughts?

Agreed.

PR #187, opened 2021-03-26
PR #176 follow up

antocuni:
This is a small followup of #176 to fix a couple of things which were noticed after the PR was already merged.

### DIFF
--- a/hpy/debug/src/debug_handles.c
+++ b/hpy/debug/src/debug_handles.c
@@ -78,6 +78,7 @@ void DHPy_invalid_handle(HPyContext *dctx, DHPy dh)
     uh_res = HPy_CallTupleDict(uctx, info->uh_on_invalid_handle, HPy_NULL, HPy_NULL);
     if (HPy_IsNull(uh_res))
         print_error(uctx, "Error when executing the on_invalid_handle callback");
+    HPy_Close(uctx, uh_res);
 }
 
 void DHPy_close(HPyContext *dctx, DHPy dh)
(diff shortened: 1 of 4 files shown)


hpy/debug/src/debug_handles.c
```diff
@@ -74,23 +74,10 @@ void DHPy_invalid_handle(HPyContext *dctx, DHPy dh)
...
-    uh_res = HPy_CallTupleDict(uctx, info->uh_on_invalid_handle, uh_args, HPy_NULL);
-    if (HPy_IsNull(uh_res)) {
+    uh_res = HPy_CallTupleDict(uctx, info->uh_on_invalid_handle, HPy_NULL, HPy_NULL);
+    if (HPy_IsNull(uh_res))
         print_error(uctx, "Error when executing the on_invalid_handle callback");
-        goto exit;
-    }
- exit:
-    HPy_Close(uctx, uh_args);
-    HPy_Close(uctx, uh_res);
```
hodgestar, line 93: Should we not still close `uh_res`?

antocuni: ouch, sure! I simplified the code too much. Ironically, this is the kind of errors that the debug mode would catch, but we can't run the debug mode under itself :)

hodgestar: Hehe. Indeed. :)

antocuni:
@hodgestar I addressed your remarks, I'm going to merge this without further review since it's a very simple PR

Issue #186, opened 2021-03-25
Add tests for bases and dict arguments to HPyErr_NewException


hodgestar:
And to HPyErr_NewExceptionWithDoc.

PR #193, opened 2021-04-02
Implemented tests for base and dict for HPyErr_NewException and HPyErr_NewExceptionWithDoc

cklein:
```
pytest test/test_hpyerr.py::TestErr
= test session starts =
platform darwin -- Python 3.9.2, pytest-6.2.2, py-1.10.0, pluggy-0.13.1
rootdir: /Users/cklein/github/hpyproject/hpy
collected 51 items

test/test_hpyerr.py ...................................................                                                                                                                                                                [100%]

= 51 passed, 1 warning in 40.63s =
```

For https://github.com/hpyproject/hpy/issues/186

### DIFF
--- a/test/test_hpyerr.py
+++ b/test/test_hpyerr.py
@@ -405,49 +405,42 @@ class TestErr(HPyTest):
     def test_HPyErr_NewException(self):
         import pytest
         mod = self.make_module("""
-            HPyDef_METH(f, "f", f_impl, HPyFunc_O)
-            static HPy f_impl(HPyContext *ctx, HPy self, HPy arg)
+            HPyDef_METH(f, "f", f_impl, HPyFunc_VARARGS)
+            static HPy f_impl(HPyContext *ctx, HPy self, HPy *args, HPy_ssize_t nargs)
             {
                 static HPy h_FooError = HPy_NULL;
-                if (!HPy_IsTrue(ctx, arg)) {
-                    // cleanup and close the FooError which we created earlier
-                    if (!HPy_IsNull(h_FooError))
-                        HPy_Close(ctx, h_FooError);
-                    return HPy_Dup(ctx, ctx->h_None);
-                }
-                h_FooError = HPyErr_NewException(ctx, "mytest.FooError",
-                                                 HPy_NULL, HPy_NULL);
-                if (HPy_IsNull(h_FooError))
-                    return HPy_NULL;
-                HPyErr_SetString(ctx, h_FooError, "hello");
-                return HPy_NULL;
-            }
-            @EXPORT(f)
-            @INIT
-        """)
-        with pytest.raises(Exception) as exc:
-            mod.f(True)
-        assert issubclass(exc.type, Exception)
-        assert exc.type.__name__ == 'FooError'
-        assert exc.type.__module__ == 'mytest'
-        assert exc.type.__doc__ is None
-        mod.f(False) # cleanup
+                HPy arg,
+                    h_base = HPy_NULL,
+                    h_dict = HPy_NULL,
+                    h_doc = HPy_NULL;
 
-    def test_HPyErr_NewExceptionWithDoc(self):
-        import pytest
-        mod = self.make_module("""
-            HPyDef_METH(f, "f", f_impl, HPyFunc_O)
-            static HPy f_impl(HPyContext *ctx, HPy self, HPy arg)
-            {
-                static HPy h_FooError = HPy_NULL;
+                if (!HPyArg_Parse(ctx, NULL, args, nargs, "O|OOO", &arg, &h_base, &h_dict, &h_doc)) {
+                    return HPy_NULL;
+                }
+
                 if (!HPy_IsTrue(ctx, arg)) {
                     // cleanup and close the FooError which we created earlier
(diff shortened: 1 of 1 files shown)


antocuni (changes requested):
thank you for this, new tests are always welcome!

It's not your fault because we already have duplicated code for this, but I fear that it's becoming a bit too much for this single function.

Maybe it's possible to improve the situation by merging all these tests into a single one by doing the following:
1. make `f` to accept two additional arguments: the base and the doc
2. if the base and/or the doc is `None`, we pass `HPy_NULL` to `HPyErr_NewException`
3. else, we pass the argument we passed to `f`

This way, it should be possible to test all the combinations, shouldn't it?

cklein:
Sure, you are right, this is a lot of duplication. I will take care of it.

cklein:
I did a lot back and forth. I'm not sure this is the most elegant solution possible.

antocuni (changes requested):
thank you, this looks much better.
The only problem is that we can't use `@pytest_collecting.mark.parametrize`: these tests are automatically used also on PyPy, where they are the so-called [app tests](https://doc.pypy.org/en/latest/coding-guide.html#mixed-level-tests-deprecated).
The TL;DR explanation is that for those we use a limited mock of pytest and we can't use the full pytest features.

I think that for this test the best solution is to just call `mod.f` multiple times, each time with different args. I.e. something like this:

```python
mod = self.make_module(...)
with pytest.raises(Exception) as exc:
    mod.f(base=None, doc=None, dict=None)
...
with pytest.raises(Exception) as exc:
    mod.f(base=RuntimeError, doc=None, dict=None)
...
# etc etc
```

I don't think it's necessary to test all possible combinations. Probably it's enough to test:
1. `doc=None, base=None, dict=None`
2. `doc='hello', base=None, dict=None`
3. `doc=None, base=RuntimeError, dict=None`
4. `doc=None, base=None, dict={...}`
5. `doc='hello', base=RuntimeError, dict={...}`

test/test_hpyerr.py
```diff
@@ -402,64 +402,78 @@ def test_errorval_returned_by_api_functions_int(self):
         assert mod.f([100, 200, 300]) == 3
         assert mod.f(None) == -42
 
-    def test_HPyErr_NewException(self):
+    @pytest_collecting.mark.parametrize("doc", [None, b"mytest"])
+    @pytest_collecting.mark.parametrize('base', [True, False])
+    @pytest_collecting.mark.parametrize("dict_arg", [True, False])
+    def test_HPyErr_NewException(self, doc, base, dict_arg):
```
antocuni, line 408: why `True/False` for base and dict? I think it's simpler to just pass `None/RuntimeError` and `None/{...}`?

cklein: Yes, I think it's easier and much cleaner. The other approach was due to the fact that I didn't feel very familiar with HPy.
Not that I do now, but after having written some stuff, it starts to feel better.

antocuni: yes sure, that's normal. I'm glad that you are getting used to HPy :)

antocuni:
@cklein the test looks very good now, thank you.
I have another idea for how to remove the `@pytest.mark.parametrize` which is not supported by PyPy. Something like this:
```python
    def test_HPyErr_NewException(self, doc, base, dict_arg):
        mod = self.make_module(...)
        def check(base, doc, dict):
            with pytest.raises(Exception) as exc:
                mod.f(False, base, dict, doc)
            assert issubclass(exc.type, RuntimeError if base else Exception)
            assert exc.type.__name__ == 'FooError', exc.value
            assert exc.type.__module__ == 'mytest'
            if doc is None:
                assert exc.type.__doc__ is None
            else:
                assert exc.type.__doc__ == doc.decode("utf-8")

            if dict_arg:
                assert exc.type.__dict__["test"] == "pass"
            mod.f(True, None, None, None) # cleanup

        check(base=None, doc=None, dict=None)
        check(base=RuntimeError, doc=None, dict=None)
        check(base=None, doc=b'mytest', dict=None)
        check(base=None, doc=None, dict={"test": "pass"})
        check(base=RuntimeError, doc=b'mytest', dict={"test": "pass"})
```

cklein:
Oh, I didn't know about parametrize and PyPy. I changed it according to your proposal.
Note that I'm still doing the kwargs dance because If I understand correctly, if`None` is just passed to `mod.f`, there needs to be a check if that value is `None` and then set it to `HPy_NULL`, right?

```
if(HPy_Is(ctx, h_doc, ctx->h_None) {
    h_doc = HPy_NULL;
}
```

antocuni:
> Oh, I didn't know about parametrize and PyPy. I changed it according to your proposal.
> Note that I'm still doing the kwargs dance because If I understand correctly, if`None` is just passed to `mod.f`, there needs to be a check if that value is `None` and then set it to `HPy_NULL`, right?
> 
> ```
> if(HPy_Is(ctx, h_doc, ctx->h_None) {
>     h_doc = HPy_NULL;
> }
> ```

yes exactly, that was my idea!

cklein:
Ah, OK, let me have another iteration then

antocuni:
perfect! Thank you

PR #194, opened 2021-04-02
Empty initializers and zero size arrays are not ISO C compatible

cklein:
If compiled with all warnings turned on, this produced the following warnings:

```
hpy/devel/src/runtime/ctx_call.c:32:23: warning: use of GNU empty initializer extension [-Wgnu-empty-initializer]
        HPy items[] = {};
                      ^
hpy/devel/src/runtime/ctx_call.c:32:23: warning: zero size arrays are an extension [-Wzero-length-array]
2 warnings generated.
```

It seems to work fine with gcc and clang but might cause trouble on other platforms in the future.
There's another way to create an empty tuple using the TupleBuilder.

### DIFF
--- a/hpy/devel/src/runtime/ctx_call.c
+++ b/hpy/devel/src/runtime/ctx_call.c
@@ -29,7 +29,7 @@ ctx_CallTupleDict(HPyContext *ctx, HPy callable, HPy args, HPy kw)
     else {
         // args is null, but kw is not, so we need to create an empty args tuple
         // for CPython's PyObject_Call
-        HPy items[] = {};
+        HPy *items = NULL;
         HPy empty_tuple = HPyTuple_FromArray(ctx, items, 0);
         obj = PyObject_Call(_h2py(callable), _h2py(empty_tuple), _h2py(kw));
         HPy_Close(ctx, empty_tuple);


hpy/devel/src/runtime/ctx_call.c
```diff
@@ -29,8 +29,8 @@ ctx_CallTupleDict(HPyContext *ctx, HPy callable, HPy args, HPy kw)
     else {
         // args is null, but kw is not, so we need to create an empty args tuple
         // for CPython's PyObject_Call
-        HPy items[] = {};
-        HPy empty_tuple = HPyTuple_FromArray(ctx, items, 0);
+        HPyTupleBuilder builder = HPyTupleBuilder_New(ctx, 0);
+        HPy empty_tuple = HPyTupleBuilder_Build(ctx, builder);
         obj = PyObject_Call(_h2py(callable), _h2py(empty_tuple), _h2py(kw));
```
antocuni, line 34: what about using `HPy *items = NULL` instead? It should work fine with `HPyTuple_FromArray`, and `HPyTupleBuilder` looks overkill for this simple task

antocuni (changes requested):
Thank you for catching this!
I suggested a simpler fix (assuming it works :))

hpy/devel/src/runtime/ctx_call.c
cklein: Seems to work fine! I changed my PR

antocuni: thank you!

Issue #191, opened 2021-03-27
repr of universal ABI modules


antocuni:
Consider this:
```
$ cd proof-of-concept/
$ python setup.py --hpy-abi=universal build_ext -i
$ python
>>> import pof
>>> pof.__file__
'/home/antocuni/pypy/hpy/proof-of-concept/pof.hpy.so'
>>> pof
<module 'pof' from '/home/antocuni/pypy/hpy/proof-of-concept/pof.py'>
>>> pof.__spec__
ModuleSpec(name='pof', loader=<_frozen_importlib_external.SourceFileLoader object at 0x7fa4ed81c6e0>, origin='/home/antocuni/pypy/hpy/proof-of-concept/pof.py')
```

Note that `pof.__file__` says `pof.hpy.so`, but `repr(pof)` says `pof.py`. I **think** that the repr uses `pof.__spec__.origin` instead of `pof.__file__`.

This is even more obscure because our stub **does** save the new `__spec__` on the just-loaded module, see line 123:
https://github.com/hpyproject/hpy/blob/cc98cac7f623572c4b0096c1767d738b57b4beae/hpy/devel/__init__.py#L113-L127

I think that `importlib` overrides the `__spec__` later.  Probably there are higher-priority issues than this, but nevertheless the current behavior is a bit obscure, since we have a module which claims to be `pof.py` but it has the contents of `pof.hpy.so`.

cklein:
You are right with your first assumption, because the module has a `__spec__` attribute, [_module_repr_from_spec](https://github.com/python/cpython/blob/a7ff6df60c05e1b69fca743573b1e118bebf121d/Lib/importlib/_bootstrap.py#L297-L298) is executed which [returns the origin](https://github.com/python/cpython/blob/a7ff6df60c05e1b69fca743573b1e118bebf121d/Lib/importlib/_bootstrap.py#L591-L592).

At the time when `m.__spec__ = __spec__` is executed, `__spec__.origin` is set to `hello_new.py` which makes to my fairly limited understanding because it's the spec of the module currently being imported, which is `hello_new.py`.
Setting `m.__spec__.origin = ext_filepath` seems to return what you are expecting:

```
>>> hello_new
<module 'hello_new' from '...hello_new.hpy.so'>
```

antocuni:
thank you for investigating into this @cklein. If you want to make a PR I'll be happy to accept it (if you don't feel like I'll just open one next week and I'll give you the credit).

[...]

PR #195, opened 2021-04-03
Changing spec.origin to name of shared library

cklein:
This makes `repr(mod)` show the name of the shared library
instead of the python bootstrap module for universal ABI

This addresses https://github.com/hpyproject/hpy/issues/191

### DIFF
--- a/test/test_importing.py
+++ b/test/test_importing.py
@@ -25,7 +25,7 @@ class TestImporting(HPyTest):
                 del sys.modules[name]
         return module
 
-    def test_importing_attributes(self):
+    def test_importing_attributes(self, hpy_abi, tmpdir):
         import pytest
         if not self.supports_ordinary_make_module_imports():
             pytest.skip()
@@ -40,3 +40,11 @@ class TestImporting(HPyTest):
         assert mod.__spec__.loader is mod.__loader__
         assert mod.__spec__.name == 'mytest'
         assert mod.__file__
+
+        if hpy_abi == 'cpython':
+            from sysconfig import get_config_var
+            ext = get_config_var('EXT_SUFFIX')
+        else:
+            ext ='.hpy.so'
+
+        assert repr(mod) == '<module \'mytest\' from \'{}\'>'.format(tmpdir.join('mytest' + ext))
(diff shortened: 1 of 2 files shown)


test/test_importing.py
```diff
@@ -40,3 +40,11 @@ def test_importing_attributes(self):
         assert mod.__spec__.loader is mod.__loader__
         assert mod.__spec__.name == 'mytest'
         assert mod.__file__
+
+        if hpy_abi == 'cpython':
+            from distutils.sysconfig import get_config_var
+            ext = get_config_var('EXT_SUFFIX')
+        else:
+            ext ='.hpy.so'
+
```
antocuni, line 49: this is probably wrong on windows (and maybe mac), but I think it's fine not to care until we actually start to run tests on those platforms.

antocuni (approved):
the PR LGTM, thank you. I'm adding @hodgestar as a reviewer because he's most familiar than me about this area of the code

test/test_importing.py
```diff
@@ -40,3 +40,11 @@ def test_importing_attributes(self):
         assert mod.__spec__.loader is mod.__loader__
         assert mod.__spec__.name == 'mytest'
         assert mod.__file__
+
+        if hpy_abi == 'cpython':
+            from distutils.sysconfig import get_config_var
```
mattip, line 45:
```suggestion
            from sysconfig import get_config_var
```

mattip: See also https://bugs.python.org/issue39825. `distutils` is officially deprecated, anything hpy needs should be imported from `sysconfig` or `setuptools`

test/test_importing.py
cklein: I'm currently developing on a mac,  it's fine:
```
In [3]: sysconfig.get_config_var("EXT_SUFFIX")
Out[3]: '.cpython-39-darwin.so'
```

 On windows it should probably be ".pyd", but it also needs to be fixed in https://github.com/hpyproject/hpy/blob/master/hpy/devel/__init__.py#L248

hodgestar (approved):

hodgestar:
Thanks for fixing, @cklein.

PR #197, opened 2021-04-03
Implementation for comparing object identity.

cklein:
This allows comparing two objects for equality, e.g. `HPy_Is(h_obj, ctx->h_None)`.
I ran into this issue when trying to remove the duplicated code in https://github.com/hpyproject/hpy/pull/193.
Comparing objects for equality is documented in https://github.com/hpyproject/hpy/blob/master/docs/api.rst but I could not find the implementation, also not in the history.


antocuni:
> Comparing objects for equality is documented in https://github.com/hpyproject/hpy/blob/master/docs/api.rst but I could not find the implementation, also not in the history.

yes, you are correct: we always had the plan to introduce `HPy_Is`, but never did so far, thank you for taking care for this.

One nitpick: this is comparing object *identity*, not equality (which is done by calling `HPy_RichCompare`). I have renamed the PR accordingly

PR #198, opened 2021-04-03
Extended proof-of-concept

cklein:
Added some methods as examples:

* tp_destroy
* tp_richcompare
* a method without any arguments
* a method with varargs

The implementation is non-sense, its only purpose is to be a demo.

### DIFF
--- a/proof-of-concept/Makefile
+++ b/proof-of-concept/Makefile
@@ -1,11 +1,11 @@
-pof.cpython-37m-x86_64-linux-gnu.so: pof.c
-	python3.7 setup.py build_ext -i -f
+.PHONY: shared-library
+shared-library: pof.c
+	python setup.py build_ext -i -f
 
 .PHONY: test
-test: pof.cpython-37m-x86_64-linux-gnu.so
-	python3.7 -m pytest test_pof.py
+test: shared-library
+	python -m pytest test_pof.py
 
-universal: pof.hpy.so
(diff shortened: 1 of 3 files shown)


antocuni:
that's a good question: do we want to showcase everything inside proof-of-concept, or to keep it more minimal just to show people how an hpy module look like?

I think I prefer to keep it minimal: if we start to include *every* feature, it will quickly become a monster, and if you want to see examples of features you can always look at tests (maybe it's even a good idea to insert links to tests inside the documentation, when we will write it :)).

@hodgestar @mattip @rlamy @fangerer: do you have opinions on this?

mattip:
I think having a simple minimal complete example is preferable.

fangerer:
I do also agree with @antocuni and @mattip : IMO we should keep it minimal.
However, I also think that having some more examples would be helpful. Maybe we can just use the tests for this purpose? Those are currently hard to read because they are using templates but we could have a tool that generates the C sources.

cklein:
Yes, keeping the poc small sounds reasonable. Generating code from tests is a good idea, it's actually one of my main references when looking for an example.

antocuni:
this looks a great idea. We are ALREADY generating a complete .c files to run the tests, so it's just a matter of saving them somewhere which can be linked from the docs.
And then we also need to write the docs of course :man_facepalming:

PR #202, opened 2021-04-08
Update roadmap.

hodgestar:
We implemented many new things and the roadmap got out of date. :)

### DIFF
--- a/docs/overview.rst
+++ b/docs/overview.rst
@@ -285,7 +285,7 @@ been implemented yet:
   - handles are intended to be short-lived, but sometimes one needs a
     long-lived reference to a Python object. In HPy, we call this
     long-lived reference an `HPyField`, but we still need to implement it.
-    We also need `HPy_Store` and `HPy_Load` to save and load these fields.
+    We also need `HPy_Store` and `HPy_Load` to save and load these fields. This will allow alternative implementations to use a moving GC.
 
   - add C-level module state. Often an extension needs module state that is
     accessible from C (e.g. if a module implements a new type `ArrayType`,
--- a/docs/overview.rst
+++ b/docs/overview.rst
@@ -282,6 +282,19 @@ been implemented yet:
     is to work on the "hard" features to prove that the HPy approach works, and
     we will port new functions as needed
 
+  - handles are intended to be short-lived, but sometimes one needs a
+    long-lived reference to a Python object. In HPy, we call this
+    long-lived reference an `HPyField`, but we still need to implement it.
+    We also need `HPy_Store` and `HPy_Load` to save and load these fields.
+
+  - add C-level module state. Often an extension needs module state that is
+    accessible from C (e.g. if a module implements a new type `ArrayType`,
+    many of the module methods written in C may need access to `ArrayType`)
+    but HPy does not have convenient support for this yet.
+
+  - improve the debug mode so that it returns a C traceback showing where
+    a handle was leaked or a closed handle was used.
+
   - there is no integration with Cython. The medium-term plan is to extend
     Cython to automatically generate HPy-compatible C code.
 
--- a/docs/overview.rst
+++ b/docs/overview.rst
@@ -277,7 +277,7 @@ However, there is still a long road before HPy is usable for the general
 public. In particular, the following features are on our roadmap but have not
 been implemented yet:
 
-  - mnay of the original Python/C functions have not been ported to
+  - many of the original Python/C functions have not been ported to
     HPy yet. Porting most of them is straighforward, so for now the priority
     is to work on the "hard" features to prove that the HPy approach works, and
     we will port new functions as needed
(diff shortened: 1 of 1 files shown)


hodgestar:
@antocuni @rlamy @fangerer I should write some more about what we're working on at the moment and what is planned for the near future. Any suggestions?

docs/overview.rst
```diff
@@ -238,55 +237,53 @@ You have two choices:
...
-    will port new functions as needed
-
-  - the debug mode simply does not exist (yet!)
-
-  - there is no standard/easy way to integrate HPy into a
-    ``distutils/setuptools`` based workflow. The only way to compile HPy
-    extensions right now is to manually clone the git repo and tweak your
-    ``setup.py``. Eventually, we plan to offer a workflow which integrates
-    seamlessly with ``pip``, ``setuptools``, etc.
+  - mnay of the original Python/C functions have not been ported to
```
gracinet, line 280: Typo here (should be "many")

docs/overview.rst
```diff
@@ -238,55 +237,53 @@ You have two choices:
...
-following milestones have been reached:
+HPy is still in the early stages of development, but many big pieces are
+already in place. As on April 2021, the following milestones have been reached:
+
+  - one can write extensions which expose module-level functions, with all
+    the various kinds of calling conventions.
+
+  - there is support for argument parsing (i.e. the equivalents of
+    `PyArg_ParseTuple` and `PyArg_ParseTupleAndKeywords`).
+
```
antocuni, line 248: in the part about future work we should mention the "argument-clinic-like" way of specifying calling convention, which among the other things will allow the JITs and Cython to directly call the impl function, bypassing the argument parsing entirely

docs/overview.rst
```diff
@@ -238,55 +237,53 @@ You have two choices:
...
-    ``setup.py``. Eventually, we plan to offer a workflow which integrates
-    seamlessly with ``pip``, ``setuptools``, etc.
+  - many of the original Python/C functions have not been ported to
+    HPy yet. Porting most of them is straighforward, so for now the priority
+    is to work on the "hard" features to prove that the HPy approach works, and
+    we will port new functions as needed
 
   - there is no integration with Cython. The medium-term plan is to extend
-    Cython to automatically generate HPy-compatible C code
+    Cython to automatically generate HPy-compatible C code.
```
antocuni, line 288: more future work:
- `HPyField`/`HPy_Load`/`HPy_Store`
- improving the debug mode so that it also tells you the C traceback of the place who leaked a handle or used a closed one
- design a way to define and access a C-level module state

antocuni:
> @antocuni @rlamy @fangerer I should write some more about what we're working on at the moment and what is planned for the near future. Any suggestions?

I put some suggestions inside the review, thank you for doing this.
I think that thinking about the roadmap and the next step is a great idea not only for the general pubic but also for ourselves.
Maybe we should do a call to discuss this better?

fangerer:
> @antocuni @rlamy @fangerer I should write some more about what we're working on at the moment and what is planned for the near future. Any suggestions?

Also thank you for updating the overview. I would maybe add that we plan to auto-generate examples from our tests (this idea came up during a PR that would have extended the proof of concept).
Other than that, I don't know exactly what we is planned for the near future. You already mentioned the _Cython backend_. Everything else depends a bit on what we need for NumPy, right?
I agree with @antocuni : We should maybe do another call to discuss the next steps.

fangerer (approved):

docs/overview.rst
hodgestar: Added.

hodgestar:
I added the items @antocuni suggested. I agree that it would be good to have a call to talk about the roadmap.

docs/overview.rst
```diff
@@ -238,55 +237,66 @@ You have two choices:
...
 
-  - there is no standard/easy way to integrate HPy into a
-    ``distutils/setuptools`` based workflow. The only way to compile HPy
-    extensions right now is to manually clone the git repo and tweak your
-    ``setup.py``. Eventually, we plan to offer a workflow which integrates
-    seamlessly with ``pip``, ``setuptools``, etc.
+  - improve the debug mode so that it returns a C traceback showing where
+    a handle was leaked or a closed handle was used.
 
   - there is no integration with Cython. The medium-term plan is to extend
```
paugier, line 298: Other Python compilers could also be mentioned here. When HPy will be usable with Numpy, it will also be good to extend/modify [Pythran](https://github.com/serge-sans-paille/pythran/) to generate HPy code. Note that since Pythran C++ only uses the CPython C-API for converting Python object to pure C++ objects, it should be much simpler than extending Cython (yes, also much less useful given how much Cython is used everywhere!).

antocuni: I agree it might be an interesting goal. I am not sure about putting it on the roadmap though, because AFAIK nobody of the core team is going to work on that anytime soon

antocuni (approved):

PR #206, opened 2021-04-11
Resolve unused function warning

cklein:
It adds a cpp conditional to not include the function `linked_item_sanity_check` when `NDEBUG` is not set.
The function is only used when `NDEBUG` is set and therefore causes a compiler warning for non-debug builds.


hodgestar (approved):
Thanks!

PR #207, opened 2021-04-12
Upgrade infer to 1.1.0

thedrow:
infer 1.0 removed their dependency on Python 2 which makes our build environment easier to maintain.
This, on top of various bug fixes, should be reason enough to upgrade.


hodgestar (approved):
Thanks for doing this. Looks good to me!

PR #208, opened 2021-04-14
Extend PYTHONPATH instead of overwriting it

stefanor:
Allows `test_hpyerr` to pass when using `PYTHONPATH` to point to a built hpy C extension.


hodgestar (approved):
Thanks!

PR #210, opened 2021-05-05
Rename DHPy_wrap into DHPy_open.

antocuni:
This is better because the old DHPy_wrap was asymmetric
w.r.t. DHPy_unwrap. Now it is clearer that DHPy_open must be followed by
DHPy_close.

This PR is against master, but I plan to backport/cherry-pick it also to `release/0.0.2`


hpy/debug/src/debug_internal.h
```diff
@@ -36,6 +35,21 @@
...
+
+     - DHPy_close() moves a DHPy from info->open_handles to info->closed_handles
+
+     - if closed_handles is too big, the oldest DHPy is freed by DHPy_free()
+
+     - to allocate memory for a new DHPy, DHPy_open() does the following:
+
+         * if closed_handles is empty, it malloc()s memory for a new DHPy
+
+         * if it's not empty, it reuses the memory of the oldest DHPy in the
```
hodgestar, line 51: This description makes it sound like `DHPy_open` will always use a closed handle even if there is only one handle in `info->closed_handles`? I sort of expected `DHPy_open` to always malloc a new handle and then for the handle to be freed when `info->closed_handles` eventually gets too big.

antocuni: you are correct,t the comment was wrong. I tried to fix it in 2b08586

hodgestar (approved):

rlamy (approved):
+1

PR #212, opened 2021-06-08
Unicode building API survey

antocuni:
Before designing an API for building bytes and unicode objects, I though it was a good idea to collect into one document all the info about the current CPython API and their usage in the real world.

Once this PR is merged, I'll open an issue to discuss various possible designs (the alternative was to include this whole document into the issue but it was way too long).

### DIFF
--- a/docs/misc/str-builder-api.rst
+++ b/docs/misc/str-builder-api.rst
@@ -324,25 +324,30 @@ write directly to it:
     buf[2] = 'l';
     ...
 
-Another way to get a pointer to the raw-buffer is to call
-``PyUnicode_DATA()``, which returns a ``void *``. However, since the type and
-kind of the buffer is unknown, you can't write directly into it, so there are
-basically two options:
+The other way to get a pointer to the raw-buffer is to call
+``PyUnicode_DATA()``, which returns a ``void *``: the only reasonable way to
+write something in this buffer is to ``memcpy()`` the data from another
+``str`` buffer of the same kind. This technique is used for example by
+`CPython's textio.c
+<https://github.com/antocuni/cpython/blob/7b3ab5921fa25ed8b97b6296f97c5c78aacf5447/Modules/_io/textio.c#L344>`_.
 
-1. use ``PyUnicode_WRITE()``: this seems to be used only by code emitted by
-   Cython and sip.
-
-2. ``memcpy()`` memory from another ``str`` buffer of the same kind. This
-   technique is used for example by
-   `CPython's textio.c <https://github.com/antocuni/cpython/blob/7b3ab5921fa25ed8b97b6296f97c5c78aacf5447/Modules/_io/textio.c#L344>`_. Outside CPython, the only usage I
-   could find is by
-   `Cython <https://github.com/hpyproject/top4000-pypi-packages/blob/0cd919943a007f95f4bf8510e667cfff5bd059fc/etop1000/0158-Cython-0.29.23/Cython/Utility/StringTools.c#L857>`_.
-
-In addition to the Cython code linked above, there are 29 usages of
-``PyUnicode_DATA`` in the Top 4000 packages, all of them using it to call
-``PyUnicode_READ``.
+Outside CPython, the only usage of this technique is inside cython's helper
+function `__Pyx_PyUnicode_Join
+<https://github.com/hpyproject/top4000-pypi-packages/blob/0cd919943a007f95f4bf8510e667cfff5bd059fc/top1000/0158-Cython-0.29.23/Cython/Utility/StringTools.c#L857>`_.
 
 This probably means that we don't need to offer untyped raw-buffer writing for
 HPy. If we really need to support the ``memcpy`` use case, we can probably
 just offer a special function in the builder API.
 
+PyUnicode_WRITE, PyUnicode_WriteChar
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Outside CPython, ``PyUnicode_WRITE()`` is used only inside Cython's helper
+functions
+(`1 <https://github.com/hpyproject/top4000-pypi-packages/blob/0cd919943a007f95f4bf8510e667cfff5bd059fc/top1000/0158-Cython-0.29.23/Cython/Utility/StringTools.c#L865>`_,
+`2 <https://github.com/hpyproject/top4000-pypi-packages/blob/0cd919943a007f95f4bf8510e667cfff5bd059fc/top1000/0158-Cython-0.29.23/Cython/Utility/StringTools.c#L914-L926>`_).
+Considering that Cython will need special support for HPy anyway, this means
+that we don't need an equivalent of ``PyUnicode_WRITE`` for HPy.
+
+Similarly, ``PyUnicode_WriteChar()`` is used only once, inside
+`JPype <https://github.com/hpyproject/top4000-pypi-packages/blob/0cd919943a007f95f4bf8510e667cfff5bd059fc/etop1000/0546-JPype1-1.2.1/native/python/jp_pythontypes.cpp#L196>`_.
--- a/docs/misc/str-builder-api.rst
+++ b/docs/misc/str-builder-api.rst
@@ -18,7 +18,7 @@ Some terminology:
     * "unicode" or "unicode string" indicates ``str``
 
 .. note::
-   In this issue we are translating ``PyUnicode_*`` functions into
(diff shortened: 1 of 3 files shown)


docs/misc/str-builder-api.rst
```diff
@@ -0,0 +1,388 @@
...
+
+2. What are the constraints for alternative implementations and the problems
+   of the current C API.
+
+3. What are the patterns used to build string objects in the existing
+   extensions.
+
+Some terminology:
+
+    * "string" indicates either ``bytes`` or ``str`` objects aaa
```
mattip, line 17: Typo to check reviewer is attentive detected
```suggestion
    * "string" means both ``bytes`` and ``str`` objects
```

docs/misc/str-builder-api.rst
```diff
@@ -0,0 +1,388 @@
...
+"raw-buffer API".
+
+Unicode
+~~~~~~~
+
+Similarly to ``bytes``, there are several ways to build a ``str``:
+
+.. code-block:: c
+
+    PyObject *PyUnicode_FromString(const char *u);
```
mattip, line 60:
```suggestion
    PyObject* PyUnicode_FromString(const char *u);
```

docs/misc/str-builder-api.rst
```diff
@@ -0,0 +1,388 @@
...
+    PyObject* PyUnicode_FromFormat(const char *format, ...);
+    PyObject* PyUnicode_New(Py_ssize_t size, Py_UCS4 maxchar);
+
+
+.. note::
+   ``PyUnicode_FromString{,AndSize}`` take an UTF-8 string in input
+
+The following functions are used to initialize an uninitialed object, but I
+could not find any usage of them outside CPython itself, so I think they can
+be safely ignored for now:
```
mattip, line 72: As another point of reference: PyPy does not implement them, and no issue has been reported that they are missing.

docs/misc/str-builder-api.rst
```diff
@@ -0,0 +1,388 @@
...
+- use standard C patterns such as ``*p++ = ...`` or similar.
+
+
+Problems and constraints
+------------------------
+
+``bytes`` and ``str`` are objects are immutable: the biggest problem of the
+current API boils down to the fact that the API allows to construct objects
+which are not fully initialized and to mutate them during a
+not-well-specificed "initialization phase".
```
mattip, line 190: Isn't this the goal of `_PyUnicode_Ready` in [PEP 393](https://www.python.org/dev/peps/pep-0393/)?

mattip:
Thanks. It is nice to see that many of the stranger and more difficult interfaces are rarely used.

docs/misc/str-builder-api.rst
antocuni: no, `PyUnicode_READY` is needed if you use the legacy (now deprecated) APIs to build unicode (e.g. `PyUnicode_FromUnicode`). `PyUnicode_READY` is also deprecated and will be removed in 3.12.

If you use `PyUnicode_New + PyUnicode_1BYTE_DATA` you don't need to call `PyUnicode_READY`.

hodgestar (approved):
Thanks for putting this together! I'm happy for this to be merged and to discuss details in follow up PRs or discussions.

My own thoughts:

* I'd like to keep separate APIs for unicode and bytes for clarity and because conceptually the two are very different and so might have very different implementations.

* I would really like an API that provides reasonable performance while hiding the underlying data storage. As a consumer of the API, I encountered the new unicode KIND system in CPython and then just deleted the C speedup code rather than try port it.

* It would be amazing if we could somehow provide a potentially zero copy API for bytes, because it enables some amazing things like reading a ZeroMQ packet off the network straight into a numpy array without any memory being copied (and this example already works with the existing C API).

* It would be good to have a zero copy API for Unicode too, but that is more complex because the underlying implementation could be UTF8, or UTF16, or a KIND system like CPython has.

</2c>

docs/misc/str-builder-api.rst
```diff
@@ -0,0 +1,390 @@
...
+   extensions.
+
+Some terminology:
+
+    * "string" means both ``bytes`` and ``str`` objects
+    * "unicode" or "unicode string" indicates ``str``
+
+.. note::
+   In this document we are translating ``PyUnicode_*`` functions into
+   ``HPyStr_*``. See issue #xxx for more discussion about the naming convention.
```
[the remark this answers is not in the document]
TeamSpen210: My reading of this is that HPy would use `HPyBytes_*` and `HPyStr_*` - use the Python name for the type since we don't need to distinguish from the Py2 string type.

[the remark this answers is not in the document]
antocuni: yes exactly, my proposal will be to have `HPyBytes_*` and `HPyStr_*`, as opposed to `HPyBytes_*` and `HPyUnicode_*`.
I'll open an issue for that so we can discuss the naming in a separate place than the API itself

docs/misc/str-builder-api.rst
```diff
@@ -0,0 +1,390 @@
...
+
+To initialize a ``str`` object using the raw-buffer interface, you need to get
+a pointer to the buffer.  The vast majority of code uses
+``PyUnicode_{1,2,4}BYTE_DATA`` to get a buffer of type ``Py_UCS{1,2,4}*`` and
+write directly to it:
+
+.. code-block:: c
+
+    PyObject *s = PyUnicode_New(size, 127);
+    Py_UCS1 *buf = PyUnicode_1BYTE_DATA(s);
```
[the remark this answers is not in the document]
antocuni: well no, my point is that in this specific example you know at compile time that the kind will be `1BYTE`, because you used `maxchar=127`.

antocuni:
> Thanks for putting this together! I'm happy for this to be merged and to discuss details in follow up PRs or discussions.
> 
> My own thoughts:
> 
> * I'd like to keep separate APIs for unicode and bytes for clarity and because conceptually the two are very different and so might have very different implementations.

yes, I agree. I think we will end up with an `HPyBytesBuilder` and an `HPyStrBuilder`, but it will be nice if we can design an API which is consistent between the two

> * I would really like an API that provides reasonable performance while hiding the underlying data storage. As a consumer of the API, I encountered the new unicode KIND system in CPython and then just deleted the C speedup code rather than try port it.

yes, but I fear that we can't forget the real-world usage. We want people to be able to port their code easily, don't we?

> * It would be amazing if we could somehow provide a potentially zero copy API for bytes, because it enables some amazing things like reading a ZeroMQ packet off the network straight into a numpy array without any memory being copied (and this example already works with the existing C API).

yes, and I agree that it's an use case that we should preserve.

> * It would be good to have a zero copy API for Unicode too, but that is more complex because the underlying implementation could be UTF8, or UTF16, or a KIND system like CPython has.

yes. I think that the best we can aim is to have an API in which you declare what is the kind of data you are going to have, and the implementation can either do a copy or a direct write depending on the details. E.g., if you use `HPyStrBuilder_UCS2` you will end up with no-copy on CPython but a copy on PyPy, but with a hypothetical `HPyStrBuilder_UTF8` you would end up with copy/decoding on CPython and no-copy on PyPy.

hodgestar:
> > * I would really like an API that provides reasonable performance while hiding the underlying data storage. As a consumer of the API, I encountered the new unicode KIND system in CPython and then just deleted the C speedup code rather than try port it.
> 
> yes, but I fear that we can't forget the real-world usage. We want people to be able to port their code easily, don't we?

Perhaps the KIND API can still exist but be deprecated and be a bit slower on implementations other than CPython?
 
> > * It would be good to have a zero copy API for Unicode too, but that is more complex because the underlying implementation could be UTF8, or UTF16, or a KIND system like CPython has.
> 
> yes. I think that the best we can aim is to have an API in which you declare what is the kind of data you are going to have, and the implementation can either do a copy or a direct write depending on the details. E.g., if you use `HPyStrBuilder_UCS2` you will end up with no-copy on CPython but a copy on PyPy, but with a hypothetical `HPyStrBuilder_UTF8` you would end up with copy/decoding on CPython and no-copy on PyPy.

Maybe there can be a way to ask the implementation "what is the best format to supply the characters in"? Not 100% sure such a thing is a good idea. Perhaps it only opens a can of worms.

antocuni:
> Perhaps the KIND API can still exist but be deprecated and be a bit slower on implementations other than CPython?

I think this will be better discussed in the proper issue, once I'll create it :)

fangerer (approved):
nice summary, thank you

PR #218, opened 2021-07-08
Fix AttributeError: 'Extension' object has no attribute 'hpy_abi'

paugier:
Fix #215

With this patch, one can again build packages with both hpy and standard extensions.

PR #220, opened 2021-07-08
merge release/0.0.2 into master

antocuni:
This PR merges the changes I did in `release/0.0.2` into `master`. 
At the moment of writing the latest commit of this branch is tagged as `0.0.2rc4`, and the `sdist` is published on PyPI: please try to `pip install hpy==0.0.2rc4` and play a bit with it to check that it works as expected.
Once we finalize all the changes and this PR is approved, we can tag `0.0.2` and merge the PR to master.

Summary of changes:

**Debug mode**

  1. tweak the debug mode so that it can be used also in PyPy: 
  2. backport PR #210 
  3. tweak so that it works also with the ancient version of pytest used by PyPy
 
**setup.py / hpy.devel**

  4. Backport PR #218 
  5. make `--hpy-abi=universal` the default on non-CPython implementations
  6. rename the PyPI distribution to `hpy` instead of `hpy.devel`
  7. Exit early with a clear error message in case you try to `pip install hpy` on a non-cpython
  8. Add a `description` and a `long_description` for PyPI

**tests**

  9. add additional tests for the debug mode

About point (7), here is what happens if you try to `pip install` or to `setup.py install` on PyPy and GraalPython:

**PyPy**:
```
$ pypy setup.py install
ERROR: Cannot install and/or update hpy on this python implementation:
    sys.implementation.name == 'pypy'

This python implementation comes with its own version of hpy==0.0.2rc4

If you are trying to install hpy through pip, consider to put the
following in your requirements.txt, to make sure that pip will NOT
try to re-install it:
    hpy==0.0.2rc4
```

**GraalPython**:
```
ERROR: Cannot install and/or update hpy on this python implementation:
    sys.implementation.name == 'graalpython'

This python implementation comes with its own version of hpy,
but the exact version could not be determined.
```

**Hypotetical non-cpython implementation which does not support hpy**
```
ERROR: Cannot install and/or update hpy on this python implementation:
    sys.implementation.name == 'pypy'

This python implementation does not seem to support hpy:
(built-in module _hpy_universal not found).
Please contact your vendor for more informations.
```

@fangerer note that the message given for GraalPython is different that PyPy: this is because graalpython does not expose `_hpy_universal.get_version()`, maybe you should consider adding it.


hpy/devel/__init__.py
```diff
@@ -10,6 +11,9 @@
 # NOTE: this file is also imported by PyPy tests, so it must be compatible
 # with both Python 2.7 and Python 3.x
 
+DEFAULT_HPY_ABI = 'universal'
```
hodgestar, line 14: I'm not sure how I feel about the default ABI being different on different Python implementations. It surprised me a bit, and I'm worried it will continue to surprise people.

On the other hand, I can see that on CPython one might want the default to be `cpython` right now and on other Python implementations one would want the default to be something that works well there.

Happy for this to stay this way, just flagging it for thinking about if necessary.

hodgestar (approved):

hodgestar:
@antocuni Some conflicts need resolving (which is a bit odd) but the changes generally look sane to me.

hpy/devel/__init__.py
antocuni: I agree with your concerns. Having different defaults is sub-optimal. However, consider that:
1. On CPython, we surely want `cpython` by default, to avoid unwanted slowdowns. I don't want to risk that people starts tweeting that hpy is slow
2. On PyPy, we surely want `universal` by default, because `cpython` does not work at all (or if it does, it's utterly slow)
3. At the moment, people are already used to recompile extensions for every python version they need. So if they just continue to do that, they will get the "best" result on CPython/PyPy/GraalPython.
4. Eventually it will be nice to have `universal` by default, but at the moment it's a bit useless anyway, since we don't have a way to distribute them.

PR #225, opened 2021-07-18
Refactor and simplify the headers in hpy/devel/include

antocuni:
The current headers are a mess, this PR tries to simplify/clean/improve the situation.

Major points:

1. Less pollution of the `#include` "namespace". On master we put in the main include path `hpy.h`, `common/`, `cpython/` and `universal/`. Now we only have `hpy.h` and `hpy/`

2. Unify the Universal and CPython headers as much as possible. On master we have `hpy.h`, `universal/hpy.h` and `cpython/hpy.h`, and the latter two contains many duplicate or almost duplicate code. Now we have a single `hpy.h` which contains a lot of shared definitions, and `#includes` the appropriate files from `hpy/universal/*.h` and `hpy/cpython/*.h` as needed.

3. Kill the ugly hacks around `_HPy_IMPL_NAME` and `implementation.h`. On master, `implementation.h` is shared between the CPython and Universal modes and we use macro tricks to adapt it to both cases. But it turns out that the vast majority of it is autogenerated, and so it's much easier to autogenerate two different versions directly. There were also a few non-autogenerated functions in the old `implementation.h` but it was weird anyway: we already have a way to share `ctx_*` implementations betweent he two ABI modes, so we can just use it also in this case.

4. `HPyAPI_FUNC` and `HPyAPI_STORAGE` were a mess and were not used consistently. It was needed because of the hack in point (3), but now that it's gone we can more easily distinguish the `HPy_*` trampolines from the actual `ctx_*` implementations. `HPyAPI_RUNTIME_FUNC` is also no longer needed.  **NOTE**: on master we use `HPyAPI_FUNC(int) foo(void);`, but here I chose to use `HPyAPI_FUNC int foo(void);`. The reason is that it's much easier to autogenerate, because to emit the former syntax you need to mess with `pycparser` nodes.

The final result is still not perfect. In particular, `hpy/cpython/misc.h` contains too much stuff, and most of it could be probably autogenerated. However, it can still be done in follow-up PRs.


hpy/devel/include/hpy.h
```diff
@@ -1,10 +1,115 @@
...
+
+   * in Universal ABI mode, the meaning of ._i depends on the implementation:
+
+       - CPython (i.e., the code in hpy/universal/src/): ._i is the bitwise
+         invert of a PyObject*
+
+       - PyPy: ._i is an index into a list
+
+       - GraalPython: ???
+
```
antocuni, line 52: @fangerer @timfel do you want to fill in few words for GraalPython?

hpy/devel/include/hpy/runtime/argparse.h
```diff
@@ -3,10 +3,10 @@
 
 #include "hpy.h"
 
-HPyAPI_RUNTIME_FUNC(int)
+_HPy_HIDDEN int
```
hodgestar, line 6: I'm a bit uncomfortable just labeling these `_HPy_HIDDEN` because it doesn't explain why these functions are marked hidden. What differentiates them from `HPyAPI_IMPL`?  I don't know what the name should be, buI feel we *should* have a name for them so it's clear what sort of thing they are.

setup.py
```diff
@@ -39,8 +39,12 @@
 if 'HPY_DEBUG' in os.environ:
     # -fkeep-inline-functions is needed to make sure that the stubs for HPy_*
     # functions are available to call inside GDB
-    EXTRA_COMPILE_ARGS = ['-g', '-O0', '-UNDEBUG',
-                          '-fkeep-inline-functions']
+    EXTRA_COMPILE_ARGS = [
+        '-g', '-O0', '-UNDEBUG',
+        '-fkeep-inline-functions',
+        #'-Wfatal-errors',    # stop after one error (unrelated to warnings)
+        #'-Werror',           # turn warnings into errors
```
hodgestar, line 46: Was the commented out code left in unintentionally? Or do we want them to be on by default when `HPY_DEBUG` is set? Or something else? They do seem useful.

hodgestar (approved):
Looks like a good start on reorganizing to me. I left some questions / comments, but am +1 on the changes overall.

hpy/devel/include/hpy.h
```diff
@@ -1,10 +1,115 @@
...
+typedef struct _HPyContext_s HPyContext;
+
+#ifdef HPY_UNIVERSAL_ABI
+    typedef intptr_t HPy_ssize_t;
+    typedef intptr_t HPy_hash_t;
+#else
+/*  It would be nice if we could include hpy.h WITHOUT bringing in all the
+    stuff from Python.h, to make sure that people don't use the CPython API by
+    mistake. How to achieve it, though? */
```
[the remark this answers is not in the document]
antocuni: note that the comment was already there, it's not new in this PR :).
That said, I think that it's not possible to achieve what the comment says, and the only reasonable way to be sure that we are not using `Python.h` functions by mistake is indeed to compile with `--hpy-abi=universal`.

But note that this is a broader topic. Eventually we need to introduce the "hybrid ABI" which allows you to use the `legacy_*` features and things like `HPy_{From,As}PyObject`, and produces files in the form of `foo.hpy-cpython-38-x86_64-linux-gnu.so`. Let's discuss it on another issue and/or PR.

hpy/devel/include/hpy/runtime/argparse.h
antocuni: They are marked `_HPy_HIDDEN` because it's a good practice, for the very same reason why you mark most other functions as `static`, i.e. because you want export nothing by default unless you explicitly export the symbol, and C has the wrong default :man_shrugging: .
From this point of view it has nothing to do with the fact that they are or are not "API functions". E.g. we do the very same for `create_method_defs`:
https://github.com/hpyproject/hpy/blob/9dd30240456abe5b1d9eb489dd0d2f9a5784c9a4/hpy/devel/src/runtime/ctx_type.c#L90-L93

That said, it might be a good idea to mark them somehow, to be self-documenting. What about `HPyAPI_HELPER` ?

setup.py
antocuni: They are useufl -- and that's why I left them as comments -- but they don't generally work on all systems and/or compilers.
See 943a253 and 5d2c566

antocuni:
> Looks like a good start on reorganizing to me. I left some questions / comments, but am +1 on the changes overall.

thank you! I'll wait until we reach a consensus on the `HPyAPI_RUNTIME_FUNC` thing before merging

setup.py
hodgestar: I'm not a fan of commented out code, and especially when there isn't a comment saying why the code is there. My votes are either to add an explanation in a comment (+0.5) or to remove them (+0.5) or to add another flag (`HPY_DEBUG_FAILFAST`?) that activates them.

hpy/devel/include/hpy/runtime/argparse.h
hodgestar: I like `HPyAPI_HELPER`, although I think there are really two kinds of functions here:
* `HPyAPI_HELPER`s like `HPyArg_Parse`
* `HPyAPI_INTERNAL`s like `ctx_...`
with the difference being that helpers are intended to be used by writers of extensions, but never exported as symbols and internal functions are neither.

setup.py
antocuni: I think the comment is enough, no need to overengineer it. Added it in 01eaf4a

hpy/devel/include/hpy/runtime/argparse.h
antocuni: yes, I'm proposing to have:
1. `HPyAPI_FUNC`
2. `HPyAPI_IMPL` (what you call `HPyAPI_INTERNAL)
3. `HPyAPI_HELPER`

I like `IMPL` better than `INTERNAL` because "internal" sounds like an internal API which is not meant to be exposed to the user, i.e. things like CPython's  `_PyTuple_FromArray` which nowadays are in `Include/internal/*.h`.
Out `ctx` functions are different... they are not internal APIs, they are the implementation of user-facing APIs (but you need to call them through the `HPy_*` functions because you do different things depending on whether you are using the CPython or Universal ABI).

I agree that maybe we should write some internal doc to explain all of this

antocuni: I introduced `HPyAPI_HELPER` and a big explanatory comment in 195be06

hpy/devel/include/hpy.h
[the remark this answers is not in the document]
hodgestar: Happy to leave this for another issue / PR.

hpy/devel/include/hpy/runtime/argparse.h
hodgestar: I like the FUNC / IMPL / HELPER naming and woot on the big comment! :D

setup.py
hodgestar: Thanks!

PR #228, opened 2021-07-27
Introduce HPyField

### LINKED ISSUE #9: Custom object types
We need to think about custom object types with a structure declared in C.  We don't want to repeat CPython's mistake about explicit destructors everywhere containing Py_DECREF() and a hack on top of that for reference cycles.  There must be a better solution, which would also allow an implementation like PyPy's which doesn't need to walk most dead objects at all.

I would suggest to declare local variables and arguments of type "POD" (Python Object Descriptor, or handles), and use a different "PORef" type for fields in objects (Python Object Reference).

`PORef_Store(PORef *pref, POD o);     /* '*pref = o;' with conversion */`
`PORef_Clear(PORef *pref);      /* '*pref = NULL;' */`
`PORef_Load(PORef *pref);      /* returns a new POD, which must be closed */`

Then it would be mandatory to write an equivalent of tp_traverse to list all PORefs in an object.  The main difference between POD and PORef is that a POD must be closed when no longer in use, but not a PORef.  A PORef is alive if and only if it is stored in an object, the tp_traverse of this object is called, and that function lists the PORef.

Of course the implementation of both POD and PORef are `PyObject *` under the hood on CPython.  On PyPy a POD could be an integer index in some global list, but a PORef could be a real pointer which might be internally changed during tp_traverse.

antocuni:
Fixes #9


docs/porting-guide.rst
```diff
@@ -1,6 +1,115 @@
...
+
+  1. if you have a field of type ``HPyField``, you always **MUST** provide a
+     ``tp_traverse``. This is needed so that a moving GC can track the
+     relevant areas of memory;
+
+  2. ``tp_clear`` does not exist. On CPython, ``HPy`` automatically generates
+     one for you, by using ``tp_traverse`` to know which are the fields to
+     clear. Other implementations are free to ignore it, if it's not needed;
+
+  3. ``HPy_TPFLAGS_GC`` is still needed, especially on CPython. If you don't
```
hodgestar, line 93: Could we do away with ``HPy_TPFLAGS_GC`` entirely and just use the presence to `tp_traverse` to indicate that the object has `HPyField`s to track? On CPython we can then set ``Py_TPFLAGS_GC`` based on `tp_traverse` being set.

Perhaps this simplification has some backwards compatibility issues? E.g. code in extensions that wants to check whether ``HPy_TPFLAGS_GC`` is set?

hpy/devel/include/hpy/cpython/misc.h
```diff
@@ -194,6 +203,27 @@ HPyAPI_FUNC void HPy_Close(HPyContext *ctx, HPy handle)
     Py_XDECREF(_h2py(handle));
 }
 
+HPyAPI_FUNC HPyField HPyField_Store(HPyContext *ctx, HPy h)
```
hodgestar, line 206: I am wondering what happens with code like:
```
type_struct.x = HPyField_Store(ctx, h);
```
Is it possible that there is briefly a temporary variable somewhere containing an HPyField that the GC doesn't know about? Or the temptation for the user to write `if (....) { a = HPyField_Store(ctx, h); } else { a = HPyField_Store(ctx, h) } type_struct.x = a`?

It feels safer if `HPyField_Store` takes the address to store the address in directly. Then the implementation can do whatever is necessary to ensure the pointer is definitely stored safely.

hpy/tools/autogen/public_api.h
```diff
@@ -295,6 +297,39 @@ int HPyTracker_Add(HPyContext *ctx, HPyTracker ht, HPy h);
 void HPyTracker_ForgetAll(HPyContext *ctx, HPyTracker ht);
 void HPyTracker_Close(HPyContext *ctx, HPyTracker ht);
 
+/* HPyField */
+
+/* TODO / Open questions:
+
+  - what is the semantics w.r.t. overwriting HPyFields? I.e.:
+
+        pair->a = HPyField_Store(ctx, h1);
```
hodgestar, line 306: If  we used `HPyField_Store(ctx, h1, &pair->a)` then `HPyField_Store` can call clear on whatever was in `a` before overwriting it? Or perhaps raise an error if it was not cleared? That definitely seems like a good feature for debug mode if no where else.

hpy/tools/autogen/public_api.h
```diff
@@ -295,6 +297,39 @@ int HPyTracker_Add(HPyContext *ctx, HPyTracker ht, HPy h);
...
+
+         Py_DECREF(pair->a);
+         pair->a = obj1;
+         Py_INCREF(obj1);
+
+    Option 2 is described in issue #9, and requires to do something like this:
+
+        HPyField_Store(ctx, &pair->a, h1);
+
+    So that HPyField_Store can do the DECREF by itself. This is uglier, and
```
hodgestar, line 326: Ah, I see my suggestion is already described in #9. I will keep the comments here because I do think it's a good idea. :)

hodgestar (commented):
This is not a complete review, but I did read through and make some comments. The HPyField implementation looks pretty good so far.

docs/porting-guide.rst
```diff
@@ -1,6 +1,115 @@
...
+moved.
+
+Back to ``HPy`` vs ``HPyField``:
+
+  * ``HPy`` handles must be used for all C local variables, function arguments
+    and function return values. They are supposed to be short-lived and closed
+    as soon as they are no longer needed. The debug mode will report a
+    long-lived ``HPy`` as a potential memory leak.
+
+  * In PyPy, they are implemented using an indirection: they are indexes
```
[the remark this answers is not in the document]
antocuni: maybe, but I don't know the details of GraalPython. If @fangerer wants to add, it's very welcome :)

docs/porting-guide.rst
antocuni: we are discussing it right now. The problem is:
1. `Py_TPFLAGS_GC` is costly on CPython, because you need to allocate extra space for the GC header
2. on HPy you will need to add `tp_traverse` also for objects which currently don't need it on CPython

So, if we automatically add the flag based on the presence of `tp_traverse`, we end up with cases in which you pay a penalty which you don't need.

One option is to do the opposite: use the flag by default and add another flag to disable it, as an optimization. But it looks a bit obscure, especially for people who are already used to the current behavior.

hpy/devel/include/hpy/cpython/misc.h
antocuni: I think this is a comment about old code? The new signature for `HPyField_Store` does indeed what you suggest, see `public_api.h`:
```c
void HPyField_Store(HPyContext *ctx, HPyField *target_field, HPy h);
```

antocuni: ouch, my fault. I had local commits which weren't pushed :(
Anyway, if you look at what I just pushed, you see I have already addressed this :)

hpy/tools/autogen/public_api.h
antocuni: this comment no longer applies to the current implementation

docs/porting-guide.rst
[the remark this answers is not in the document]
fangerer: Basically, we do the same as PyPy. We have a Java object array and the indices into this array are then the HPy handles.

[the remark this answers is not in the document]
antocuni: cool, I mentioned GraalPython in b5da385

fangerer (commented):
Overall, looks good to me but I still have one question: The API function `HPyField_Store` already requires the used to provide the target object. If I understood correctly, this is the object that then owns the reference. Since we can now already keep track of which object contains which references, is it mandatory that the GC calls `tp_traverse` to determine the referents?
Since we don't have control over the GC in Graalpython, we cannot easily do that.

I assume we need `tp_traverse` for two reasons:
1. because of CPython's _GC_.
2. it's an optimization. Keeping track at the time of `HPyField_Store` imposes higher memory footprint since the runtime needs to store the objects **AND** write something to the HPyField.

If so, I suggest that we also state that `tp_traverse` might not be called (but must still be provided) and we could do some sanity checks in the debug mode.

hpy/devel/src/runtime/ctx_type.c
```diff
@@ -604,11 +745,80 @@ ctx_Type_GenericNew(HPyContext *ctx, HPy h_type, HPy *args, HPy_ssize_t nargs, H
...
     return _h2py(h);
 }
+
+/* ~~~ call_traverseproc_from_trampoline ~~~
+   This is used to implement tp_traverse.
+
+   From the HPy user point of view, the visitor function passed to tp_traverse
+   takes a ctx and an HPyField*:
+
+       typedef int (*HPyFunc_visitproc)(HPyContext *, HPyField *, void *);
```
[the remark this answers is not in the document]
antocuni: yes, correct. We decided that `tp_traverse` will be very low-level (like `tp_destroy`) and you won't be able to do anything "complex" from it. Basically, you should just call `HPy_VISIT()` and nothing more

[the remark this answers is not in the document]
antocuni: yes, correct. We decided that `tp_traverse` will be very low-level (like `tp_destroy`) and you won't be able to do anything "complex" from it. Basically, you should just call `HPy_VISIT()` and nothing more

hpy/tools/autogen/public_api.h
```diff
@@ -295,6 +294,42 @@ int HPyTracker_Add(HPyContext *ctx, HPyTracker ht, HPy h);
...
+   contains uninitialized values, because it would try to decref a dangling
+   pointer.
+
+   Note that HPy_New automatically zeroes the memory it allocates, so
+   everything works well out of the box. In case you are using manually
+   allocated memory, you should initialize the HPyField to HPyField_NULL.
+
+   Note the difference:
+
+     - ``obj->f = HPyField_NULL``: this should be used only to inizialize
```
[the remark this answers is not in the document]
antocuni: thank you, fixed

docs/porting-guide.rst
```diff
@@ -1,6 +1,116 @@
...
+
+
+However, if you know that your ``PyObject *`` fields will contain only
+"atomic" types, you can avoid these steps.
+
+In HPy the rules are slightly different:
+
+  1. if you have a field of type ``HPyField``, you always **MUST** provide a
+     ``tp_traverse``. This is needed so that a moving GC can track the
+     relevant areas of memory. However, you **MUST NOT** rely on
```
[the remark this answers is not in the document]
antocuni: I think it's good enough for now

PR #230, opened 2021-08-11
Add Github Actions CI, including release CI

### LINKED ISSUE #223: Improve our release process
Currently we don't have any automated release process: hpy releases (such as the upcoming 0.0.2) are manually uploaded using `twine`, and we upload only the `sdist`.

It would be nice to:
  - have an automated system which does a release whenever we push a relevant tag
  - build&upload wheels

dralley:
closes #223

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,277 @@
+---
+name: Continuous Integration tests
+
+on: [pull_request]
+
+jobs:
+  main_tests:
+    name: Main tests
+    runs-on: ${{ matrix.os }}
+    continue-on-error: true
+    strategy:
+      # Duplicate changes to this matrix to 'poc_tests'
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: ['3.9']
+        # Possible values: ['2.7', '3.6', '3.7', '3.8', '3.9', '3.10.0-rc.2']
+        include:
+          - os: ubuntu-latest
+            python-version: '3.6'
+          - os: ubuntu-latest
+            python-version: '3.7'
+          - os: ubuntu-latest
+            python-version: '3.8'
+          ## Expected failure
+          # - os: ubuntu-latest
+          #   python-version: '3.10.0-rc.2'
+
+    steps:
+      - uses: actions/checkout@v2
+
+      # - template: azure-templates/ccache.yml
+      #   parameters:
+      #     pythonVersion: $(python.version)
+      # - template: azure-templates/python.yml
+      #   parameters:
+      #     pythonVersion: $(python.version)
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install/Upgrade Python dependencies
+        run: python -m pip install --upgrade pip wheel
+
+      - name: Build
+        run: python -m pip install .
+
+      - name: Run tests
+        run: |
+          python -m pip install pytest pytest-xdist
+          python -m pytest -n auto test/
+
+
+  poc_tests:
+    name: Proof of concept tests
+    runs-on: ${{ matrix.os }}
+    continue-on-error: true
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: ['3.9']
+        include:
+          - os: ubuntu-latest
+            python-version: '3.6'
+          - os: ubuntu-latest
+            python-version: '3.7'
+          - os: ubuntu-latest
+            python-version: '3.8'
+          ## Expected failure
+          # - os: ubuntu-latest
+          #   python-version: '3.10.0-rc.2'
+
+    steps:
+      - uses: actions/checkout@v2
+
+      # - template: azure-templates/ccache.yml
+      #   parameters:
+      #     pythonVersion: $(python.version)
+      # - template: azure-templates/python.yml
+      #   parameters:
+      #     pythonVersion: $(python.version)
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install/Upgrade Python dependencies
+        run: python -m pip install --upgrade pip wheel
+        shell: bash
+
+      - name: 'Test setup.py --hpy-abi=cpython bdist_wheel'
+        run: proof-of-concept/test_pof.sh wheel cpython
+        shell: bash
+
+      - name: 'Test setup.py --hpy-abi=universal bdist_wheel'
+        run: proof-of-concept/test_pof.sh wheel universal
+        shell: bash
+
+      - name: 'Test setup.py --hpy-abi=cpython install'
+        run: proof-of-concept/test_pof.sh setup_py_install cpython
+        shell: bash
+
+      - name: 'Test setup.py --hpy-abi=universal install'
+        run: proof-of-concept/test_pof.sh setup_py_install universal
+        shell: bash
+
+      - name: 'Test setup.py --hpy-abi=cpython build_ext --inplace'
+        run: proof-of-concept/test_pof.sh setup_py_build_ext_inplace cpython
+        shell: bash
+
+      - name: 'Test setup.py --hpy-abi=universal build_ext --inplace'
+        run: proof-of-concept/test_pof.sh setup_py_build_ext_inplace universal
+        shell: bash
+
+
+  valgrind_tests:
+    name: Valgrind tests
+    runs-on: 'ubuntu-latest'
+    steps:
+      - uses: actions/checkout@v2
+
+    # - template: azure-templates/ccache.yml
+    # - template: azure-templates/python.yml
+
+      - name: Install / Upgrade system dependencies
+        run: sudo apt update && sudo apt install -y valgrind
+
+      - name: Install / Upgrade Python dependencies
+        run: python -m pip install --upgrade pip wheel
+
+      - name: Build
+        run: python -m pip install .
+
+      - name: Run tests
+        run: |
+          python -m pip install pytest pytest-valgrind
+          make valgrind
+
+
+  build_docs:
+    name: Build documentation
+    runs-on: 'ubuntu-latest'
+    steps:
+      - uses: actions/checkout@v2
+
+      # - template: azure-templates/python.yml
+
+      - name: Install / Upgrade system requirements
+        run: sudo apt update && sudo apt install -y libclang-10-dev
+
+      - name: Install / Upgrade Python requirements
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install -r docs/requirements.txt
+
+      - name: Build docs
+        run: |
+          cd docs;
+          python -m sphinx -T -W -E -b html -d _build/doctrees -D language=en . _build/html
+
+
+  c_tests:
+    name: C tests
+    runs-on: 'ubuntu-latest'
+    steps:
+      - uses: actions/checkout@v2
+      - run: make -C c_test
+
+
+  check_autogen:
+    name: Check autogen
+    runs-on: 'ubuntu-latest'
+    steps:
+      - uses: actions/checkout@v2
+
+    # - template: azure-templates/python.yml
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'
+
+      - name: Install/Upgrade Python dependencies
+        run: python -m pip install --upgrade pip wheel
+
+      - name: Install dependencies
+        run: |
+          pip install pycparser==2.20
+          pip install py==1.8.0
+          pip install packaging==19.2
+          pip install attrs==19.3.0
+      - name: make autogen
+        run: |
+          make autogen
+          if [ -z "$(git status --porcelain)" ]; then
+              # clean working copy
+              echo "Working copy is clean, everything ok"
+          else
+              # Uncommitted changes
+              echo "ERROR: uncommitted changes after running make autogen"
+              echo "git status"
+              git status
+              echo
+              echo "git diff"
+              git diff
+              exit 1
+          fi
+
+
+  check_py27_compat:
+    name: Check Python 2.7 compatibility
+    runs-on: 'ubuntu-latest'
+    steps:
+      - uses: actions/checkout@v2
+
+    # - template: azure-templates/python.yml
+    #   parameters:
+    #     pythonVersion: "2.7"
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '2.7'
+
+      - name: Install/Upgrade Python dependencies
+        run: python -m pip install --upgrade pip wheel
+
+      - name: check_py27_compat.py
+        run: |
+          python -m pip install pytest pathlib
+          python test/check_py27_compat.py
+
+
+  # cpp_check:
+  #   runs-on: 'ubuntu-latest'
+  #   steps:
+
+  #   # - template: azure-templates/cppcheck.yml
+  #   # - template: azure-templates/python.yml
+
+  #   - name: Set up Python
+  #     uses: actions/setup-python@v2
+  #     with:
+  #       python-version: '3.x'
+
+  #   - name: Run CppCheck
+  #     run: make cppcheck
+
+
+  infer:
+    name: Infer static analysis
+    runs-on: 'ubuntu-latest'
+    steps:
+      - uses: actions/checkout@v2
+
+      # - template: azure-templates/python.yml
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'
+
+      - name: Install/Upgrade Python dependencies
+        run: python -m pip install --upgrade pip wheel
+
+      - name: Install Infer
+        run: |
+          python -m pip install compiledb wheel;
+          VERSION=1.1.0; \
+          curl -sSL "https://github.com/facebook/infer/releases/download/v$VERSION/infer-linux64-v$VERSION.tar.xz" \
+          | sudo tar -C /opt -xJ && \
+          echo "/opt/infer-linux64-v$VERSION/bin" >> $GITHUB_PATH
+
+      - name: Run Infer
+        run: make infer
--- a/.github/workflows/release-pypi.yml
+++ b/.github/workflows/release-pypi.yml
@@ -1,0 +1,82 @@
+---
+name: Release to PyPI
+
+on:
+  release:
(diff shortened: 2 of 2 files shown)


dralley:
Here's the building PyPI wheels job passing (except for the upload part for obvious reasons): https://github.com/dralley/hpy/pull/1/checks?check_run_id=3296113677#step:4:10

I attempted to make artifact caching work but wasn't fully successful. Here's what I did for ccache, and you can see that it installed properly but for some reason the cache was not being used (no hits, but also no misses).

```yaml
      - name: ccache
        if: ${{ !startsWith(runner.os, 'Windows') }}
        uses: hendrikmuhs/ccache-action@v1
        with:
          key: ${{ matrix.os }}-${{ matrix.python-version }}
          max-size: 100M
```

https://github.com/dralley/hpy/pull/1/checks?check_run_id=3296448518#step:21:10

So, I pulled that back out, and just left the places where the template was being included commented.  Github actions does not currently support textual includes of steps.

And here's what I did for Python caching, but this strategy requires that pip be updated first, and there's only like 4 top-level packages being downloaded to begin with.

```yaml
- name: Get pip cache dir
  id: pip-cache
  run: |
    echo "::set-output name=dir::$(pip cache dir)"

- name: pip cache
  uses: actions/cache@v2
  with:
    path: ${{ steps.pip-cache.outputs.dir }}
    key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
    restore-keys: |
      ${{ runner.os }}-pip-
```

You could hardcode the path for every OS like this but I didn't really feel like it was worth the hassle for such a small number of packages.

```yaml
 - uses: actions/cache@v2
   if: startsWith(runner.os, 'Linux')
   with:
     path: ~/.cache/pip
     key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
     restore-keys: |
       ${{ runner.os }}-pip-

 - uses: actions/cache@v2
   if: startsWith(runner.os, 'macOS')
   with:
     path: ~/Library/Caches/pip
     key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
     restore-keys: |
       ${{ runner.os }}-pip-

 - uses: actions/cache@v2
   if: startsWith(runner.os, 'Windows')
   with:
     path: ~\AppData\Local\pip\Cache
     key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
     restore-keys: |
       ${{ runner.os }}-pip-
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,257 @@
...
+jobs:
+  main_tests:
+    name: Main tests
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: ['3.9']
+        # ['2.7', '3.6', '3.7', '3.8', '3.9', '3.10.0-rc.1', 'pypy-2.7', 'pypy-3.6', 'pypy-3.7-nightly']
+        include:
```
dralley, line 15: It might be better to just skip the matrix above and manually declare the permutations as done below?

.github/workflows/release-pypi.yml
```diff
@@ -0,0 +1,79 @@
...
+      with:
+        path: dist/*.tar.gz
+
+
+  build_bdist:
+    name: Build binary wheels
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        # Windows build fails
```
dralley, line 38: See: https://github.com/dralley/hpy/runs/3287054071?check_suite_focus=true#step:4:218

.github/workflows/release-pypi.yml
```diff
@@ -0,0 +1,79 @@
...
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
+
+    # for other architectures, see: https://cibuildwheel.readthedocs.io/en/stable/faq/#emulation
+    - name: Build wheels for CPython
+      uses: pypa/cibuildwheel@v2.1.1
+      env:
+        CIBW_SKIP: 'cp310-* pp*'  # 3.10 and pypy builds are broken
```
dralley, line 54: See for 3.10: https://github.com/dralley/hpy/runs/3286978327?check_suite_focus=true#step:3:1181

The pypy issue is the same as before.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,257 @@
...
+          - os: ubuntu-latest
+            python-version: '3.6'
+          - os: ubuntu-latest
+            python-version: '3.7'
+          - os: ubuntu-latest
+            python-version: '3.8'
+          ## Expected failure
+          # - os: ubuntu-latest
+          #   python-version: '3.10.0-rc.1'
```
dralley, line 24: See: https://github.com/dralley/hpy/pull/1/checks?check_run_id=3286253606

.github/workflows/ci.yml
```diff
@@ -0,0 +1,257 @@
...
+          ## Expected failure
+          # - os: ubuntu-latest
+          #   python-version: 'pypy-3.6'
```
dralley, line 27: See: https://github.com/dralley/hpy/runs/3295502351?check_suite_focus=true#step:4:797

.github/workflows/ci.yml
```diff
@@ -0,0 +1,257 @@
...
+            python-version: '3.6'
+          - os: ubuntu-latest
+            python-version: '3.7'
+          - os: ubuntu-latest
+            python-version: '3.8'
+          - os: ubuntu-latest
+            python-version: '3.10.0-rc.1'
+          ## Expected failure
+          # - os: ubuntu-latest
+          #   python-version: 'pypy-3.6'
```
dralley, line 106: Same problem as pypi before.

hendrikmuhs:
> I attempted to make artifact caching work but wasn't fully successful. Here's what I did for ccache, and you can see that it installed properly but for some reason the cache was not being used (no hits, but also no misses).

This sounds like your compiler isn't calling ccache. In addition to integrating the action you need to change your build steps and point it to call `ccache` instead of `/usr/bin/{c++, g++, clang, ...}`.

The easiest is `PATH` manipulation, e.g. like python:

https://github.com/python/cpython/blob/main/.github/workflows/build.yml#L166

This prefixes `PATH` with `/usr/lib/ccache/`, if you look there you find symbolic links `{c++, g++, clang, ...} -> ccache`

There are also other ways, e.g. you explicitly telling your build tool to use `ccache` as compiler, e.g. via the cmake action, the README in https://github.com/hendrikmuhs/ccache-action has an example for that.

hodgestar:
I've added a simple documentation build sanity check to the existing Azure CI pipeline in #237. It would be good if it can be included in the new GitHub Actions in this PR, but if that is difficult, we can always re-add it later.

dralley:
My personal preference would be to merge as is, and then add back caching and documentation in a later PR

dralley:
Done, I think.  It doesn't look like any of the documentation needs to change, just the additional CI job.

.github/workflows/ci.yml
antocuni: I don't have any preference. In theory, if we declare that we support a Python version the tests should pass on all the supported OSes so the matrix looks better.
Do you have any specific reason to prefer the explicit list?

.github/workflows/ci.yml
antocuni: this is a failure that we would also have on master, is it correct? I.e. it's an hpy problem, it's not related to the choice of azure vs github actions?

.github/workflows/ci.yml
antocuni: ideally it would be nice to have a way to reuse the very same combinations of python/os that we use for `main_tests` to ensure that we never forget to add supported versions also here.
But I don't know if it's technically possible: if not, repetition is fine but maybe we could add a comment above which says "if you add a new python version, please also update `poc_tests`", and maybe move the `poc_tests` block closer to `main_tests`.

.github/workflows/release-pypi.yml
```diff
@@ -0,0 +1,79 @@
+---
+name: Release to PyPI
+
+on:
+  release:
+    types: [created]
+
```
antocuni, line 7: I don't know much of github action. I assume this means that the workflow runs whenever we push a tag?

.github/workflows/release-pypi.yml
antocuni: I'm confused. Windows tests pass but windows build fails? That's curious, but not a merge blocker of course.

.github/workflows/ci.yml
antocuni: yes, we can't/shouldn't run tests on pypy.
PyPy releases its own bundled version of hpy and take care of running its own test suite, so here we just need to test cpython

.github/workflows/release-pypi.yml
antocuni: ... and the same answer an well 😅

.github/workflows/release-pypi.yml
```diff
@@ -0,0 +1,79 @@
...
+      env:
+        CIBW_SKIP: 'cp310-* pp*'  # 3.10 and pypy builds are broken
+        CIBW_ARCHS: 'auto64'  # only 64-bit
+        CIBW_TEST_REQUIRES: pytest pytest-xdist
+        CIBW_TEST_COMMAND: pytest -n auto {project}/test/
+
+    - uses: actions/upload-artifact@v2
+      with:
+        path: ./wheelhouse/*.whl
+
```
antocuni, line 62: If I understand correctly this makes the wheel available for download directly from github actions?
Nice! :)

.github/workflows/release-pypi.yml
```diff
@@ -0,0 +1,79 @@
...
+      - uses: actions/download-artifact@v2
+        with:
+          name: artifact
+          path: dist
+
+      - uses: pypa/gh-action-pypi-publish@v1.4.2
+        with:
+          user: __token__
+          password: ${{ secrets.pypi_api_token }}
+          # To test: repository_url: https://test.pypi.org/legacy/
```
antocuni, line 79: I don't understand this comment. Does it mean that currently the builds are uploaded to `test.pypi.org`?

.github/workflows/release-pypi.yml
```diff
@@ -0,0 +1,79 @@
...
+    steps:
+      - uses: actions/download-artifact@v2
+        with:
+          name: artifact
+          path: dist
+
+      - uses: pypa/gh-action-pypi-publish@v1.4.2
+        with:
+          user: __token__
+          password: ${{ secrets.pypi_api_token }}
```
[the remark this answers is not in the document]
antocuni: is it something which we should done after we merge the PR, or it might be useful to do it even before, for testing?

antocuni (approved):
sorry if it took so long to review but thank you a lot for doing it.
Overall the PR looks very good to me: I added some comments but most of them are just genuine questions more than requests of change :)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,279 @@
...
+        python-version: ['3.9']
+        include:
+          - os: ubuntu-latest
+            python-version: '3.6'
+          - os: ubuntu-latest
+            python-version: '3.7'
+          - os: ubuntu-latest
+            python-version: '3.8'
+          - os: ubuntu-latest
+            python-version: '3.10.0-rc.1'
```
antocuni, line 125: we should comment out this as it's failing. It seems that hpy itself is broken on py3.10

antocuni:
> Here's the building PyPI wheels job passing (except for the upload part for obvious reasons): https://github.com/dralley/hpy/pull/1/checks?check_run_id=3296113677#step:4:10

it seems the link is broken (it opens something but there is a banner saying that id 3296113677 is not found.
Maybe it's because github deletes old runs after a while?

> 
> I attempted to make artifact caching work but wasn't fully successful. Here's what I did for ccache, and you can see that it installed properly but for some reason the cache was not being used (no hits, but also no misses).
> 
> ```yaml
>       - name: ccache
>         if: ${{ !startsWith(runner.os, 'Windows') }}
>         uses: hendrikmuhs/ccache-action@v1
>         with:
>           key: ${{ matrix.os }}-${{ matrix.python-version }}
>           max-size: 100M
> ```
> 
> https://github.com/dralley/hpy/pull/1/checks?check_run_id=3296448518#step:21:10

the [README](https://github.com/hendrikmuhs/ccache-action) of ccache-action says that enabling the action is not enough and that you need to explicitly hack the build step to use ccache, e.g. by manipulating `$PATH`. Maybe it's enough to set the `CC` env variable in the build steps.


> And here's what I did for Python caching, but this strategy requires that pip be updated first, and there's only like 4 top-level packages being downloaded to begin with.
[cut]

I think it's fine to forget this is it's only 4 packages and it's hard/complicate to support.

.github/workflows/ci.yml
dralley: I believe so, yes.  Azure Pipelines isn't testing 3.10 RC.

I think I messed up this link though, it doesn't include a 3.10 runner, and I definitely meant to link to one.

.github/workflows/ci.yml
dralley: This one actually wasn't failing.   IIRC it was some kind of runtime failure rather than a build failure.  I'll try to hunt down a CI log that demonstrates it.

dralley:
This is kinda weird, all of my CI links got shuffled somehow, I was pretty sure they were correct at the time :)

Let me re-enable everything and give it another run

.github/workflows/release-pypi.yml
dralley: No - the default `repository_url` is the correct one, but you can override it with whichever value you like for testing purposes.  

I can remove that comment if it adds no value.

.github/workflows/release-pypi.yml
dralley: I think it does make the artifacts downloadable directly from github (I haven't tried it but that's my impression) but that's not the primary reason.

The main benefit is that you can put the sdist package build + test and the bdist package(s) build + test in separate parallel jobs, and only upload the artifacts (from a third job) once all of the jobs have finished.  But each job obviously has a fresh CI image so the artifacts need to be shared somehow. 

If you just uploaded the artifacts from each of the two jobs then you could create a situation where one job passes and starts uploading packages to PyPI before the other has completed all of its work successfully, and that wouldn't be good.


>This uploads artifacts from your workflow allowing you to share data between jobs and store data once a workflow is complete.

https://github.com/actions/upload-artifact

.github/workflows/ci.yml
dralley: Not especially.  The only reason would be that if the matrix is only going to be 1x3 then it doesn't serve very much purpose - but that's not a real *problem* per se.

.github/workflows/ci.yml
dralley: I'll drop it from the PR then.

.github/workflows/ci.yml
dralley: I don't think that's possible, at least not yet.

.github/workflows/release-pypi.yml
dralley: You can do that, but I think this method triggers when you create a release from the github UI.

https://docs.github.com/en/actions/reference/events-that-trigger-workflows#release

I think doing it on tag push would look like this

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags

.github/workflows/release-pypi.yml
antocuni, line 79:
```suggestion
          # uncomment the following to upload to test.pypi.org
          # repository_url: https://test.pypi.org/legacy/
```

antocuni: aaah ok, *now* I understand what is the point of the comment. I made a suggestion to make it clearer (and this is not a nitpick, I really didn't understand it at first 😅)

.github/workflows/ci.yml
antocuni: no, I think we want to support multiple Python versions so it will be 3xN at least.
+1 for the matrix

.github/workflows/ci.yml
antocuni: ok, then I vote for putting a comment as a reminder

.github/workflows/release-pypi.yml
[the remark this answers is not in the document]
dralley: It doesn't matter when you do it really, the job will just fail until it's added.  That's not even really a problem, you could do the next release in the traditional way, but let the job be triggered anyway - the build + test will happen normally but it won't upload any packages.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,276 @@
...
+        python-version: ['3.9']
+        include:
+          - os: ubuntu-latest
+            python-version: '3.6'
+          - os: ubuntu-latest
+            python-version: '3.7'
+          - os: ubuntu-latest
+            python-version: '3.8'
+          - os: ubuntu-latest
+            python-version: '3.10.0-rc.2'
```
dralley, line 70: Never mind this does fail at runtime https://github.com/dralley/hpy/pull/1/checks?check_run_id=3551010371#step:5:214

.github/workflows/ci.yml
```diff
@@ -0,0 +1,276 @@
...
+        include:
+          - os: ubuntu-latest
+            python-version: '3.6'
+          - os: ubuntu-latest
+            python-version: '3.7'
+          - os: ubuntu-latest
+            python-version: '3.8'
+          ## Expected failure
+          - os: ubuntu-latest
+            python-version: '3.10.0-rc.2'
```
dralley, line 25: Failing main tests: https://github.com/dralley/hpy/pull/1/checks?check_run_id=3551010532#step:6:42

.github/workflows/release-pypi.yml
dralley: It seems to have been resolved in the meantime?  But it looks like there is maybe another issue.

https://github.com/dralley/hpy/runs/3551010010?check_suite_focus=true#step:4:265

.github/workflows/ci.yml
antocuni: ok, then I propose to just skip 3.10 for now

.github/workflows/release-pypi.yml
antocuni: I'm fine to skip windows build and merge the PR ASAP. Then we can fix windows later

antocuni:
@dralley merged! Thank you very much for all the effort which you put into this, I think it's a great step forward for us.
Also, feel free to come to the IRC channel if you want to discuss things in a more real time fashion.

PR #234, opened 2021-09-02
Add HPyUnicode_AsUTF8AndSize, HPyUnicode_DecodeFSDefault and "s" support in HPyArg_Parse

steve-s:
APIs needed to port `psutils`

### DIFF
--- a/hpy/devel/src/runtime/argparse.c
+++ b/hpy/devel/src/runtime/argparse.c
@@ -54,6 +54,29 @@
  * ``d (float) [double]``
  *     Convert a Python floating point number to a C double.
  *
+ * Strings and buffers
+ * ~~~~~~~~~~~~~~~~~~~~~~~~
+ *
+ * These formats allow accessing an object as a contiguous chunk of memory.
+ * You don't have to provide raw storage for the returned unicode or bytes
+ * area.
+ *
+ * In general, when a format sets a pointer to a buffer, the pointer is valid
+ * only until the corresponding HPy handle is closed.
+ *
+ * ``s (unicode) [const char*]``
+ *
+ * Convert a Unicode object to a C pointer to a character string.
+ * A pointer to an existing string is stored in the character pointer
+ * variable whose address you pass.  The C string is NUL-terminated.
+ * The Python string must not contain embedded null code points; if it does,
+ * a `ValueError` exception is raised. Unicode objects are converted
+ * to C strings using 'utf-8' encoding. If this conversion fails,
+ * a `UnicodeError` is raised.
+ *
+ * Note: This format does not accept bytes-like objects and is therefore
+ * not suitable for filesystem paths.
+ *
  * Handles (Python Objects)
  * ~~~~~~~~~~~~~~~~~~~~~~~~
  *
@@ -77,11 +100,6 @@
  *     valid Python value. See
  *     `Truth Value Testing <https://docs.python.org/3/library/stdtypes.html#truth>`_
  *     for more information about how Python tests values for truth.
- *
- * ``s (unicode) [const char*]``
- *      Convert a Python unicode object to a C string. The lifetime of the C string
- *      is tied to the underlying Python object (TODO: should be tied to the handle
- *      instead?).
  *
  * Options
  * ~~~~~~~
@@ -350,6 +368,10 @@ parse_item(HPyContext *ctx, HPyTracker *ht, HPy current_arg, int current_arg_tmp
         }
         HPy_ssize_t size;
         const char *data = HPyUnicode_AsUTF8AndSize(ctx, current_arg, &size);
+        if (data == NULL) {
+            set_error(ctx, ctx->h_SystemError, err_fmt, "unicode conversion error");
+            return 0;
+        }
         // loop bounded by size is more robust/paranoid than strlen
         HPy_ssize_t i;
         for (i = 0; i < size; ++i) {
@@ -359,7 +381,7 @@ parse_item(HPyContext *ctx, HPyTracker *ht, HPy current_arg, int current_arg_tmp
             }
         }
         if (data[i] != '\0') {
-            set_error(ctx, ctx->h_ValueError, err_fmt, "missing terminating null character");
+            set_error(ctx, ctx->h_SystemError, err_fmt, "missing terminating null character");
             return 0;
         }
         *output = data;
--- a/hpy/devel/src/runtime/argparse.c
+++ b/hpy/devel/src/runtime/argparse.c
@@ -78,6 +78,11 @@
  *     `Truth Value Testing <https://docs.python.org/3/library/stdtypes.html#truth>`_
  *     for more information about how Python tests values for truth.
  *
+ * ``s (unicode) [const char*]``
+ *      Convert a Python unicode object to a C string. The lifetime of the C string
+ *      is tied to the underlying Python object (TODO: should be tied to the handle
+ *      instead?).
+ *
  * Options
  * ~~~~~~~
  *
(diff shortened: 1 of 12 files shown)


test/test_argparse.py
```diff
@@ -56,6 +56,20 @@ def test_b(self):
...
+        assert mod.f("hello HPy") == "hello HPy"
+        with pytest.raises(ValueError) as err:
+            mod.f(b"hello\0HPy".decode('utf-8'))
+        assert str(err.value) == (
+            "function embedded null character"
+        )
+        with pytest.raises(TypeError) as err:
+            mod.f(b"hello HPy")
+        # No assertion of the message for now
+        # CPython's message reads: argument {N} must be str, not int
```
steve-s, line 71: This is a question of how far we want to go in compatibility of `HPyArg_Parse` with `PyArg_ParseTuple`. To have the same message we'd need to change the current structure of `argparse.c` a bit.

hpy/devel/src/runtime/argparse.c
```diff
@@ -78,6 +78,11 @@
  *     `Truth Value Testing <https://docs.python.org/3/library/stdtypes.html#truth>`_
  *     for more information about how Python tests values for truth.
  *
+ * ``s (unicode) [const char*]``
+ *      Convert a Python unicode object to a C string. The lifetime of the C string
+ *      is tied to the underlying Python object (TODO: should be tied to the handle
```
fangerer, line 83: IMO, the lifetime of the pointer should be tied to the handle.
The alternative is that it's a fresh "copy" and the caller then owns the pointer and must dealloc itself.
@antocuni @hodgestar @rlamy  What do you think?

fangerer (commented):
Looks good to me but let's wait for a response concerning the lifetime of the `char *`.

test/test_argparse.py
fangerer: I don't think that we need to have the same error message. It's important to have the same error type.

hpy/devel/src/runtime/argparse.c
hodgestar: @fangerer Agreed, with the caveat that I have no idea how hard those will be to implement.

hpy/devel/src/runtime/argparse.c
```diff
@@ -78,6 +78,11 @@
  *     `Truth Value Testing <https://docs.python.org/3/library/stdtypes.html#truth>`_
  *     for more information about how Python tests values for truth.
  *
+ * ``s (unicode) [const char*]``
```
hodgestar, line 81: I guess `s` should go in its own new section, not under `Miscellaneous`. The CPython documentation uses the heading `Strings and buffers` which probably makes sense since they share many common considerations (e.g. the lifetime and deallocation of the returned memory).

hodgestar: We should also mention precisely what the returned buffer contains (i.e. UTF-8 bytes with a NUL at the end). I would recommend actually copying the CPython documentation for `s` and then adjusting it as necessary.

hpy/devel/src/runtime/argparse.c
fangerer: I would not _enforce_ it for the sake of performance on CPython. For GraalPython (and maybe also PyPy) it's easier to attach the pointer to the handle and also free it when the handle is closed.
For CPython, we can still always return the same pointer. It's just important that the user does not rely on the pointer being valid after the handle was closed.

The second suggestion (i.e. the caller gets a fresh pointer and owns it) is probably a bad idea because that means we need allocate a fresh string and copy contents every time (even in CPython).

test/test_argparse.py
antocuni: I agree that the exact message is not important. Pinging @hodgestar since I think he's the author of the comment

hpy/devel/src/runtime/argparse.c
steve-s: I updated the documentation to

> the pointer is valid only until the corresponding HPy handle is closed

I think that it's fine if in the non-debug mode the pointer stays valid. No one should rely on that and ideally the debug mode should catch any errors. I can take a look at enforcing that in debug mode as one of the next things. For now I think the specification is enough.

hpy/devel/src/runtime/argparse.c
steve-s: Fixed

hpy/devel/src/runtime/argparse.c
```diff
@@ -337,11 +342,37 @@ parse_item(HPyContext *ctx, HPyTracker *ht, HPy current_arg, int current_arg_tmp
...
-    default:
+    case 's': {
+        const char **output = va_arg(*vl, const char **);
+        if (!HPyUnicode_Check(ctx, current_arg)) {
+            set_error(ctx, ctx->h_TypeError, err_fmt, "a str is required");
+            return 0;
+        }
+        HPy_ssize_t size;
+        const char *data = HPyUnicode_AsUTF8AndSize(ctx, current_arg, &size);
+        // loop bounded by size is more robust/paranoid than strlen
```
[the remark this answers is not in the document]
steve-s: Fixed

test/test_argparse.py
steve-s: I am author of the comment :-) I think it may be useful to keep it just so we know. I intentionally did not make it a "TODO" comment, because I also think it's fine as it is.

hpy/devel/src/runtime/argparse.c
steve-s: Moreover, I think this could be turned into more generic rule that the lifetime of any pointers returned by HPy API is tied to the HPy handle that "owns" them. 

It will be more fun with mutable data if there will be any such API. There it is also a question of when any changes to the data should be visible to other handles referencing the same object.

fangerer: We already had a similar discussion a while ago when we were discussing the semantics of `HPy_AsStruct` (formerly `HPy_Cast`).
I was bringing up that if we would (eventually) have multi-threading then we would need to define this properly. In CPython, it's just shared memory.

test/test_argparse.py
hodgestar: I don't think we need to have the same error message as PyArg_ParseTuple but I think we should check the error message here because this error message is generated by HPy code (i.e. our code) and we should check that the code is doing the right thing (i.e. generating a sensible error message with everything in the right place, etc).

antocuni (approved):
LGTM, thank you for doing it.

Ideally, I would like to add some logic in the debug mode to copy&free the buffer returned by `HPyUnicode_AsUTF8AndSize`, but it's not necessary to do it in this PR.
I'll be happy to give some guidance on how to extend the debug mode if necessary.

hpy/devel/src/runtime/argparse.c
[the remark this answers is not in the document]
steve-s: For this the Python implementation would have to be faulty and return wrong `size`. That's why I also changed the error to `SystemError` (was `ValueError`).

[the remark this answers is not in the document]
steve-s: You meant the `HPyUnicode_AsUTF8AndSize` returns `NULL` situation. Hmm, I am not sure how to construct such string without some hacks via the C API. On the Python side it would not allow me to create a string with invalid bytes sequence.

hpy/devel/src/runtime/argparse.c
```diff
@@ -337,11 +360,41 @@ parse_item(HPyContext *ctx, HPyTracker *ht, HPy current_arg, int current_arg_tmp
...
+        for (i = 0; i < size; ++i) {
+            if (data[i] == '\0') {
+                set_error(ctx, ctx->h_ValueError, err_fmt, "embedded null character");
+                return 0;
+            }
+        }
+        if (data[i] != '\0') {
+            set_error(ctx, ctx->h_SystemError, err_fmt, "missing terminating null character");
+            return 0;
+        }
```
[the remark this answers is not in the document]
steve-s: Yeah so here applies this comment:

> For this the Python implementation would have to be faulty and return wrong size. That's why I also changed the error to SystemError (was ValueError).

hpy/devel/src/runtime/argparse.c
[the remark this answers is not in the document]
steve-s: It being not possible would also explain why CPython uses `SystemError` and not `ValueError`

hpy/devel/src/runtime/argparse.c
steve-s: I've created https://github.com/hpyproject/hpy/issues/236 for the debug mode enhancement.

test/test_argparse.py
steve-s: Fixed in the last commit

hpy/devel/src/runtime/argparse.c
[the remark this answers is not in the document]
antocuni: ah indeed, I think you are right. Nevermind then

hpy/devel/src/runtime/argparse.c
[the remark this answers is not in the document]
antocuni: yes, you are right. Sorry for the noise

PR #235, opened 2021-09-02
Add HPyErr_SetFromErrno

steve-s:
This is necessary for porting `psutils`

### DIFF
--- a/test/test_hpyerr.py
+++ b/test/test_hpyerr.py
@@ -139,23 +139,23 @@ class TestErr(HPyTest):
         import pytest
         import errno
         mod = self.make_module("""
-            #include <stdio.h>
+            #include <errno.h>
 
             HPyDef_METH(f, "f", f_impl, HPyFunc_O)
             static HPy f_impl(HPyContext *ctx, HPy self, HPy type)
-            {
-                fseek(stdin, -200, -10);
+            {{
+                errno = {errno};
                 HPyErr_SetFromErrno(ctx, type);
                 return HPy_NULL;
-            }
+            }}
             @EXPORT(f)
             @INIT
-        """)
-        for type in [OSError, IOError]:
+        """.format(errno = errno.EINVAL))
+        for type in [OSError, TimeoutError]:
             with pytest.raises(type) as err:
                 mod.f(type)
 
-            assert err.value.errno in [errno.EINVAL, errno.ESPIPE]
+            assert err.value.errno == errno.EINVAL
 
     def test_h_exceptions(self):
         import pytest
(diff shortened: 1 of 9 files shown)


test/test_hpyerr.py
```diff
@@ -135,6 +135,28 @@ def test_HPyErr_SetObject(self):
...
+            static HPy f_impl(HPyContext *ctx, HPy self, HPy type)
+            {
+                fseek(stdin, -200, -10);
+                HPyErr_SetFromErrno(ctx, type);
+                return HPy_NULL;
+            }
+            @EXPORT(f)
+            @INIT
+        """)
+        for type in [OSError, IOError]:
```
antocuni, line 154: on Python3, `IOError` is just an alias to `OSError`. Why do you need to test both here?

test/test_hpyerr.py
```diff
@@ -135,6 +135,28 @@ def test_HPyErr_SetObject(self):
...
+            with pytest.raises(type) as err:
+                mod.f(type)
+
+            assert err.value.errno in [errno.EINVAL, errno.ESPIPE]
```
antocuni, line 158: I suppose this is needed because `fseek` could return either error, but I had to read its manpage to understand that. Maybe it's worth a short comment to explain

antocuni (changes requested):
LGTM in general but windows tests are failing

test/test_hpyerr.py
steve-s: Yes, good point I'll add some explanatory comment. Maybe I'll have to change this altogether, because it does not work on Windows it seems at least from the azure pipelines.

test/test_hpyerr.py
steve-s: Ah, good point. I wanted to test it with at least two exceptions, so I searched how the API is used in packages and saw two two types being used. I'll find some other suitable type

steve-s:
I forgot that `PyErr_SetFromErrno` also conveniently returns `NULL`. I think it would be nice to have that also in HPy. I'll wait for the CI jobs to finish and then I'll push one more commit with this change:
```
--- a/hpy/tools/autogen/public_api.h
+++ b/hpy/tools/autogen/public_api.h
@@ -179,7 +179,7 @@ HPy HPy_CallTupleDict(HPyContext *ctx, HPy callable, HPy args, HPy kw);
 void HPy_FatalError(HPyContext *ctx, const char *message);
 void HPyErr_SetString(HPyContext *ctx, HPy h_type, const char *message);
 void HPyErr_SetObject(HPyContext *ctx, HPy h_type, HPy h_value);
-void HPyErr_SetFromErrno(HPyContext *ctx, HPy h_type);
+HPy HPyErr_SetFromErrno(HPyContext *ctx, HPy h_type);
```

antocuni:
ok, waiting for your last commit before merging then

steve-s:
> ok, waiting for your last commit before merging then

done (also rebased on master)

antocuni (approved):

PR #237, opened 2021-09-07
Documentation build cleanup

hodgestar:
Our documentation currently isn't building on readthedocs and it's producing quite a few warnings. This PR intends to clean all that up.

Issues addressed so far:

- [x] Add double underscores to anonymous external references to prevent RST warnings.
- [x] Reference the non-existent debug mode documentation consistently throughout the docs.
- [x] Give runtime and helper API documentation different section names to prevent a Sphinx warning.
- [x] Update removal of HPyAPI_HELPER define to help spinx-c-autodoc / clang properly find these functions.
- [x] Remove attempt to call cindex.Index.create that segfaults in some cases.

Still to do:

- [x] Debug the failing [readthedocs builds](https://readthedocs.org/projects/hpy/builds/) (hopefully easier now that less is going wrong)
- [x] Discuss the complete removal of the call to `cindex.Index.create`
- [x] Discuss making the build fail in the are warnings produced while running the documentation build (this would have picked up some of the issues that resulted in our build missing documentation or failing)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -179,6 +179,10 @@ jobs:
       make valgrind
 
 - job: BuildDocs
+  # This job checks that the documentation builds successfully without
+  # warnings but does not store the built documentation. The canonical
+  # copy of the documentation is built by readthedocs and viewable at
+  # https://docs.hpyproject.org/.
   pool:
     vmImage: 'ubuntu-latest'
   displayName: 'Build documentation'
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -177,3 +177,17 @@ jobs:
   - script: |
       pip install pytest pytest-azurepipelines pytest-valgrind
       make valgrind
+
+- job: BuildDocs
+  pool:
+    vmImage: 'ubuntu-latest'
+  displayName: 'Build documentation'
+  steps:
+  - template: azure-templates/python.yml
+  - script: python -m pip install --upgrade pip
+    displayName: 'Upgrade pip'
+  - script: python -m pip install -r docs/requirements.txt
+    displayName: 'Install documentation requirements'
+  - script: |
+      cd docs;
+      python -m sphinx -T -W -E -b html -d _build/doctrees -D language=en . _build/html
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -184,6 +184,8 @@ jobs:
   displayName: 'Build documentation'
   steps:
   - template: azure-templates/python.yml
+  - script: sudo apt update && sudo apt install -y libclang-10-dev
+    displayName: 'Install libclang-10-dev'
   - script: python -m pip install --upgrade pip
     displayName: 'Upgrade pip'
   - script: python -m pip install -r docs/requirements.txt
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -14,7 +14,6 @@
 # import sys
 # sys.path.insert(0, os.path.abspath('.'))
 
-import sys
 import os
 import re
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -68,10 +68,25 @@ def pre_process(app, filename, contents, *args):
 def setup(app):
     app.connect("c-autodoc-pre-process", pre_process)
 
+
 def setup_clang():
     """
-    The clang library needs to find libclang*.so, but it seems there isn't a
-    fully portable solution. Let's try our best.
+    Make sure clang is set up correctly for the sphinx_c_autodoc extension.
+
+    The Python clang package requires a matching libclang*.so. Our
+    ``doc/requirements.txt`` file specifies ``clang==10.0.1``, so we need
+    ``libclang-10``.
+
+    On Ubuntu 20.04 (and possibly later), this can be installed with
+    ``apt install libclang-10-dev`` and the Python clang package finds the
+    appropriate .so automatically.
+
+    However, ReadTheDocs has an older Ubuntu that only packages libclang-6.0.
+    The Python ``clang==10.0.1`` packages supports this older .so, but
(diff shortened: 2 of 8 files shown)


docs/conf.py
```diff
@@ -84,7 +84,7 @@ def setup_clang():
         return
 
     try:
-        cindex.Index.create()
+        pass  # cindex.Index.create()
```
hodgestar, line 87: @antocuni Are you okay with me just deleting this call `cindex.Index.create()` now? It segfaults consistently on my machine. Not sure if your setup still needs it or not though.

fangerer (approved):
thanks !

docs/conf.py
antocuni: it segfaults also for me, I'm fine with deleting.
But maybe keep a comment saying that on ubuntu you need `apt install libclang-10-dev`?

hodgestar:
It looks like sorting out the other issues (probably the segfault on Ubuntu) fixed the readthedocs builds. Woot.

hodgestar:
@fangerer @antocuni (or anyone else): Ready for a proper review.

I added some very basic documentation of the debug mode to make the documentation build pass without warnings.

azure-pipelines.yml
```diff
@@ -177,3 +177,19 @@ jobs:
...
+  - template: azure-templates/python.yml
+  - script: sudo apt update && sudo apt install -y libclang-10-dev
+    displayName: 'Install libclang-10-dev'
+  - script: python -m pip install --upgrade pip
+    displayName: 'Upgrade pip'
+  - script: python -m pip install -r docs/requirements.txt
+    displayName: 'Install documentation requirements'
+  - script: |
+      cd docs;
+      python -m sphinx -T -W -E -b html -d _build/doctrees -D language=en . _build/html
```
antocuni, line 195: if I understand it correctly, the point of this job is just to check that the docs build, and then we discard the result? If so, maybe add a comment to explain that.
Also, note that as soon as we merge #230 we probably we need re-add this job to the CI

docs/debug-mode.rst
```diff
@@ -0,0 +1,17 @@
+Debug Mode
+==========
+
+HPy includes a debug mode that may be activated at *run time* with
+*no overhead*. This is possible because the whole of the HPy API is provided
```
antocuni, line 5:
```suggestion
HPy includes a debug mode which includes a lot of useful run-time checks to
ensure that C extensions use the API correctly. The major points of the debug mode are:

    1. no special compilation flags are required: it is enough to compile the extension 
       with the Universal ABI.
    
    2. The debug mode can be activated at *import time*, and it can be activated
       per-extension.
    
    3. You pay the overhead of the debug mode only if you use it. Extensions loaded 
       without the debug mode run at full speed.

This is possible because the whole of the HPy API is provided
```

docs/misc/str-builder-api.rst
```diff
@@ -227,12 +227,12 @@ widely used outside CPython. A simple ``grep`` found only 17 matches in the
 4000 packages, although some are in very important packages such as
 `cffi <https://github.com/hpyproject/top4000-pypi-packages/blob/0cd919943a007f95f4bf8510e667cfff5bd059fc/top100/0021-cffi-1.14.5/c/wchar_helper_3.h#L36>`_,
 ``markupsafe``
-(`1 <https://github.com/hpyproject/top4000-pypi-packages/blob/0cd919943a007f95f4bf8510e667cfff5bd059fc/top100/0024-MarkupSafe-2.0.1/src/markupsafe/_speedups.c#L106>`_,
-`2 <https://github.com/hpyproject/top4000-pypi-packages/blob/0cd919943a007f95f4bf8510e667cfff5bd059fc/top100/0024-MarkupSafe-2.0.1/src/markupsafe/_speedups.c#L132>`_,
-`3 <https://github.com/hpyproject/top4000-pypi-packages/blob/0cd919943a007f95f4bf8510e667cfff5bd059fc/top100/0024-MarkupSafe-2.0.1/src/markupsafe/_speedups.c#L158>`_)
+(`1 <https://github.com/hpyproject/top4000-pypi-packages/blob/0cd919943a007f95f4bf8510e667cfff5bd059fc/top100/0024-MarkupSafe-2.0.1/src/markupsafe/_speedups.c#L106>`__,
```
antocuni, line 230: just for curiosity, what is the difference between a single and a double `__`?

antocuni (approved):
LGTM, thanks!

antocuni:
> See, for example, https://docs.hpyproject.org/en/feature-doc-build-cleanup/api-reference/index.html

it's a 404. Maybe we need to enable some readthedocs setting to build the docs also for PRs?

docs/misc/str-builder-api.rst
hodgestar: A single `_` automatically creates a reference with the name of the link, which is a problem if multiple links have the same text. The double `__` prevents the reference being created.

azure-pipelines.yml
hodgestar: Correct. I will add a comment. When we add it back to GitHub actions it might be nice so that one can download the docs from the build for debugging purposes.

I already noted the existence of this PR in #230.

hodgestar:
> it's a 404. Maybe we need to enable some readthedocs setting to build the docs also for PRs?

Meh. I deactivated the automated readthedocs builds for this branch once I knew they were working. I didn't realize that would also make the URL not work.

azure-pipelines.yml
hodgestar: Comment added to explain things.

hodgestar:
Just waiting for the CI pipelines to finish and then I'll merge in the morning. Merging should also update the readthedocs documentation (which is a bit behind because the builds were failing there for awhile).

PR #239, opened 2021-09-13
Introduce helper HPy_BuildValue

steve-s:
In order to make porting to HPy more convenient I believe we need to provide API like `Py_BuildValue`, because otherwise there is too much error checking and handle closing boilerplate code. An alternative or additional API could be to extend the tuple builder with functions like `TupleBuilder_Set_i(ctx, builder, index, int_value)`. We can also have both and have `HPy_BuildValue` be built upon that API.

~This PR is intended as a discussion starter. It provides skeleton of the `HPy_BuildValue` implementation as I see it. Without the `TupleBuilder_Set_i` part for now.~

Update: the PR is now good to merge from my perspective. It provides all the formatting units used in `psutils` and I think it's now easy to add new less frequently used units.

### DIFF
--- a/hpy/devel/src/runtime/buildvalue.c
+++ b/hpy/devel/src/runtime/buildvalue.c
@@ -2,7 +2,8 @@
  * Implementation of HPy_BuildValue.
  *
  * HPy_BuildValue creates a new value based on a format string from the values
- * passed in an array. Returns HPy_NULL in case of an error and raises an exception.
+ * passed in variadic arguments. Returns HPy_NULL in case of an error and raises
+ * an exception.
  *
  * HPy_BuildValue does not always build a tuple. It builds a tuple only if its format
  * string contains two or more format units. If the format string is empty, it returns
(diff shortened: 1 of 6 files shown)


test/test_hpybuildvalue.py
```diff
@@ -0,0 +1,51 @@
...
+This means that global imports will NOT be visible inside the test
+functions. In particular, you have to "import pytest" inside the test in order
+to be able to use e.g. pytest.raises (which on PyPy will be implemented by a
+"fake pytest module")
+"""
+from .support import HPyTest
+from .test_hpylong import unsigned_long_bits
+
+
+class TestParseItem(HPyTest):
```
steve-s, line 13: Wrong name, I'll fix that

hpy/devel/src/runtime/buildvalue.c
```diff
@@ -0,0 +1,145 @@
+/**
+ * Implementation of HPy_BuildValue.
+ *
+ * HPy_BuildValue creates a new value based on a format string from the values
+ * passed in an array. Returns HPy_NULL in case of an error and raises an exception.
```
fangerer, line 5: `passed in an array` is out-dated. I think it should be `passed as varargs`.

hpy/devel/src/runtime/buildvalue.c
```diff
@@ -0,0 +1,145 @@
...
+    if (size < 0) {
+        return HPy_NULL;
+    } else if (size == 1) {
+        return build_single(ctx, &fmt, &values);
+    } else {
+        return build_tuple(ctx, &fmt, &values, size, '\0');
+    }
+}
+
+static HPy_ssize_t count_items(HPyContext *ctx, const char *fmt, char end)
```
fangerer, line 63: Since we are doing a pre-parsing of the format string here anyway:
We could add a hook to `HPy_BuildValue` that calls a context function instead of doing the parsing in C.
So, if the hook function is available, the `count_items` could then just extract the varargs into an array.
This sounds a bit awkward but in this way, we could avoid multiple upcalls while varargs unmarshalling is still done in the C extension's binary.
For CPython, the context function would just be `NULL`.
What do you think?

hpy/devel/src/runtime/buildvalue.c
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        case 'i':
+            return HPyLong_FromLong(ctx, (long)va_arg(*values, int));
+
+        case 'f':
+            return HPyFloat_FromDouble(ctx, (double)va_arg(*values, double));
+
+        default: {
+            HPyErr_SetString(ctx, ctx->h_SystemError, "NULL object passed to Py_BuildValue");
```
fangerer, line 118: I think the error message should be `"bad format char passed to HPy_BuildValue"`

fangerer (changes requested):

hpy/devel/src/runtime/buildvalue.c
steve-s: I am not sure if the C specification guarantees us that something like raw memory copy of a `double` to `uint64_t` will work as expected and you'd see the right bits (in the right order) on the other side. It would probably work, but not sure if we want to rely on that.

What I think the spec guarantees is casting `double*` to `void*` and then back, but it seems that one cannot just take address of `va_arg`, so we'd have to stack allocate array for each primitive type, store all the values there and then take pointers to them to create the final `void*` array. 

Alternatively, we can just pass the individual primitive arrays to the context function, so its signature would be something like `ctx_BuildValue(const char* fmt, int[] items_i, long[] items_l, double[] items_d, HPy[] items_hpy)`. The set of supported primitive values is part of the contract in any case regardless of whether they are in the signature. This would seem like reasonable approach to me.

Having the ability to handle this sometimes complex operation efficiently in the runtime without calling back and forth between HPy and the runtime seems like useful thing for GraalPython. What about PyPy? OTOH, we could say the same about `HPyArg_Parse`, so how far do we want to go? I would be for this simple `HPy_BuildValue` for now and eventually adding the possibility to provide optimized version by the context in the future.

fangerer: Passing individual arrays sounds good to me.
I think the _hook approach_ would also be reasonable for `HPyArg_Parse`.
I'm fine if we add the hook in a later stage.

steve-s:
I believe I hopefully addressed all @antocuni comments. Unless there's anything else to fix, I'd say let's merge this next week.

steve-s:
I've rebased the PR on master to also trigger the tests once more and if they pass I'll merge this.

PR #242, opened 2021-09-23
Always define HPY macro when building HPy extensions

steve-s:
This macro would, for example, allow to share code between HPy extensions and CPython extensions that have not been ported to HPy yet. In general I think it may be useful to detect if the code is being compiled for HPy or not.

### DIFF
--- a/hpy/devel/__init__.py
+++ b/hpy/devel/__init__.py
@@ -217,6 +217,7 @@ class build_hpy_ext_mixin:
         ext.hpy_abi = self.distribution.hpy_abi
         ext.include_dirs += self.hpydevel.get_extra_include_dirs()
         ext.sources += self.hpydevel.get_extra_sources()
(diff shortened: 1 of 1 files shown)


hodgestar (approved):
This looks like a good idea to me. I wouldn't mind some others chiming in on the exact name of the macro (e.g. `HPY` is short and clear, but would it clash with anything else, or not really?)

steve-s:
> This looks like a good idea to me. I wouldn't mind some others chiming in on the exact name of the macro (e.g. `HPY` is short and clear, but would it clash with anything else, or not really?)

Yeah I wasn't sure about that either. More opinions would be useful.

thedrow:
I don't think HPY would clash with anything and I think it's safe to merge this.
If I'm wrong, we can change it until we reach 1.0.

thedrow (approved):

PR #244, opened 2021-09-23
Add HPyErr_SetFromErrnoWithFilenameObjects and HPyErr_SetFromErrnoWithFilename

steve-s:
Recently I've added `HPyErr_SetFromErrno`, but now I think should have done it this way: 

* `HPyErr_SetFromErrnoWithFilenameObjects` is context function
* `HPyErr_SetFromErrnoWithFilenameObject` and `HPyErr_SetFromErrno` are helpers defined in HPy that just delegate to it passing `HPy_NULL` as default value for the file arguments

I also added `HPyErr_SetFromErrnoWithFilename`. This one could also be just a helper function, but that would also mean one additional upcall to FS decode, so I am inclined to keeping it a context function.


steve-s:
Ah I forgot the tests, will be added soon

antocuni:
Look generally good, but I'm not 100% that they should go to "runtime helpers".
The alternative is to define them directly as `static inline` functions in some `.h` file (e.g. `hpy/devel/include/macros.h`)

PR #251, opened 2021-10-19
Extend HPy call API

fangerer:
As discussed in #122  in this month's dev call, this PR adds APIs for calling functions and methods.
We don't want to add all the specific API functions like in the C API but we want to provide functions for two general use cases:

* Call some callable or method by using some existing arguments tuple or keywords dict (probably created in Python land)
* Call some callable or method by crafting the arguments in C.

Hence, this PR adds following:

* `HPy_CallMethodTupleDict`/ `HPy_CallMethodTupleDict_s`:
    Call a method using an argument tuple and a keyword dict.
    The name is given as Python unicode object (first variant) or as C string (suffix `_s`).
* `HPy_CallVectorDict`:
    Call a callable object using a handle array and a keyword dict.
* `HPy_CallMethodVectorDict`:
    Call a method using a handle array and a keyword dict.

I'm still not sure about the best way how to specify the keywords.
CPython 3.9 basically provides both variants: providing them as Python dict or providing an array of C strings (and the values are appended to the arguments array).
For now, I've implemented the dict variant because I think it's not very common to craft and pass keyword args in C.
According to an analysis of the top4000, the most frequently used functions are `PyObject_CallMethod` and `PyObject_CallFunction` and none of those allows to pass keyword arguments.


hodgestar:
The previous discussion of some of the options: https://github.com/hpyproject/hpy/issues/122

fangerer:
@hodgestar @rlamy @antocuni @mattip This should be ready for a second review round (actually, the first serious one).

hodgestar:
@fangerer `PyObject_Vectorcall` is unfortunately not a simple call in CPython -- it's a calling interface to an underlying PyObject protocol https://docs.python.org/3/c-api/call.html#the-vectorcall-protocol.

For example, `nargsf` is not the number of arguments, but rather the number of arguments XORed with a flag and if the flag is set then `args` is not just the pointer to the start of an array of arguments, but a pointer to the second argument of an array instead (but only in the case of `PyObject_Vectorcall` and not `PyObject_VectorcallMethod`.

`PyObject_VectorcallDict` is not fast, and so neither `HPy_CallVectorDict` or `HPy_CallVectorMethod` can be fast either. The "fast" calls are `PyObject_Vectorcall` and `PyObject_VectorcallMethod` which can call the underlying C function directly without having to repack arguments.

If we plan to implement the vectorcall protocol, we should probably add it to objects too and match the semantics and name the methods to match (e.g. `HPyVectorcall_XXX`).

If we don't plan to implement the vectorcall protocol, we should probably not use `vector` in our API function names and should probably choose more sensible semantics.

Hopefully this doesn't mess with your plans too much -- I discovered all of this while writing up #122 ages ago, but I don't think it made it into the issue write up very clearly. :/

fangerer:
@hodgestar Thanks for your comments.

> `PyObject_Vectorcall` is unfortunately not a simple call in CPython

Sure, but it's the API with the lowest overhead *if* `tp_vectorcall` is provided (by the user; not the generic one), right?
That's why I used it to implemented `ctx_Call*VectorDict`. We can improve that by testing if `PyType_HasFeature(type, Py_TPFLAGS_HAVE_VECTORCALL)` and then decide how to call it.

> For example, nargsf is not the number of arguments, but rather the number of arguments XORed with a flag and if the flag is set then args is not just the pointer to the start of an array of arguments, but a pointer to the second argument of an array instead (but only in the case of PyObject_Vectorcall and not PyObject_VectorcallMethod.

I'm aware of that and I'm sure that we agreed on not not doing the same.

> `PyObject_VectorcallDict` is not fast

It's still fast if no keyword arguments were specified because it then doesn't need to unpack anything. Sorry, I forgot to add some comments about this. According to the analysis of the top4000 packages (https://paste.opendev.org/show/bPLIEafAmoHq5prb9F8f/), the most frequently used call functions are `PyObject_CallMethod` and `PyObject_CallFunction` and neither of them allows to pass keyword args. So, my assumption basically was that keyword args are rarely provided.
Anyway, that's something I wanted to discuss: should we use a signature similar to `PyObject_Vectorcall` (i.e. having a `const char *kwnames` instead of the dict) or should we pass keywords as dict?

> If we plan to implement the vectorcall protocol, we should probably add it to objects too and match the semantics and name the methods to match (e.g. HPyVectorcall_XXX).
If we don't plan to implement the vectorcall protocol, we should probably not use vector in our API function names and should probably choose more sensible semantics.

Good question. I'm not sure if it makes sense to implement the vectorcall protocol. CPython also uses it internally but PyPy and Graalpython don't. Furthermore, our `HPyFunc_KEYWORDS` signature is already compatible to `HPy_CallMethodVectorDict`. So, I thought this would be a good fit. I suggest to remove the `Vector` from the names.

> Hopefully this doesn't mess with your plans too much -- I discovered all of this while writing up #122 ages ago, but I don't think it made it into the issue write up very clearly. :/

Don't worry, I just try to make some progress. There is no need to rush such things.

hpy/devel/src/runtime/ctx_call.c
```diff
@@ -36,3 +39,106 @@ ctx_CallTupleDict(HPyContext *ctx, HPy callable, HPy args, HPy kw)
...
+        /* If we have keywords, we use PyObject_VectorcallDict and lookup the method manually */
+        HPy callable = HPy_GetAttr(ctx, receiver, name);
+        if (HPy_IsNull(callable)) {
+            return HPy_NULL;
+        }
+        py_args = (PyObject **) alloca(nargs * sizeof(PyObject *));
+        _harr2pyarr(py_args, 0, args, nargs);
+        result = PyObject_VectorcallDict(_h2py(callable), py_args, nargs, _h2py(kw));
+        HPy_Close(ctx, callable);
+    }
```
antocuni, line 114: doing the manual lookup looks a bit weird. I think we are forced to do so because `PyObject_VectorcallMethod` uses a different convention for keyword arguments, is it correct?
If so, this might be a signal that our original choice of passing a `HPy dict` for keywords is not the best and that maybe we should follow more closely what vectorcall does?

test/test_call.py
```diff
@@ -88,6 +105,258 @@ def g():
...
+                return "key=" + str(key);
+
+        test_args = (
+            # (receiver, method, args_tuple)
+            (dict, (dict(a=0, b=1), ), {}),
+            (dict, tuple(), dict(a=0, b=1)),
+        )
+        for receiver, args_tuple, kw in test_args:
+            assert mod.call(receiver, *args_tuple, kw=kw) == receiver(*args_tuple, **kw)
+
```
antocuni, line 230: maybe it's just me, but I find this style of testing very difficult to read. To start with the comment `# (receiver, method, args_tuple)` looks wrong.
I think the following is orders or magnitude easier to read.
```python
assert mod.call(dict, dict(a=0, b=1), kw={}) == dict((dict(a=0, b=1)) # or even == {'a': 0, 'b': 1}
assert mod.call(dict, kw=dict(a=0, b=1)) == dict(a=0, b=1)
```
If the code above is not equivalent to the original code, it's a good sign that the original code is too complex to read 😅.

Feel free to ignore this suggestion though.

antocuni (changes requested):

antocuni:
I left some comments inline in the review. However, this probably requires a much deeper discussion.
@hodgestar already mentioned the discussion in #122 , but we already had more discussions in #147 , in particular https://github.com/hpyproject/hpy/pull/147#issuecomment-752040156.
Probably it would be best to keep the discussions about API design in the proper issue and keep the PRs for discussing only implementation details. Let's continue the discussion on #122.

fangerer:
@wjakob thanks for you comment and I agree. 

Since this PR is already a bit older, I'm not sure if I remember all the details correctly. The vectorcall protocol as such makes, for sure, sense in CPython. My question, if it makes sense, was because in HPy we would have the possibility to define a different protocol. But then we again need to do (costly) conversions in some cases which we should avoid.

> looks like HPy at this point provides most of the facilities that would be needed to create an extra backend for [nanobind](https://github.com/wjakob/nanobind)

I, of course, know nanobind and would love to see an HPy backend for it.

I will revisit this PR in the near future but this PR alone won't be enough (since it's just about an object calling API like https://docs.python.org/3/c-api/call.html#object-calling-api). We will also need to have the ability to define types that implement the vectorcall protocol (i.e. specify `tp_vectorcall_offset`).

mattip:
This has conflicts

hpy/devel/include/hpy/cpython/misc.h
```diff
@@ -276,6 +276,32 @@ HPyAPI_FUNC HPy HPy_CallTupleDict(HPyContext *ctx, HPy callable, HPy args, HPy k
     return ctx_CallTupleDict(ctx, callable, args, kw);
 }
 
+#if PY_VERSION_HEX < 0x03090000
+#define PyObject_Vectorcall _PyObject_Vectorcall
```
antocuni, line 280:
```suggestion
#    define PyObject_Vectorcall _PyObject_Vectorcall
```

hpy/devel/include/hpy/cpython/misc.h
```diff
@@ -276,6 +276,32 @@ HPyAPI_FUNC HPy HPy_CallTupleDict(HPyContext *ctx, HPy callable, HPy args, HPy k
...
+HPyAPI_FUNC HPy HPy_Call(HPyContext *ctx, HPy callable, const HPy *args, size_t nargs, HPy kwnames)
+{
+    if (sizeof(HPy) == sizeof(PyObject *)) {
+        return _py2h(PyObject_Vectorcall(_h2py(callable), (PyObject *const *)args, nargs, _h2py(kwnames)));
+    }
+    return ctx_Call(ctx, callable, args, nargs, kwnames);
+}
+
+#if PY_VERSION_HEX < 0x03090000
+#undef PyObject_Vectorcall
```
antocuni, line 292:
```suggestion
#    undef PyObject_Vectorcall
```

hpy/devel/include/hpy/cpython/misc.h
```diff
@@ -276,6 +276,32 @@ HPyAPI_FUNC HPy HPy_CallTupleDict(HPyContext *ctx, HPy callable, HPy args, HPy k
...
+#define PyObject_Vectorcall _PyObject_Vectorcall
+#endif
+
+HPyAPI_FUNC HPy HPy_Call(HPyContext *ctx, HPy callable, const HPy *args, size_t nargs, HPy kwnames)
+{
+    if (sizeof(HPy) == sizeof(PyObject *)) {
+        return _py2h(PyObject_Vectorcall(_h2py(callable), (PyObject *const *)args, nargs, _h2py(kwnames)));
+    }
+    return ctx_Call(ctx, callable, args, nargs, kwnames);
+}
```
antocuni, line 289: I'm confused by this check. Can it ever happen that `sizeof(HPy)` != `sizeof(PyObject *)`?
There is already code around which relies on the fact that in CPython mode `HPy` is basically a `PyObject *` in disguise: so if this check is needed, then we have lots of problems everywhere else.
If it's not needed, then we should kill it.

hpy/devel/include/hpy/inline_helpers.h
```diff
@@ -459,4 +459,30 @@ HPySlice_AdjustIndices(HPyContext *_HPy_UNUSED_ARG(ctx), HPy_ssize_t length,
     return 0;
 }
 
+HPyAPI_INLINE_HELPER HPy
+HPy_CallMethodTupleDict_s(HPyContext *ctx, HPy receiver, const char *name, HPy args, HPy kw)
```
antocuni, line 463: I wonder whether we should swap `receiver` and `name` here, to better match the signature in `HPy_CallMethod`

hpy/devel/src/runtime/ctx_call.c
```diff
@@ -36,3 +39,81 @@ ctx_CallTupleDict(HPyContext *ctx, HPy callable, HPy args, HPy kw)
     }
     return _py2h(obj);
 }
+
+#if PY_VERSION_HEX < 0x03090000
+#define PyObject_Vectorcall _PyObject_Vectorcall
```
antocuni, line 44:
```suggestion
#    define PyObject_Vectorcall _PyObject_Vectorcall
```

hpy/devel/src/runtime/ctx_call.c
```diff
@@ -36,3 +39,81 @@ ctx_CallTupleDict(HPyContext *ctx, HPy callable, HPy args, HPy kw)
...
+    for (size_t i = 0; i < n_all_args; i++) {
+        args[i+1] = _h2py(h_args[i]);
+    }
+
+    return _py2h(PyObject_Vectorcall(_h2py(h_callable), args+1,
+            nargs | PY_VECTORCALL_ARGUMENTS_OFFSET, kwnames));
+}
+
+#if PY_VERSION_HEX < 0x03090000
+#undef PyObject_Vectorcall
```
antocuni, line 77:
```suggestion
#    undef PyObject_Vectorcall
```

hpy/devel/src/runtime/ctx_call.c
```diff
@@ -36,3 +39,81 @@ ctx_CallTupleDict(HPyContext *ctx, HPy callable, HPy args, HPy kw)
...
+    PyObject *kwnames;
+    size_t n_all_args;
+
+    if (HPy_IsNull(h_kwnames)) {
+        kwnames = NULL;
+        n_all_args = nargs;
+    } else {
+        kwnames = _h2py(h_kwnames);
+        assert(kwnames != NULL);
+        assert(PyTuple_Check(kwnames));
```
antocuni, line 59: does this need to be an `assert()`, or it should be a more proper check?
What happens if the user passes a non-tuple as kwnames?
Maybe it's good enough to have an `assert()` here, but then we should have a more proper check in debug mode

antocuni (approved):
LGTM!

hpy/devel/include/hpy/inline_helpers.h
fangerer: good point; I'll do that

PR #253, opened 2021-10-20
Make the documentation up to date with the API changes

steve-s:
Fixes the out of date code samples in the documentation.

Adds an example module next to the documentation and in the documentation refers to code snippets from that example module. Note that I'd like to take all the code samples verbatim from existing and tested code, including `setup.py`, for example, and in such case this has to be a separate module. Installing it and running its simple tests takes about 2s on my machine, so I don't think it will be an issue if we have a few separate example modules.

I would like to first check what people think about this approach and then I'd like to go about fixing other code samples, which may require one more example module.

Adds a new make target 'docs-examples-tests' and new CI job that runs this target.

@hodgestar I suggest we have one make target and one CI job that runs all the examples from the documentation (I adapted the CI config from your PR). Furthermore, I am thinking about putting all the documentation examples into `docs/example/`. What do you think?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -138,6 +138,36 @@ jobs:
           python -m pip install pytest pytest-valgrind
           make valgrind
 
+  docs_examples_tests:
+    name: Porting example tests
+    runs-on: ${{ matrix.os }}
+    continue-on-error: true
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: ['3.9']
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install/Upgrade Python dependencies
+        run: python -m pip install --upgrade pip wheel
+        shell: bash
+
+      - name: Install HPy
+        run: python -m pip install .
+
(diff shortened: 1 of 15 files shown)


mattip:
The windows run of the test step takes 5 seconds vs. the 3 seconds for linux. It is hard to judge what will happen as the number of small snippets increases. The main tests take around 27 minutes on windows and 5 minutes on linux. I would not want to design yet more "small snippet" tests. From my experience, most of the time on windows is due to opening files, which is much much slower on windows.

hodgestar:
> The windows run of the test step takes 5 seconds vs. the 3 seconds for linux. It is hard to judge what will happen as the number of small snippets increases. The main tests take around 27 minutes on windows and 5 minutes on linux. I would not want to design yet more "small snippet" tests. From my experience, most of the time on windows is due to opening files, which is much much slower on windows.

@mattip Do you have a suggested approach? Not having tests will just mean the documentation examples are repeatedly broken without anyone noticing.

We could combine them into one set of documentation tests (with sub folders for different sections but only a single pytest run and GitHub Action job) if that would help?

steve-s:
I think that most of the code snippets from docs could be crammed into one module and even one C file. Except the "getting started" section, which is a bit special, because there we want to show the whole thing including `HPyDef` array initialization and `setup.py`, so there is no room left to put more stuff into the module. That won't be the case for more advanced code snippets I believe.

Issue #236, opened 2021-09-03
Debug mode: enforce lifetime of raw data pointers returned from the API


steve-s:
Some APIs, such as `HPyUnicode_AsUTF8AndSize`, return raw data pointers. The debug mode should:
* make a copy of the data and return that copy
* erase the copy once the intended validity of the raw data pointer ends (when the corresponding HPy handle is closed)
* nice to have: allocate the copy in write protected memory if the API returns what should be a read-only view on the data

PR #254, opened 2021-10-20
Extend debug tests (rebased)

steve-s:
Several tests infrastructure refactorings that will be used more in my next PR (#236) + few new tests.

* factor out code for running python subprocesses in the tests (fixture `python_subprocess`)
   * runs the subprocess tests in all the supported "ABI modes"
   * updated `HPy_FatalError` tests to use this new internal API
* add `hpy_debug_cature` fixture for capturing calls to the "invalid handle" hook in the debug mode
   * allow to reset the hook back to `NULL` (via setting it to `None` on the Python level)
* Split test_handles to test_handles_leak and test_handles_invalid
   * test_handles_invalid can run with the leak detector on
   * add few more tests to test_handles_invalid

@rlamy hope this will work fine with the PyPy's apptests

Note: rebased version of https://github.com/hpyproject/hpy/pull/247

### DIFF
--- a/test/debug/test_handles_invalid.py
+++ b/test/debug/test_handles_invalid.py
@@ -8,6 +8,30 @@ def hpy_abi():
         yield "debug"
 
 
+def test_no_invalid_handle(compiler, hpy_debug_capture):
+    # Basic sanity check that valid code does not trigger any error reports
+    mod = compiler.make_module("""
+        HPyDef_METH(f, "f", f_impl, HPyFunc_O)
+        static HPy f_impl(HPyContext *ctx, HPy self, HPy arg)
+        {
+            HPy x = HPyLong_FromLong(ctx, 42);
+            HPy y = HPyLong_FromLong(ctx, 2);
+            HPy arg_dup = HPy_Dup(ctx, arg);
+            HPy_Close(ctx, y);
+            HPy b = HPy_Dup(ctx, x);
+            HPy_Close(ctx, x);
+            HPy_Close(ctx, arg_dup);
+            return b;
+        }
+
+        @EXPORT(f)
+        @INIT
+    """)
+    assert mod.f("hello") == 42
+    assert mod.f("world") == 42
+    assert hpy_debug_capture.invalid_handles_count == 0
+
+
 def test_cant_use_closed_handle(compiler, hpy_debug_capture):
     mod = compiler.make_module("""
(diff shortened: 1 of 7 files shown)


test/debug/test_handles_invalid.py
```diff
@@ -0,0 +1,109 @@
...
+    """)
+    mod.f("hello leaks!")
+    assert hpy_debug_capture.invalid_handles_count == 0
+    mod.g()
+    assert hpy_debug_capture.invalid_handles_count == 1
+
+
+def test_invalid_handle_crashes_python_if_no_hook(compiler, python_subprocess, fatal_exit_code):
+    if not SUPPORTS_SYS_EXECUTABLE:
+        pytest.skip("no sys.executable")
```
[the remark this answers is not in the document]
rlamy: Yes, that's fine for pypy. I'm less ure about `from test.support import SUPPORTS_SYS_EXECUTABLE`, but that's easy to change in the pypy tests.

rlamy (approved):
LGTM, sorry for the late review.

steve-s:
> LGTM, sorry for the late review.

np and thanks for the review. I'll consider adding `@pytest.mark.skipif` in some other PR and merge this as is.

test/support.py
```diff
@@ -272,6 +287,45 @@ def load_cpython_module(self, name, so_filename):
...
+            load_module = "import {} as mod;".format(mod.name)
+        if self.verbose:
+            print("\n---\nExecuting in subprocess: {}\n".format(load_module + code))
+        result = subprocess.run([sys.executable, "-c", load_module + code], env=env,
+                                stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        if self.verbose:
+            out = result.stdout.decode('ascii')
+            err = result.stderr.decode('ascii')
+            print("\n---\n{out}\n{err}\n---\n".format(out=out, err=err))
+        return result
```
antocuni, line 326: this looks potentially problematic: if for any reason the subprocess prints a non-ascii char, this would throw `UnicodeDecodeError`, I think.
Probably it's safer to `.decode('latin-1')` (which always works because it maps the whole range 0-255, even if it might result in funny question marks).

steve-s: I sneaked the fix of this into: https://github.com/hpyproject/hpy/pull/259

Issue #255, opened 2021-10-22
test_gc_track fails on debug build of CPython 3.9.2


steve-s:
I am using CPython 3.9.2 built from sources using `../configure --with-pydebug` and I noticed that this test fails:

```
> make && python -m pytest -k test_gc_track -s -v test
...
platform linux -- Python 3.9.2, pytest-6.2.5, py-1.10.0, pluggy-1.0.0 -- <....>
cachedir: .pytest_cache
rootdir: <....>
collected 747 items / 741 deselected / 6 selected                                                                                                                 

test/test_hpyfield.py::TestHPyField::test_gc_track[cpython] ../Modules/gcmodule.c:2162: visit_validate: Assertion failed: PyObject_GC_Track() object is not valid
Enable tracemalloc to get the memory block allocation traceback

object address  : 0x7f395e626780
object refcount : 1
object type     : 0x558c7a375ed0
object type name: mytest.Pair
object repr     : <mytest.Pair object at 0x7f395e626780>

Fatal Python error: _PyObject_AssertFailed: _PyObject_AssertFailed
Python runtime state: initialized
```

(tested on rev 58edcd8)

PR #256, opened 2021-11-01
CI job that runs main tests on CPython debug build

steve-s:


hodgestar:
Some notes on build times:
* This adds a 5m 30s build step, but many build steps are run in parallel, so the effect on the overall time looks negligible.
* The total build time is largely equal to the build time of the Valgrind step.

hodgestar (approved):
Looks good. Thanks!

steve-s:
I am not sure that it did what it should have, because the tests passed. I'll play with this PR a bit to find out what's going on.

steve-s:
I am convinced now that HPy was built and the tests have been executed on debug build of CPython, but with this approach we can cover only the CPython version that is the default in Ubuntu, which is at this point `3.8`, which happens to not trigger issue #255 that triggered this exercise. 

Alternative would be to use 3rd party action: https://github.com/deadsnakes/action. I would be up for giving it a go. Are there any issues we see with using 3rd party GitHub action?

mattip:
Using the deadsnakes action sounds reasonable.

steve-s:
Ok, so now we get the expected error. It happens on 3.7 and 3.9 and weirdly not on 3.8.

On unrelated note: when looking at the logs I think I may found out why Windows builds are slower: it uses `pytest-forked` on POSIX systems, but not on Windows (no forking there), so there the tests do not run in parallel.

steve-s:
I rebased the PR on master and added logic to skip the failing tests if debug build of CPython is detected.

hodgestar (approved):
Happy for this to be merged as soon as the CI run finishes successfully.

I was going to suggest adding an `is_debug_build` helper, but since these checks should all be removed eventually, let's not add a helper that we don't have a long term use case for.

PR #257, opened 2021-11-01
Don't clobber legacy tp_dealloc unnecessarily.

rlamy:
We now only use hpytype_dealloc if the user supplied HPy_tp_traverse
or HPy_tp_destroy.

### DIFF
--- a/hpy/devel/src/runtime/ctx_type.c
+++ b/hpy/devel/src/runtime/ctx_type.c
@@ -10,6 +10,7 @@
 #endif
 
 static bool has_tp_traverse(HPyType_Spec *hpyspec);
+static bool has_tp_destroy(HPyType_Spec *hpyspec);
 
 
 /* This is a hack: we need some extra space to store random data on the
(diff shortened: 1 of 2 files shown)


hpy/devel/src/runtime/ctx_type.c
```diff
@@ -352,7 +353,8 @@ create_slot_defs(HPyType_Spec *hpyspec, HPy_ssize_t base_member_offset,
     hpyslot_count++;        // Py_tp_methods
     hpyslot_count++;        // Py_tp_members
     hpyslot_count++;        // Py_tp_getset
-    hpyslot_count++;        // Py_tp_dealloc
+    if (has_tp_traverse(hpyspec) || has_tp_destroy(hpyspec))
+        hpyslot_count++;        // Py_tp_dealloc
```
antocuni, line 357: I fear that this style (which was introduced by me, so I'm the culprit :)) might quickly become too hard to read/maintain, especially because the logic needs to be kept in sync with the lines 435-438 below.
Maybe introducing a helper `needs_py_tp_dealloc()` will result in better code?

antocuni (commented):
LGTM, but I have one question: what happens if I specify both a `HPy_tp_traverse` (or `HPy_tp_destroy`) and a legacy `Py_tp_dealloc`? If I read the logic correctly, the legacy `Py_tp_dealloc` will be overwritten by our custom one.

I think we should catch and forbid this case (or alternatively, add some complicate logic to handle it correctly)

hpy/devel/src/runtime/ctx_type.c
rlamy: Good point, done!

rlamy:
I don't think it makes sense to have both a legacy `Py_tp_dealloc` and `HPy_tp_traverse` or `HPy_tp_destroy`, since the former completely takes over object destruction. So I disallow it now.

antocuni (approved):
LGTM

PR #259, opened 2021-11-03
Debug mode: enforce lifetime of char pointers returned from the API

steve-s:
Implementation of https://github.com/hpyproject/hpy/issues/236.

~This PR is built on top of https://github.com/hpyproject/hpy/pull/254. It probably makes sense to first review & merge that PR, or look at the diff the two PRs (available here: https://github.com/steve-s/hpy/pull/1)~

### DIFF
--- a/hpy/debug/src/_debugmod.c
+++ b/hpy/debug/src/_debugmod.c
@@ -75,13 +75,19 @@ static UHPy get_open_handles_impl(HPyContext *uctx, UHPy u_self, UHPy u_gen)
 }
 
 HPyDef_METH(get_closed_handles, "get_closed_handles", get_closed_handles_impl,
-            HPyFunc_NOARGS, .doc=
+            HPyFunc_VARARGS, .doc=
             "Return a list of all the closed handle in the cache")
-static UHPy get_closed_handles_impl(HPyContext *uctx, UHPy u_self)
+static UHPy get_closed_handles_impl(HPyContext *uctx, UHPy u_self, HPy *args, HPy_ssize_t nargs)
 {
     HPyContext *dctx = hpy_debug_get_ctx(uctx);
     HPyDebugInfo *info = get_info(dctx);
-    return build_list_of_handles(uctx, u_self, &info->closed_handles, 0);
+    long gen = 0;
+    if (nargs > 0) {
+        gen = HPyLong_AsLong(uctx, args[0]);
+        if (HPyErr_Occurred(uctx))
+            return HPy_NULL;
+    }
+    return build_list_of_handles(uctx, u_self, &info->closed_handles, gen);
 }
 
 HPyDef_METH(get_closed_handles_queue_max_size, "get_closed_handles_queue_max_size",
@@ -105,6 +111,30 @@ static UHPy set_closed_handles_queue_max_size_impl(HPyContext *uctx, UHPy u_self
     if (HPyErr_Occurred(uctx))
         return HPy_NULL;
     info->closed_handles_queue_max_size = size;
+    return HPy_Dup(uctx, uctx->h_None);
+}
+
+HPyDef_METH(get_protected_raw_data_max_size, "get_protected_raw_data_max_size",
+get_protected_raw_data_max_size_impl, HPyFunc_NOARGS, .doc=
+"Return the maximum size of the retained raw memory associated with closed handles")
+static UHPy get_protected_raw_data_max_size_impl(HPyContext *uctx, UHPy u_self)
+{
+    HPyContext *dctx = hpy_debug_get_ctx(uctx);
+    HPyDebugInfo *info = get_info(dctx);
+    return HPyLong_FromSsize_t(uctx, info->protected_raw_data_max_size);
+}
+
+HPyDef_METH(set_protected_raw_data_max_size, "set_protected_raw_data_max_size",
+set_protected_raw_data_max_size_impl, HPyFunc_O, .doc=
+"Set the maximum size of the retained raw memory associated with closed handles")
+static UHPy set_protected_raw_data_max_size_impl(HPyContext *uctx, UHPy u_self, UHPy u_size)
+{
+    HPyContext *dctx = hpy_debug_get_ctx(uctx);
+    HPyDebugInfo *info = get_info(dctx);
+    HPy_ssize_t size = HPyLong_AsSize_t(uctx, u_size);
+    if (HPyErr_Occurred(uctx))
+        return HPy_NULL;
+    info->protected_raw_data_max_size = size;
     return HPy_Dup(uctx, uctx->h_None);
 }
 
@@ -177,6 +207,18 @@ static UHPy DebugHandle_is_closed_get(HPyContext *uctx, UHPy self, void *closure
 {
     DebugHandleObject *dh = DebugHandleObject_AsStruct(uctx, self);
     return HPyBool_FromLong(uctx, dh->handle->is_closed);
+}
+
+HPyDef_GET(DebugHandle_raw_data_size, "raw_data_size", DebugHandle_raw_data_size_get,
+.doc="Size of retained raw memory. FOR TESTS ONLY.")
+static UHPy DebugHandle_raw_data_size_get(HPyContext *uctx, UHPy self, void *closure)
+{
+    DebugHandleObject *dh = DebugHandleObject_AsStruct(uctx, self);
+    if (dh->handle->associated_data) {
+        return HPyLong_FromSsize_t(uctx, dh->handle->associated_data_size);
+    } else {
+        return HPyLong_FromLong(uctx, -1);
+    }
 }
 
 HPyDef_SLOT(DebugHandle_cmp, DebugHandle_cmp_impl, HPy_tp_richcompare)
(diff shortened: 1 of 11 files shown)


test/test_hpyfield.py
```diff
@@ -88,218 +88,3 @@ def EXPORT_PAIR_TYPE(self, *defines):
                 .defines = Pair_defines
             };
```
steve-s, line 89: Opps, this test should not have been removed, it's some local debugging/tinkering leftover. Will revert the file back to its master contents.

hpy/debug/src/_debugmod.c
```diff
@@ -75,13 +75,19 @@ static UHPy get_open_handles_impl(HPyContext *uctx, UHPy u_self, UHPy u_gen)
...
+            HPyFunc_VARARGS, .doc=
             "Return a list of all the closed handle in the cache")
-static UHPy get_closed_handles_impl(HPyContext *uctx, UHPy u_self)
+static UHPy get_closed_handles_impl(HPyContext *uctx, UHPy u_self, HPy *args, HPy_ssize_t nargs)
 {
     HPyContext *dctx = hpy_debug_get_ctx(uctx);
     HPyDebugInfo *info = get_info(dctx);
-    return build_list_of_handles(uctx, u_self, &info->closed_handles, 0);
+    long gen = 0;
+    if (nargs > 0) {
```
antocuni, line 85: we should maybe complain if `nargs > 1`?

hpy/debug/src/debug_internal.h
```diff
@@ -88,6 +88,10 @@ typedef struct DebugHandle {
     UHPy uh;
     long generation;
     bool is_closed;
+    // pointer to and size of any raw data associated with
+    // the lifetime of the handle:
+    void *associated_data;
+    HPy_ssize_t associated_data_size;
```
antocuni, line 94: this looks certainly good for this specific PR.
However, we are wasting some space even for handles which don't require these fields (which are most of them).
For the future, we might want to design a sort of "`DebugHandle` subclassing" to make it possible to create different kinds of handles with different kinds of extra data, maybe?

hpy/debug/src/memprotect.c
```diff
@@ -0,0 +1,60 @@
...
+
+void *raw_data_copy(const void* data, HPy_ssize_t size, bool write_protect) {
+    void* new_ptr = mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+    memcpy(new_ptr, data, size);
+    if (write_protect) {
+        mprotect(new_ptr, size, PROT_READ);
+    }
+    return new_ptr;
+}
+
```
antocuni, line 22: I think you addressed this concern in another comment but I wanted to double check. What happens if you `mmap()` lots of small buffers? The returned regions are aligned to page boundary (so, 4096 in general), but I assume that this consumes only the virtual address space and not the physical memory?

hpy/debug/src/memprotect.c
```diff
@@ -0,0 +1,60 @@
...
+void raw_data_protect(void* data, HPy_ssize_t size) {
+    // Override the data with some garbage in hope that the program will
+    // eventually crash or give incorrect result if it reads the garbage
+    char val = 0;
+    for (HPy_ssize_t i = 0; i < size; ++i) {
+        ((char*)data)[i] = 143 + val;
+        val = (val + 1) % 10;
+    }
+}
+
```
antocuni, line 54: I wonder whether we could write something which is more readable in a debugger, like `0xDEADBEEF` or `"THIS MEMORY WAS FREED"`. But it's also true that this implementation will almost never be used in practice as linux is covered by the other one

antocuni (commented):
Looks like a brilliant idea to me :100: 

I think it deserves some better documentation though. I don't think we have user-facing docs for the debug mode yet, but for sure you should add some lenghty comment at the start of `_debug_internal.h` which explains the general idea of this check and how the various `get_protected_raw_data_size` & co. works, how/when the raw memory is allocated and freed, etc., because it's a bit hard to follow the logic by just reading the code.

Also, random idea for the future: I wonder whether we can write an implementation of `raw_data_*` which somehow integrates with [AddressSanitizer](https://github.com/google/sanitizers/wiki/AddressSanitizer) and/or whether it would have any advantage over the current mmap-based implementation.

hpy/debug/src/debug_internal.h
steve-s: Yes, I was concerned about this too. For now I don't see any other significantly better solution. I think the subclassing would not play well with the reusing of closed handles? In the future, in case we need to attach even more data to the debug handles, maybe we can use indirection with just one field in `DebugHandle` pointing to `DebugHandleData` or such and that would be allocated only if needed.

steve-s:
> Also, random idea for the future: I wonder whether we can write an implementation of `raw_data_*` which somehow integrates with [AddressSanitizer](https://github.com/google/sanitizers/wiki/AddressSanitizer) and/or whether it would have any advantage over the current mmap-based implementation.

Right, the advantage would be that instead of just failing, we'd get a trace and all the info that AddressSanitizer provides. That would be cool. I could not find any API that AddressSanitizer would provide though. We would need to be able to mark some memory as "read only" and then as "invalid".

In general providing stack trace and some additional info would be useful for other debugging features like the invalid handle detection.

hpy/debug/src/debug_internal.h
antocuni: yes, if we do "subclassing" we would need to maintain a separate queue of closed handle for each different kind (or at least for each different sizeof()).
`DebugHandleData` works in theory but I don't know how good it is in practice, because you pay the cost of the pointer + the hidden space cost of an extra malloc (but I don't know how much they are).

Another idea to keep into consideration is to use a `union` if/when we will need more kind of data.

hpy/debug/src/memprotect.c
antocuni: > I think you cannot get smaller amount of physical memory than a page? 

indeed, I think you are correct.

> On the other hand, the number of the buffers is limited by the max number of handles, so it should be fine I believe [cut]

yes, this looks correct as well. I vote for including your computation/explanation in the "big docstring which explains how it works".

> This may mean that debug mode run may run out of memory when normal run wouldn't, but that may happen for other reasons too, so not sure if this is a big concern. The necessary amount of memory is "just" 2x, i.e., grows linearly, so I think it's fine?

yes, I think it's fine if debug mode requires more memory. Maybe at some point we can think of introducing optional checks (like, I want to enable leak checks but not raw buffers to save memory or so) but I don't think it's too important now.

antocuni:
> Right, the advantage would be that instead of just failing, we'd get a trace and all the info that AddressSanitizer provides. That would be cool. I could not find any API that AddressSanitizer would provide though. We would need to be able to mark some memory as "read only" and then as "invalid".

yes but now that I think of it, this would work only if the user compiles its own code with ASan. In that case, it's probably enough to just `free()` the temporary buffer and let ASan to detect it as normal.
A good solution could be to offer different strategies to handle raw data so that the user can select the one which fits better his/her needs:
- `mmap()`-based (if available)
- "poisoning-based" (the default if mmap is not available)
- `free()`-based (useful with ASan)
- ...

> In general providing stack trace and some additional info would be useful for other debugging features like the invalid handle detection.

yes, that's also something which I wanted to add (and that should be optional since it's probably costly). Each `DebugHandle` should record the stacktrace of where it was created and/or closed.
Integrating with this PR should be "easy": we could install a fault handler, check whether the memory being accessed belongs to one of the closed handles and report it accordingly.
It's not necessary to do that in this PR of course, but could be a good idea to open an issue so that we remember the idea

steve-s:
Last time I checked getting a stack trace in C in at least so-so portable way was not a simple thing to do. There are some libraries providing that functionality. What do you think about bringing some 3rd party dependency for that? We can probably make it at least optional with `dlopen`/`dlsym` instead of linking with it directly, or we can inline it, if it has appropriate license.

antocuni:
> Last time I checked getting a stack trace in C in at least so-so portable way was not a simple thing to do. There are some libraries providing that functionality. What do you think about bringing some 3rd party dependency for that? We can probably make it at least optional with `dlopen`/`dlsym` instead of linking with it directly, or we can inline it, if it has appropriate license.

yes, it's not portable and doesn't always work but that's life.
For linux, the simplest thing to do is to use `backtrace` which should be available everywhere. The most general solution is probably `libunwind` though. Whether to introduce a hard dependency or not, I don't know. It's probably better to discuss these details in an issue of its own than here, so we can make sure that the discussion will not be lost

hpy/debug/src/memprotect.c
steve-s: Done. I went for `0xbad0da7a`, because we want the data to be invalid when decoded. In the tests of this feature (`test_charptr.py`), when mmap is not available, we're now checking that the program fails with `UnicodeDecodeError`.

steve-s:
> I think it deserves some better documentation though. I don't think we have user-facing docs for the debug mode yet, but for sure you should add some lenghty comment at the start of `_debug_internal.h` which explains the general idea of this check and how the various `get_protected_raw_data_size` & co. works, how/when the raw memory is allocated and freed, etc., because it's a bit hard to follow the logic by just reading the code.

this is done now

steve-s:
@antocuni I believe I addressed all your comments. If you want, please, re-review the PR or let me know that you'd like to do so later. Otherwise I'd merge this next week.

PR #260, opened 2021-11-05
Implement HPy_tp_finalize

rlamy:
This implements `HPy_tp_finalize` and adds the `HPyFunc_DESTRUCTOR` `HPyFunc_Signature`.

I'm a bit unsure about how `hpytype_dealloc()` and `HPy_tp_finalize` should interact. Currently, `HPy_tp_finalize` has no effect on whether we supply `hpytype_dealloc()` as the type's tp_dealloc, which allows the user to implement a legacy tp_dealloc if they need to. OTOH, that means that `test_tp_finalize_nongc()` passes only because `subtype_dealloc()` (CPython's default tp_dealloc for heaptypes)  happens to do the right thing in this case.

### DIFF
--- a/test/test_slots.py
+++ b/test/test_slots.py
@@ -66,6 +66,88 @@ class TestSlots(HPyTest):
         gc.collect()
         assert mod.get_destroyed_x() == 7
 
+    def test_tp_finalize_nongc(self):
+        import gc
+        mod = self.make_module("""
+            @DEFINE_PointObject
+            @DEFINE_Point_new
+
+            static long finalized_x;
+
+            HPyDef_SLOT(Point_finalize, Point_finalize_impl, HPy_tp_finalize)
+            static void Point_finalize_impl(HPyContext *ctx, HPy obj)
+            {
+                PointObject *point = PointObject_AsStruct(ctx, obj);
+                finalized_x += point->x;
+            }
+
+            HPyDef_METH(get_finalized_x, "get_finalized_x", get_finalized_x_impl, HPyFunc_NOARGS)
+            static HPy get_finalized_x_impl(HPyContext *ctx, HPy self)
+            {
+                return HPyLong_FromLong(ctx, finalized_x);
+            }
+
+            @EXPORT_POINT_TYPE(&Point_new, &Point_finalize)
+            @EXPORT(get_finalized_x)
+            @INIT
+       """)
+        point = mod.Point(7, 3)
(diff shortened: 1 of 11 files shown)


test/test_slots.py
```diff
@@ -66,6 +66,88 @@ def test_tp_destroy(self):
...
+            @EXPORT_POINT_TYPE(&Point_new, &Point_finalize)
+            @EXPORT(get_finalized_x)
+            @INIT
+       """)
+        point = mod.Point(7, 3)
+        assert mod.get_finalized_x() == 0
+        del point
+        gc.collect()
+        assert mod.get_finalized_x() == 7
+        gc.collect()
```
fangerer, line 99: In Graalpython, our `gc.collect()` does not ensure that the GC is really finished after the call (could be a parallel GC) and also does not ensure that the GC will collect that object (if there is no memory pressure, the GC could decide to just leave it).

Long story short: Could you add something like `supports_sync_gc` to class `HPyTest` (just like `supports_refcounts`) and use it here?

hodgestar: @fangerer Is there a way to ensure GC collection on GraalPython? Perhaps we can instead have a `.gc_collect()` helper method that PyPy and GraalPython could implement differently if needed? Otherwise it becomes very hard to test these kinds of things.

fangerer: No, sorry, I don't see a way to ensure that. What we usually do is to repeat the test in a loop (with a maximum number of retries) and also insert a sleep to give the GC enough time. But that's also just best-effort and does not ensure success.

What about introducing a custom pytest mark for such tests?

rlamy: Adding a pytest mark seems like a good solution - CPython and pypy would just ignore it and GraalPython would deal with it however it wants. I'll let the GraalPython folks implement it though, since you already know which tests cause issues and how to handle them.

fangerer: Okay, I'll implement it and push to this PR.

rlamy:
I'd like to get this merged ASAP.
@fangerer, would it be OK to merge now and add the mark in a follow-up PR?

fangerer:
@rlamy Sorry, I forgot about this (and was on vacation). Yes, please go ahead and merge. I'll do it in a separate PR.

fangerer (approved):

rlamy:
Thanks!

PR #261, opened 2021-11-12
Fix: Do not visit uninitialized HPyField during traversal.

### LINKED ISSUE #255: test_gc_track fails on debug build of CPython 3.9.2

fangerer:
Should fix #255 .
I'm not sure why we did not copy everything from `Py_VISIT` to `HPy_VISIT` but clearly, `Py_VISIT` will only visit non-null `PyObject *`.
The only drawback now is that we need to do an additional dereference of `hpyfield`.
However, the alternative would be to expose something like `PyObject_GC_Track` in HPy and don't it in `ctx_New`. But I would really not like that.

### DIFF
--- a/hpy/devel/include/hpy.h
+++ b/hpy/devel/include/hpy.h
@@ -100,6 +100,7 @@ typedef struct { intptr_t _i; } HPyTracker;
 #define HPy_IsNull(h) ((h)._i == 0)
 
 #define HPyField_NULL ((HPyField){0})
+#define HPyField_IsNull(f) ((f)._i == 0)
 
 /* Convenience functions to cast between HPy and void*.  We need to decide
    whether these are part of the official API or not, and maybe introduce a
--- a/hpy/devel/include/hpy/hpyfunc.h
+++ b/hpy/devel/include/hpy/hpyfunc.h
@@ -97,9 +97,11 @@ typedef int (*HPyFunc_visitproc)(HPyField *, void *);
  */
 #define HPy_VISIT(hpyfield)                                             \
     do {                                                                \
-        int vret = visit(hpyfield, arg);                                \
-        if (vret)                                                       \
-            return vret;                                                \
+        if (!HPyField_IsNull(*hpyfield)) {                              \
+            int vret = visit(hpyfield, arg);                            \
+            if (vret)                                                   \
+                return vret;                                            \
+            }                                                           \
     } while (0)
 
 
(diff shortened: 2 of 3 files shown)


hpy/devel/include/hpy.h
```diff
@@ -100,6 +100,7 @@ typedef struct { intptr_t _i; } HPyTracker;
 #define HPy_IsNull(h) ((h)._i == 0)
 
 #define HPyField_NULL ((HPyField){0})
+#define HPyField_IsNull(f) ((f)._i == 0)
```
fangerer, line 103: I would actually not like to expose `HPyField_IsNull`  but since `HPy_VISIT` is public, there are not many options. Anyway, I think it's not a problem since we already expose `HPyField_NULL` and so already implicitly specify that uninitialized `HPyField` has value `0`.

rlamy: This macro is identical to `HPy_IsNull`, I wonder if we should turn them both into static inline functions to prevent confusions.

rlamy (approved):
LGTM

hpy/devel/include/hpy.h
fangerer: IIRC we had a comment in an older version of `hpy.c` that also asked this question and since `HPy_FromVoidP` and `HPy_AsVoidP` are also inline functions, it would make sense to me.
Let me merge this PR and I'll open a separate PR to do that. Then we can invite people to discuss that.

Issue #189, opened 2021-03-26
Refactor HPyModuleDef names


antocuni:
https://github.com/hpyproject/hpy/blob/cc98cac7f623572c4b0096c1767d738b57b4beae/hpy/devel/include/common/hpymodule.h#L7-L16

at the beginning, we had to introduce `dummy` because in CPython mode, `HPyModuleDef` was a typedef to `PyModuleDef`. Nowadays, we use our own `HPyModuleDef` even in CPython mode, so we should kill this historical leftover.

Moreover, we should consider to remove the `m_` prefix in front of fields. We don't use these kind of prefixes anywhere else AFAIK

hodgestar:
Things might get a bit more complicated:

`PyModuleDef` is sort of a Python type -- it has a `PyObject_HEAD and is returned as a `PyObject` by `PyModuleDef_Init`.  Of course one also creates instances of the struct directly.  Maybe this means that HPyModuleDef should become an actual type, or that there should be a separate type (maybe `HPyModuleDefObject` ?) that is returned by `HPyModuleDef_Init`.

antocuni:
> `PyModuleDef` is sort of a Python type

yes but I don't think it's relevant here. In HPy, the only thing that you can do with an `HPyModuleDef` is to call `HPyModule_Create`: the actual `PyMethodDef` is created under the hood. Look at the code below, the `PyModule_HEAD_INIT` part is memcopied from `empty_moduledef`, and `dummy` is completely ignored:

https://github.com/hpyproject/hpy/blob/459758683e764a142ae82e9973c1f0730f651eaa/hpy/devel/src/runtime/ctx_module.c#L10-L30

PR #262, opened 2021-11-12
Remove HPyModuleDef.dummy and remove m_ prefix of fields.

fangerer:
As suggested in #189, removes field `HPyModuleDef.dummy` and also removes the `m_` prefix.
This is a breaking change, i.e, will cause that current HPy modules will no longer compile without adoption and also introduces binary compatibility. Therefore, I think we should do such changes earlier than later.

If approved, I will also migrate all other modules in `hpyproject`.


auvipy (approved):

hodgestar (approved):
Looks good to me! Thanks @fangerer for doing this cleanup.

PR #264, opened 2021-11-18
Debug mode: detect invalid handles returned from downcalls

### LINKED ISSUE #258: Debug mode: check that returned value is a valid handle
I noticed that CPython debug build fails on an assertion with code similar to this: 
```
static HPy g_impl(HPyContext *ctx, HPy self, HPy arg) {
   return arg;
}
```

in the current scheme of the debug mode where we check the handle validity in `HPy_Close` and watch for leaked handles, this is not detected. I think we need to add a check of the validity of return values into the `debug_ctx_CallRealFunctionFromTrampoline`.

steve-s:
Fixes #258. See the test for use-cases.

### DIFF
--- a/test/support.py
+++ b/test/support.py
@@ -9,9 +9,11 @@ PY2 = sys.version_info[0] == 2
 # the current Python to be launched via, e.g., `python_subprocess.run(...)`.
 # By default is `True` if sys.executable is set to a true value.
 SUPPORTS_SYS_EXECUTABLE = bool(getattr(sys, "executable", None))
-IS_VALGRIND_RUN = 'valgrind' in os.environ.get('PYTHONMALLOC', '')
 # True if we are running on the CPython debug build
 IS_PYTHON_DEBUG_BUILD = hasattr(sys, 'gettotalrefcount')
+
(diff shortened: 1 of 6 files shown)


steve-s:
@antocuni if you have a moment, please review. This is basically what you suggested in https://github.com/hpyproject/hpy/issues/258 + some tests.

test/support.py
```diff
@@ -9,6 +9,11 @@
 # the current Python to be launched via, e.g., `python_subprocess.run(...)`.
 # By default is `True` if sys.executable is set to a true value.
 SUPPORTS_SYS_EXECUTABLE = bool(getattr(sys, "executable", None))
+# True if we are running on the CPython debug build
+IS_PYTHON_DEBUG_BUILD = hasattr(sys, 'gettotalrefcount')
+
+def is_valgrind_run(config):
```
steve-s, line 15: Note to myself: this is probably unused leftover

fangerer (approved):

steve-s:
Rebased on master to be able to pass the CPython debug build jobs.

Issue #188, opened 2021-03-26
Re-enable cppcheck


antocuni:
In commit a91f8f9 (PR #187 ) I had to disable cppcheck, because it failed for no good reason.

I have no idea of what's happening, we should try again in some days/weeks and maybe re-enable it.

thedrow:
On CPPCheck 2.5 `make cppcheck` passes successfully.

This sounds like a weird bug that should be reported.

antocuni:
do you feel like making a PR with `cppcheck` 2.5 and see if it passes the CI?

thedrow:
Since snap didn't properly work, we'd probably need to install it using Linuxbrew, I guess.
If that's OK with you, I can do it.

antocuni:
Sorry for the late response. I don't know what is linuxbrew but as long as it works, that's ok with me :)

thedrow:
Homebrew for linux...

fangerer:
I've re-enabled `cppcheck` but using Ubuntu's version `1.90`. I think this is better than nothing for now. We may think about moving to a newer version.

mattip:
cppcheck is failing CI, and I think it is not a good investment of the core team to fix it. Perhaps someone will step up at some point but I think we should disable it.

fangerer:
@mattip I can already reproduce the problem in a virtual machine using nearly the same system and versions as GitHub's docker images. Maybe I can quickly find the problem but as you said, I won't invest too much time.

PR #265, opened 2021-11-22
Re-enable cppcheck

fangerer:
#188

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -300,9 +300,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-    # - template: azure-templates/cppcheck.yml
-    # - template: azure-templates/python.yml
-    #
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -294,20 +294,25 @@ jobs:
           python test/check_py27_compat.py
 
 
-  # cpp_check:
-  #   runs-on: 'ubuntu-latest'
-  #   steps:
+  cpp_check:
+    name: Cppcheck static analysis
+    runs-on: 'ubuntu-latest'
+    steps:
+      - uses: actions/checkout@v2
 
-  #   # - template: azure-templates/cppcheck.yml
-  #   # - template: azure-templates/python.yml
+    # - template: azure-templates/cppcheck.yml
+    # - template: azure-templates/python.yml
+    #
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'
 
-  #   - name: Set up Python
-  #     uses: actions/setup-python@v2
-  #     with:
-  #       python-version: '3.x'
+      - name: Install Cppcheck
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -294,20 +294,25 @@ jobs:
...
+    name: Cppcheck static analysis
+    runs-on: 'ubuntu-latest'
+    steps:
+      - uses: actions/checkout@v2
 
-  #   # - template: azure-templates/cppcheck.yml
-  #   # - template: azure-templates/python.yml
+    # - template: azure-templates/cppcheck.yml
+    # - template: azure-templates/python.yml
+    #
```
hodgestar, line 305: These lines can be deleted now -- they were the recipes for installing cppcheck and python.

hodgestar (commented):
Thank you for bringing this back. The cppcheck is failing, but happy for this to be merged once it passes again.

fangerer:
> The cppcheck is failing, but happy for this to be merged once it passes again.

Interesting. I ran the action in my fork before I opened this PR and it passed. I'll investigate ...

.github/workflows/ci.yml
```diff
@@ -294,20 +294,25 @@ jobs:
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'
 
-  #   - name: Set up Python
-  #     uses: actions/setup-python@v2
-  #     with:
-  #       python-version: '3.x'
+      - name: Install Cppcheck
+        run: sudo apt-get -qq -y install cppcheck
```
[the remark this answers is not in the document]
fangerer: Thanks, I wasn't aware of that.
I did also have a look at the available GH Action plugins and one of them basically does nothing else than installing `cppcheck` via APT. I can have a look at other plugins to see how they install it.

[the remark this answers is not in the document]
hodgestar: I'm happy for us to keep the simple apt-get install for this.

hodgestar (approved):
Looks good!

hpy/devel/src/runtime/buildvalue.c
```diff
@@ -97,23 +97,24 @@ HPyAPI_HELPER
...
-            return HPy_Dup(ctx, result);
-        } else {
-            return result;
+            result = HPy_Dup(ctx, result);
         }
     } else {
-        return build_tuple(ctx, &fmt, &values, size, '\0');
+        result = build_tuple(ctx, &fmt, &values, size, '\0');
     }
+    va_end(values);
```
thedrow, line 116: This is the same problem as #73. We should run cppcheck in any version since it prevents these kind of errors which could escalate to security issues.

fangerer: Not sure if running cppcheck in _any_ version makes sense but we should at least run a decent version of it. It can also make sense to run several carefully picked versions.
However, the main problem currently is that `ubuntu-latest` in the CI provides a rather old version. We will need to manually download and build it. But that again imposes a risk since since external changes could then break our checks (any in my experience, this _will_ happen).
Anyway, please feel free to open PRs to improve the situation. I'd really appreciate that.

PR #276, opened 2022-01-12
HPy_LeavePythonExecution/HPy_ReenterPythonExecution

### LINKED ISSUE #34: API to acquire/release the GIL
Right now hpy doesn't have an API to acquire or release the GIL.
It'd be useful for C extensions to have one which acquires/releases the GIL when a GIL is used in a Python implementation and to make it a no-op when it is not.

steve-s:
This is supposed to be a 1:1 replacement for `Py_BEGIN_ALLOW_THREADS`/`Py_END_ALLOW_THREADS` when migrating CPython extensions.

Question left is if we should try to specify bit more generic GIL independent semantics like in this PR or have just `HPyEval_SaveThread` and `HPY_BEGIN_ALLOW_THREADS`. 

I could not yet come up with a concrete example where knowing that the C code will be in a section where it will not call any HPy API is beneficial other than GIL (maybe for GC?, in theory it can maybe scan stacks at that point). At the same time in the light of GIL removal efforts in CPython, it seems to me that we should not stick to obviously GIL specific APIs.

Fixes #34.

### DIFF
--- a/hpy/devel/include/hpy.h
+++ b/hpy/devel/include/hpy.h
@@ -110,6 +110,7 @@ typedef struct { intptr_t _i; } HPyField;
 typedef struct { intptr_t _lst; } HPyListBuilder;
 typedef struct { intptr_t _tup; } HPyTupleBuilder;
 typedef struct { intptr_t _i; } HPyTracker;
+typedef struct { intptr_t _i; } HPyThreadState;
 
 
 /* A null handle is officially defined as a handle whose _i is 0. This is true
--- a/test/test_00_basic.py
+++ b/test/test_00_basic.py
@@ -433,3 +433,24 @@ class TestBasic(HPyTest):
             @INIT
         """)
         assert mod.f(42) == 42
+
+    def test_leave_python(self):
+        mod = self.make_module("""
+            HPyDef_METH(f, "f", f_impl, HPyFunc_O)
+            static HPy f_impl(HPyContext *ctx, HPy self, HPy arg)
+            {
+                HPy_ssize_t data_len;
+                const char* data = HPyUnicode_AsUTF8AndSize(ctx, arg, &data_len);
+                HPy_ssize_t acount = 0;
+                HPy_BEGIN_LEAVE_PYTHON(ctx);
+                for (HPy_ssize_t i = 0; i < data_len; ++i) {
+                    if (data[i] == 'a')
+                        acount++;
+                }
+                HPy_END_LEAVE_PYTHON(ctx);
+                return HPyLong_FromSize_t(ctx, acount);
+            }
+            @EXPORT(f)
+            @INIT
+        """)
+        assert mod.f("abraka") == 3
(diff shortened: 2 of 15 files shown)


hpy/devel/include/hpy.h
```diff
@@ -140,6 +140,7 @@ typedef struct _HPyContext_s HPyContext;
 #include "hpy/runtime/helpers.h"
 
 #ifdef HPY_UNIVERSAL_ABI
+typedef struct { void* _data; } HPyThreadState;
```
steve-s, line 143: I think I went to far here, we should have this type defined the same for both CPython/Universal (here I have different typedef for CPython ABI aliasing this directly to `PyThreadState`) I'll fix that, but it otherwise does not change much from higher level view on this PR.

steve-s:
These API is needed for the matplotlib port, so it would be great to resolve this PR soonish. Would anyone mind taking a look? :-)

thedrow:
@steve-s I hope you don't mind me editing the original issue. This fixes the issue I opened :)

rlamy:
I guess this is fine for now. I feel we may have to revisit this when we implement more thread- and GIL-related APIs. I also find the names `HPy_BEGIN_LEAVE_PYTHON`/`HPy_END_LEAVE_PYTHON` a bit confusing, but I don't have a better suggestion.

qunaibit:
This PR is also needed for the `matplotlib` port.

thedrow:
In theory, after this feature is merged we can seriously consider PyO3 having a hpy backend.

hodgestar (approved):
I am not a threads / GIL expert, but this looks good to me, and I like that the API is somewhat independent of threads and the GIL and is just "I promise not to make API calls in this block".

Could we add some documentation in this PR about what exactly the HPy user is promising their code won't do inside the "LeavePythonExection" block? Not call any API function is a good start, but is it enough? E.g. one should probably also not mutate, move or free data that is managed by the interpreter or GC? And probably reading such data is quite dangerous too. Probably the best is "this code promises that it occupies a small island that is completely separate from the Python mainland".

steve-s:
> one should probably also not mutate, move or free data that is managed by the interpreter or GC? And probably reading such data is quite dangerous too. Probably the best is "this code promises that it occupies a small island that is completely separate from the Python mainland".

yes absolutely. 

I think that being able to access and even write into (except for `HPyFields`) what comes from `HPy_AsStruct(ctx,x)` (provided that it would be a valid pointer from HPy POV, i.e., handle `x` is still open), can be useful. For example, you may want to pass some data structures to some system library to do some long computations or to do a blocking call that would fill them in. Maybe the same for buffers?

There is a more general question regarding this and that is when the changes should be visible to other users that may have already retrieved a struct pointer for the same Python object or that will retrieve one while the first user is running writing some data into the pointer. One way is to say that they are visible immediately, the other is that they may be visible only once the corresponding handle is closed. In GIL situation it does not make a difference for the regions that hold GIL. We can specify this later to not derail this discussion.

> move or free data that is managed by the interpreter or GC

Can this be done somehow? Do you have some concrete examples in mind? I will update the docs.

steve-s:
I rebased the PR and pushed a fix for C++ compilation issue on Windows. I think the issue is broader (C++ does not support "Designated Initializers", so we must not use them in the headers), but in this PR I am intentionally fixing just was is necessary for it to pass the CI checks and will address that issue in a followup PR.

test/test_00_basic.py
```diff
@@ -433,3 +433,24 @@ def test_FromVoidP_AsVoidP(self):
...
+                    if (data[i] == 'a')
+                        acount++;
+                }
+                HPy_END_LEAVE_PYTHON(ctx);
+                return HPyLong_FromSize_t(ctx, acount);
+            }
+            @EXPORT(f)
+            @INIT
+        """)
+        assert mod.f("abraka") == 3
```
thedrow, line 456: This is an OK sanity test but we could do better.
We need to check that we're not crashing if another thread mutates the variable which stores the string "abraka" and then we should run the test multiple times to try to trigger a crash.
Should I open an issue about it and we'll get to it later?

steve-s: yes, good suggestion

PR #277, opened 2022-01-12
Introduce pytest mark for tests relying on sync GC.

fangerer:
As discussed in PR #260, this introduces a custom pytest mark for tests assuming that the GC runs synchronously.


rlamy (approved):
LGTM, the failures are unrelated to this PR.

PR #278, opened 2022-01-18
Consistently return HPy_NULL from all API functions that set exception.

steve-s:
Some functions that set exception were returning `HPy_NULL` some were not. This PR makes it consistent.

I pushed the `return HPy_NULL` to the generated trampoline functions, so the context functions are all `void` and the Python engine does not get to choose what this API to returns.

Instead of hand-writing the trampolines, I went down the rabbit hole of adding support for this into autogen and as part of that I moved all the parameters of code generation into one new Python file `config.py`.


hpy/tools/autogen/conf.py
```diff
@@ -0,0 +1,111 @@
+# Defines parameters for the code generation
```
steve-s, line 1: These constants were pulled from other files into this central location.

steve-s:
I think this is hopefully not very controversial. Could someone have a quick look at the refactoring in code generation please, @rlamy or @antocuni? I'd like to merge this to avoid conflicts.

hodgestar:
@steve-s What is the advantage of always returning HPy_NULL (regardless of whether an exception is set) rather than having no return? In either case one has to call HPyErr_Occurred to check. Maybe I am missing something?

steve-s:
@hodgestar this is for functions that are known to always set exception and the caller usually calls the API function and then returns `HPy_NULL` so code like this:
```
HPyErr_SetString(ctx, ctx->h_RuntimeError, "message");
return HPy_NULL;
```
can be
```
return HPyErr_SetString(ctx, ctx->h_RuntimeError, "message");
```

It's a minor convenience for the user. CPython has it the same way, so I think it's useful to match that "feature".

hodgestar:
Ah, I see -- it's about making porting easier.

As an interface it seems a bit of a mess -- e.g. what happens if one calls `HPyErr_SetString` from a function that returns an int or float or something? But the portability argument is a good one. I'm not 100% sure which is the better trade-off. Typing "return HPy_NULL" explicitly is not hard and something of a win, but it is slightly annoying to have to do everywhere and we could avoid it.

Any thoughts from anyone else?

hpy/tools/autogen/conf.py
thedrow: This might not be necessary right now but maybe the configuration could use a schema. This will make things easier as this configuration file will grow in size and options.

In the meantime, please document how to use this configuration file in a docstring.

hpy/tools/autogen/ctx.py
```diff
@@ -40,6 +41,9 @@ def declare_func(self, func):
         # turn the function declaration into a function POINTER declaration
         newnode = deepcopy(func.node)
         newnode.type = c_ast.PtrDecl(type=newnode.type, quals=[])
+        if conf.RETURN_CONSTANT.get(func.node.name):
+            make_void(newnode)
```
thedrow, line 45: Please extract this to a helper function as the code repeats itself (in more than one variation).

hpy/tools/autogen/debug.py
```diff
@@ -88,8 +91,11 @@ def gen_debug_wrapper(self, func):
         #
         assert not func.is_varargs()
         node = get_debug_wrapper_node(func)
+        const_return = conf.RETURN_CONSTANT.get(func.node.name)
+        if const_return:
+            make_void(node)
         signature = toC(node)
-        rettype = toC(node.type.type)
+        rettype = 'void' if const_return else toC(node.type.type)
```
thedrow, line 98: Please use `get_context_return_type()`.

thedrow (changes requested):
Overall, this looks good and I like the idea of having one configuration file. I have a few notes I'd like you to address first before reviewing this again.

hpy/tools/autogen/debug.py
steve-s: `get_context_return_type` used to take the function object (and reads its `node.node.type`), but in this case we need to use the new `node` created at line 93. My take was that it's better to spare the callers of `get_context_return_type` from passing `func.node` instead of `func` and have with this one special case. I've generalized `get_context_return_type` so that it can be now used also here.

hpy/tools/autogen/ctx.py
steve-s: I assume you mean the `if` condition? Extracted to `get_return_constant`. Introducing some new `make_void_if_returns_constant(func, func_node)` seemed to me as not much more concise/self-explanatory than the plain `if`.

hpy/tools/autogen/conf.py
steve-s: > In the meantime, please document how to use this configuration file in a docstring.

Do you have suggestion what else to comment on? I've added a comment to each constant in this file.

steve-s:
> Overall, this looks good and I like the idea of having one configuration file. I have a few notes I'd like you to address first before reviewing this again.

Thank you for the review @thedrow! I've addressed some of the issues raised in 9f5f9c906d79aaceab9881a4f0358a0db0b40f9f and commented on the others.

rlamy:
I'm a bit unsure about this, but consistency is important, both internally and with CPython, so I'd say either all functions should return HPy_NULL or they should return void. Now the `return PyErr_XXX` pattern is used quite a lot, especially with `PyErr_NoMemory()`:
```
(numpy-rel) ronan@Izenah:~/devel/top4000-pypi-packages$ grep -R "return PyErr" top4000 | wc -l
1174
(numpy-rel) ronan@Izenah:~/devel/top4000-pypi-packages$ grep -R "return PyErr_NoMemory" top4000 | wc -l
404
```
So the HPy_NULL option does seem to make porting easier. +1.

steve-s:
Note: rebased on the master

steve-s:
Also note that with the current implementation the returning of `HPy_NULL` is (intentionally) hardcoded in the context trampoline so in that regard it can be seen as a HPy helper. The `HPy_NULL`, in this case, does not cross the ABI boundary. Context functions are still `void`.

steve-s:
> I'm a bit unsure about this, but consistency is important, both internally and with CPython, so I'd say either all functions should return HPy_NULL or they should return void. Now the `return PyErr_XXX` pattern is used quite a lot, especially with `PyErr_NoMemory()`:
> 
> ```
> (numpy-rel) ronan@Izenah:~/devel/top4000-pypi-packages$ grep -R "return PyErr" top4000 | wc -l
> 1174
> (numpy-rel) ronan@Izenah:~/devel/top4000-pypi-packages$ grep -R "return PyErr_NoMemory" top4000 | wc -l
> 404
> ```
> 
> So the HPy_NULL option does seem to make porting easier. +1.

Thanks for the analysis. I think it's undebatable that the API must be consistent and this analysis IMO shows that returning `HPy_NULL` makes more sense.

I'd like to make progress with this PR. Any objections from the others (@antocuni, @hodgestar)? @thedrow could you please take a look at my changes.

hpy/tools/autogen/ctx.py
thedrow: You can call it `maybe_make_void` and it would be shorter.

hpy/tools/autogen/conf.py
thedrow: Since these are dictionaries which are untyped we need to document the schema.
We could use https://docs.python.org/3/library/typing.html#typing.TypedDict or dataclasses instead of dictionaries.

thedrow: We should also include documentation regarding this in the contribution guide.

hpy/tools/autogen/ctx.py
steve-s: good suggestion, done

hpy/tools/autogen/conf.py
steve-s: It seems that we do not have any extensive contribution guide or documentation? No documentation file in the repo mentions `make autogen`, the closest is https://github.com/hpyproject/hpy/blob/master/CONTRIBUTING.md but it is very basic. 

I agree that adding more extensive developer oriented documentation would be good, but I think it is out of the scope of this PR. The same applies to the suggestion to use schema. Currently in the master the "configuration" is scattered over different files, so putting it in one file is IMO improvement already. We can create a followup issue for improving this even further.

steve-s:
Addressed one more refactoring suggestion and rebased on the current master. I think this may be ready to merge. Are there any remaining objections?

rlamy:
I think we can merge this now.

PR #279, opened 2022-01-18
Add HPy_Contains as way to call the 'in' operator from HPy

steve-s:
I need `PySequence_Contains` for porting psutil on MacOS, but it seems to me that calling it just `HPy_Contains` would be more in line with `HPy_GetItem` and friends and with the notes from `leysin-2020-design-decisions.md`:

```
Question: Should we have separate HPySequence_GetItem and HPyMapping_GetItem
or just one HPy_GetItem? Should we explore the idea of protocols for such
access.

Decisions:

* We should start with `HPy_GetItem` and `HPy_SetItem` and `HPy_GetItem_i` and
  `HPy_SetItem_i` for item access and `HPy_GetAttr`, `HPy_SetAttr`,
  `HPy_GetAttr_s` and `HPy_SetAttr_s` for attribute access..
```


fangerer (approved):
Since I need that as well ...

hodgestar (approved):

PR #280, opened 2022-01-18
Add HPyErr_WarnEx and HPyUnicode_DecodeFSDefaultAndSize

steve-s:
Simply adds those new API functions that straightforwardly delegate to CPython APIs.


steve-s:
Hopefully not very contentious PR, so I am going to merge now.

PR #281, opened 2022-01-18
Add HPy_CLEAR convenience macro

steve-s:


hpy/devel/include/hpy/macros.h
```diff
@@ -68,4 +68,24 @@ typedef enum {
...
+ *
+ * Note: the issues solved by Py_CLEAR on CPython do not apply to HPy. Since
+ * handles should be always short-lived and never shared between threads or
+ * different calls to HPy module entry points, arbitrary code triggered
+ * from __del__ (which can be triggered from HPy_Close) cannot access the same
+ * stale handle value.
+ *
+ * What can be shared are HPy_Fields, but there the
+ * ``HPyField_Store(ctx, obj_h, &obj->f, HPy_NULL)`` should internally do the
+ * right thing.
```
steve-s, line 83: I assume @antocuni?

hodgestar:
@steve-s What is the background for wanting to add this? I get that it's neat to set a handle to NULL after it's closed, but if we felt this should be done all the time we could just do it in HPy_Close? I'm not saying we shouldn't have this helper -- I'm just not quite understanding the use case.

steve-s:
@hodgestar technically it doesn't gain much, but I think it may serve as kind of a documentation and C API porting "hint". 

AFAIC `Py_CLEAR` has two important implications in CPython. One does apply to HPy (setting the variable to `HPy_NULL`) and one doesn't (possibility of seeing a stale PyObject pointer) and I think that if you are porting code that uses `Py_CLEAR`, it naturally guides you to use `HPy_CLEAR`, which preserves the one applicable implication, but also (hopefully) shouts at you from its documentation that the other thing does not apply for HPy.

More concretely: this is a pattern that I saw in psutils:
```
PyObject *a = NULL, *b = NULL, *c = NULL;
for (...) {
   a = PyDict_New();
   if (a == NULL) goto error;
   b = PyDict_New();
   if (b == NULL) goto cleanup;
   // ... actually do something useful
   Py_CLOSE(a);
   Py_CLOSE(b);
}
// ...
cleaup:
   Py_XDECREF(a); // may be NULL, but never already decref'd object
   Py_XDECREF(b); // may be NULL, but never already decref'd object
```
I am myself guilty of mechanically replacing `Py_CLOSE` with `HPy_Close` and only after a moment I stopped to check what `Py_CLOSE` does exactly and if its semantic difference from `Py_DECREF` could be important for this code.

rlamy:
I'm not sure about this, it seems significantly more common to use `Py_CLEAR(foo->bar);` than `Py_CLEAR(foo);`, and the former needs to be replaced by `HPyField_Store(..., HPy_NULL);`. The latter form seems to only be used for making cleanup easier, and must be paired with a `Py_XDECREF(foo)`, so if we add `HPy_CLEAR`, we should also add an `HPy_XClose`, to support both parts of the pattern.

steve-s:
@rlamy you have a good point that convinced me. We should instead document this in the porting guide. 

Note that `HPy_Close` is already what would be `HPy_XClose`, but I don't think that it's an issue? The fact that it handles `HPy_NULL` properly is a nice addition and I assume that the difference of `Py_DECREF` vs `Py_XDECREF` is motivated by the possibility of tiny performance win when one omits the `NULL` check, but in HPy case both will be function calls (at least in universal mode), so the `HPy_NULL` check is the lest thing to worry about.

rlamy:
I was wondering where I got the idea that you can't call HPy_Close() on HPy_NULL, but so far it's actually just a CPython (and GraalPython?) implementation detail, and it doesn't work on PyPy. We could make it official, but it would need to be tested.

steve-s:
Right. I am guilty of inferring the contract from the implementation. We should agree on one or the other and document it and enforce it in debug mode. Maybe let's have an informal pool at the next meeting.

Are there any reasons for having a variant that does not expect `HPy_NULL` other than performance? It can be documentation of the intent: "this variable should not be `HPy_NULL` here", but that can be done with an assert too. Assert would be actually better for documentation/debugging purposes.

fangerer:
I'm pretty sure that it will be very hard to show any performance impact due to the additional null check. So, I would prefer that `HPy_Close` also accepts `HPy_NULL` and make it official.
If someone wants to assert that a handle is not null, I think it should be done explicitly.

Issue #174, opened 2021-02-25
Use function pointers (and not `void *`) for pointers to functions.


hodgestar:
The CPython API uses `void *` for function pointers. However, in strict standard C, these are not the same things. If would be nice to resolve this issue in HPy.

Practical considerations:

* All the compilers we currently run integration tests against (gcc, clang, soon MSC) are fine with the difference.
* On some architectures data and instruction pointers are not even the same since.
* This also creates difficulties for Rust and WebAssembly, where the distinction between data and instruction pointers can be important.

See https://github.com/encukou/abi3/issues/14 for the same issue on CPython.

See #23 for an initial attempt at making this change in HPy.

fangerer:
@qunaibit I think you already fixed that issue, right? Do you think we can close it?

qunaibit:
Yes

fangerer:
cool, thank you

PR #283, opened 2022-01-31
[gh-174] Add C++ support

qunaibit:
In the effort of porting `matplotlib` library, which is heavily implemented in C++, `void *` pointers need to be replaced with function pointers in addition to other adjustments.

I'v add C++ proof-of-concept to the current `./proof-of-concept` as a test case for the C++ support.


thedrow (changes requested):
We need infer to verify the code when compiling using C++ as well.

qunaibit:
The verification is in the `C++ proof-of-concept` and it's checked as part of the `
tests / Proof of concept tests`

fangerer:
@thedrow I'm not sure if I understand. Do you mean we should run `infer` on the C++ proof-of-concept?

thedrow:
> @thedrow I'm not sure if I understand. Do you mean we should run `infer` on the C++ proof-of-concept?

No. What you need to do is build the code using a C++ compiler and let infer inspect that.
[Here](https://github.com/hpyproject/hpy/blob/master/Makefile#L48) we build the extension using a C compiler.
We should verify that there's no difference at the very least because C++ has different rules than C.

qunaibit:
> What you need to do is build the code using a C++ compiler

Unfortunately, doing so raise couple of errors for `hpy/debug/src/_debugmod.c`:
```
error: taking address of temporary array
error: jump to label ‘error’
```

> verify that there's no difference at the very least

HPy itself will always be compiled using `C` compiler. The modifications are mainly for header files that will be included in users' C++ extensions.

qunaibit:
> we need to somehow verify that they export all the symbols correctly

HPy's symbols won't be affect by the target as it is self contained and compiled without mangling.

> we're not breaking any linter rules on that target.

linter rules for HPy can be a great addition, but out-of-scope of this PR as linter can also be applied to `C`-based extensions.

> If you're satisfied with the way things currently are

I believe so.

Thanks for the comments and suggestions

rlamy:
LGTM. Having to explicitly cast all the `void*` is a bit annoying though. Is there any way to avoid it?

qunaibit:
> Having to explicitly cast all the void* is a bit annoying though. Is there any way to avoid it?

Unfortunately, no. C++ preprocessor is very strict with function pointers.
I've tried to hide those changes from user's extension, so this should not affect any prior c ext implementations.

rlamy:
OK, we'll have to live with it, then. I'll merge.

PR #284, opened 2022-01-31
Support parsing nested tuples in HPyArg_Parse

qunaibit:
In the effort of porting `matplotlib` library, there are multiple places where `()` are being used in argparse formats.

### DIFF
--- a/hpy/devel/src/runtime/argparse.c
+++ b/hpy/devel/src/runtime/argparse.c
@@ -423,8 +423,8 @@ parse_tuple_items(HPyContext *ctx,
     for (HPy_ssize_t i = 0; i < tuple_length && fmt != fmt_end; i++) {
         if (*(*fmt) == 'O' && ht == NULL) {
             set_error(ctx, ctx->h_SystemError, err_fmt,
-                "parsing tuples cannot be used unless"
-                " an HPyTracker is provided. Please supply an HPyTracker.");
+                "parsing tuples cannot be used unless an HPyTracker is provided. "
+                "Please supply an HPyTracker.");
             return 0;
         }
         current_arg = HPy_GetItem_i(ctx, tuple, i);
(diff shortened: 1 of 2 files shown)


steve-s:
The 'O' formats in tuples/lists should have tests otherwise lgtm

rlamy (changes requested):
Looks good overall. I'm a bit unsure about the fatal errors actually. I know CPython does it that way, but wouldn't it be better to raise SystemError instead?

hodgestar:
I would like to put in a vote for *not* parsing nest tuples, for the following reasons:

1. I think this exists in Python to provide a C alternative to tuple unpacking in Python function signatures. That feature caused all sorts of issues and no longer exists.
2. The nested bracketing makes the parsing a lot more complicated to reason about.
3. I'm not really sure what the advantage of it is over just unpacking the tuple into a handle and then calling the necessary functions to extract the tuple elements individually (e.g. `HPy_GetItem_i` or `HPy_GetItem_s`).

Perhaps there really are lots of uses out there that are hard to port though?

qunaibit:
> That feature caused all sorts of issues and no longer exists.

It still exist and very much in use as per CPython doc: "[Format units for sequences may be nested.](https://docs.python.org/3/c-api/arg.html#c.PyArg_ParseTuple)"

The other two reasons have a point. As you said this can be replace with exiting APIs.

This PR is just to ease the porting effort, and it is not essential.

hodgestar:
> > That feature caused all sorts of issues and no longer exists.
> 
> It still exist and very much in use as per CPython doc: "[Format units for sequences may be nested.](https://docs.python.org/3/c-api/arg.html#c.PyArg_ParseTuple)"

I didn't mean `PyArg_Parse`, I meant `def f((x, y)): ...` which existed in Python 2.

qunaibit:
That's not part of this PR. This PR just reuse existing mechanics of HPy's 'argparse'

hodgestar:
@qunaibit Would you mind trying to understand my point? Perhaps I am not explaining it well, but you are replying with unrelated rebuttals instead of asking for clarification. The point I am trying to make is that `def f((x, y)):` is no longer valid Python syntax and parsing nested tuples in `PyArg_ParseTuple` is the equivalent C functionality, so although the C functionality still exists, it is less useful than it was.

hodgestar:
I didn't intend to end the conversation. If there are a lot of use cases or porting without it is hard or performance is bad, we could definitely add it. I am just pushing back a little on adding all the features of the existing C API.

qunaibit:
In top 4000: there are 35 usage of this pattern within 19 files. It's not a lot.

hodgestar:
Thank you for checking!

PR #285, opened 2022-02-01
 Fix CPython 3.8 and 3.9 debug builds on CI

### LINKED ISSUE #282: Fix CPython 3.8 and 3.9 debug builds on CI
The issue is probably caused by https://github.com/pypa/setuptools/issues/2946#issuecomment-1003118426

steve-s:
Fixes #282

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,6 +60,9 @@ jobs:
         os: [ubuntu-latest]
         python-version: ['3.9', '3.8', '3.7']
 
+    env:
+      SETUPTOOLS_USE_DISTUTILS: stdlib
+
     steps:
       - uses: actions/checkout@v2
 


hodgestar:
What was the thing that broke?

steve-s:
@hodgestar my understanding is that setuptools started using its own copy of distutils, which however relies on some CPython internal or semi public API that somehow does not work with the debug build. The distutils within CPython works. We'll probably have to revert this once they remove distutils from CPython for good, but then setuptools with its internl distutils copy should be hopefully working again...

steve-s:
@hodgestar Ah I see I forgot to link this with the corresponding issue.

steve-s:
Since I hope this is not very controversial and it re-enables CI checks that we've been ignoring for a while, which is not ideal, I'll go ahead with merging.

hodgestar:
@steve-s Thank you for the explanation & thank you for fixing.

PR #287, opened 2022-02-07
Add pyproject.toml to support PEP 518 and isolated builds.

domdfcoding:
This allows tools like pip to install build dependencies when building `hpy` from source. Currently, building without the `wheel` package installed fails. With this change pip will install the necessary dependencies in an isolated environment.

See https://www.python.org/dev/peps/pep-0518/ for more information.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -155,5 +155,4 @@ setup(
         ],
     },
     use_scm_version=get_scm_config,
-    setup_requires=['setuptools_scm'],
 )
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,3 @@
+[build-system]
+requires = [ "setuptools>=40.6.0", "setuptools-scm[toml]>=6.0", "wheel>=0.34.2",]
+build-backend = "setuptools.build_meta"


steve-s:
Hello Dominic,

do you know why the build for infer could be failing?

```
python3 setup.py build_ext -if -U NDEBUG | compiledb
/opt/hostedtoolcache/Python/3.10.2/x64/lib/python3.10/distutils/dist.py:274: UserWarning: Unknown distribution option: 'use_scm_version'
  warnings.warn(msg)
hpy/universal/src/hpymodule.c:13:10: fatal error: hpy/version.h: No such file or directory
   13 | #include "hpy/version.h"
      |          ^~~~~~~~~~~~~~~
compilation terminated.
error: command '/usr/bin/gcc' failed with exit code 1
```

I wonder if the other targets from the `Makefile`, which we use for development, will continue working? For example:
```
.PHONY: all
all: hpy.universal

.PHONY: hpy.universal
hpy.universal:
        python3 setup.py build_ext -if

.PHONY: dist-info
dist-info:
        python3 setup.py dist_info
```

setup.py
```diff
@@ -155,5 +155,4 @@ def get_scm_config():
         ],
     },
     use_scm_version=get_scm_config,
-    setup_requires=['setuptools_scm'],
```
steve-s, line 158: Ah, I guess we'll now have to install those manually when using plain `setup.py` based installation? We can either change the targets in the Makefile to invoke something we now need or update the CI scripts to install the necessary deps before invoking the Makefile target in the (for the infer CI job, other jobs install via `pip install .`)

hodgestar: The pip equivalent to `python setup.py develop` is a bit of a pain -- something like `pip install <write-build-requirements-by-hand-here>` followed by `pip install --no-build-isolation -e .`

I'm not sure it actually hurts to duplicate `setuptools_scm` in `setup_requires`. Maybe it would be fine to leave it in?

steve-s: So it looks like we just need to revert the changes in [setup.py](https://github.com/hpyproject/hpy/pull/287/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7) and we can merge this. @domdfcoding, would you like to do that? Otherwise we can also take this over if you want.

steve-s:
@domdfcoding, thank you for your contribution. I took the liberty to add a commit to your PR that fixes the failing job, rebase it on master and merge it here: https://github.com/hpyproject/hpy/pull/306.

PR #291, opened 2022-02-15
Add support for more HPyUnicode_* functions

qunaibit:
Required in an effort to port `matplotlib`


rlamy (approved):

TeamSpen210:
Trawling through Git history, `Py_UCS4` has always been a 32-bit unsigned integer. It was [changed to use stdint](https://github.com/python/cpython/commit/a13e367778d64b321bce8e6b39f9e679b24b9afe) in 3.6, but before then it [simply typedefs to `unsigned int`/`unsigned long`](https://github.com/python/cpython/commit/1294ad0c59d4a8bfbdd3f4606653d7aa20b2969e#diff-c0f9841e1752b5b8cd245fb5980c0efadf5bef986262d6a38ed33b5d74f57cfdR69), and has been like that since 2001 (Python 2.2+). So it entirely exists as an alias for readability.

PR #295, opened 2022-02-22
Add support for HPyLong_(AsVoidPtr, AsDouble) & HPyErr_WriteUnraisable

qunaibit:
Required for porting `matplotlib` and `kiwi`


test/test_hpyerr.py
```diff
@@ -662,3 +662,23 @@ def do_not_raise_exception(*args):
...
+            @INIT
+        """)
+        if not SUPPORTS_SYS_EXECUTABLE:
+            # if sys.executable is not available (e.g. inside pypy app-level)
+            # tests, then skip the rest of this test
+            return
+        # subprocess is not importable in pypy app-level tests
+        result = python_subprocess.run(mod, "mod.f()")
+        assert result.returncode == 0
+
```
antocuni, line 684: is it really needed to start a subprocess for this test? You could set `sys.unraisablehook` to something which records it has been invoked, then call `mod.f()` and check that the hook has been actually raised, unless I miss something

antocuni (commented):
LGTM, apart the small comment about the test

test/test_hpyerr.py
qunaibit: I didn't know about that.. nice suggestion. Thank you
I noticed that there is support for such routine in `test.support import catch_unraisable_exception` but this has only been added on the `3.8.12` release.. I added it manually until `hpy` drop support for older python releases

antocuni: ah ok, I didn't think about old pythons. +1 for subprocess then

qunaibit: approved for merge?

steve-s (approved):

PR #299, opened 2022-03-02
Introduce HPyGlobal

steve-s:
`HPyGlobal` can be alternative or replacement of the module state. Why I think `HPyGlobal` makes more sense for HPy?

- we have to pass the context around already, passing the module object around too seems redundant
- since the interpretation the opaque `HPyGlobal` "handle" is determined by the context, we can even have two implementations for CPython: one where `HPyGlobal` == `PyObject*` and other that uses some indirection via interpreter state that would support subinterpreters.
- together with some support for "get me the current context" this can make incremental porting even easier

I think the module state could be still useful for HPy to ease porting of code that already uses module state. OTOH, code that uses lots of globals, like numpy, for example, would be much easier to port with `HPyGlobal` and it seems to me that there is technically not much drawback to `HPyGlobal` over module state.


thedrow (changes requested):
What about thread safety?
I didn't read through the code yet but since this is a concern you didn't address in your PR description, I think that's something that is important to think about.

fangerer:
@thedrow Thread safety and/or sub-interpreters are still the big question. We had some discussion about `HPyGlobal´ in the last dev call (see also https://github.com/hpyproject/hpy/wiki/dev-call-20220303#hpyglobal ).
The rough idea is that each C global variable (the `HPyGlobal`) will be initialized by the first thread/interpreter that uses it. This would be the allocation of the HPy global. For that, we probably need to make the allocation part of HPy's ABI. We did not yet come up with a full solution for that. We were thinking of a global counter variable (a `uint64_t` or such) and the `HPyGlobal` as well as the counter variable are mutated with atomic operations. We were also thinking of a contract that the initialization of the HPy globals (i.e. associating a Python object with the HPy global) must happen in the module init function.

steve-s:
Originally my take was that thread safety is handled by the concrete HPy implementation depending on what it actually stores in `HPyGlobal` and whether or not it uses GIL. 

What @fangerer described could be a solution if we wanted to support embedding of multiple HPy implementations within one process, because then they need to agree on how to interpret and handle the data inside `HPyGlobal`. In such case thread safety would have to be part of that contract. It would probably have to atomically do this: increment the global counter and store it in the `HPyGlobal`.

However, I think we agreed on the call that embedding multiple HPy implementations in one process was not a (important) goal for HPy, so I think we agreed on leaving this up to the HPy implementation with the drawback that we cannot embed two implementations into one process unless they are compatible with each other.

steve-s:
I've made the `HPyGlobal`s registration part of module creation. This forces the users to use only statically initialized globals for HPyGlobal (unless they dynamically allocate the module definition itself). So I think it is good usability improvement, although the current implementation does not need it.

Example:
```
HPyGlobal myglobal;

    // ...
    .globals = {
        &myglobal;
    },
};
```

everything else as before. What is still the same: one cannot embed two incompatible HPy implementations within one process if any module is using `HPyGlobal`. I think we can make it work to mix debug and non-debug contexts, but mixing CPython and, say, GraalPython contexts will not work unless the two agree on some format of the internal data within HPyGlobal. Specification of such format as a part of HPy ABI can be added later.

thedrow (changes requested):
I'm fine with merging this as long as thread safety concerns have been properly documented.

thedrow (approved):

PR #300, opened 2022-03-07
Enable and test slots HPy_mp_*

fangerer:
This PR also fixes a small bug in the debug implementation of `HPyUnicode_AsUTF8AndSize` (function `debug_ctx_Unicode_AsUTF8AndSize`) which did not accept a `NULL` pointer for the out param `size`.


hodgestar (approved):
This looks good to me.

PR #301, opened 2022-03-10
Small cleanups: fix compiler warnings, add -Werror, use the same approach as _hconv for HPyThreadState

steve-s:


hodgestar (approved):
Looks sane to me, but it appears the Windows CI robots are unhappy.

Issue #252, opened 2021-10-20
Port creation and destruction of ndarray objects to HPy


rlamy:
The next step for #137 is to handle the lifecycle of ndarray objects in HPy. Concretely, this means:
* Convert the following legacy slots to their HPy equivalents: Py_tp_dealloc, Py_tp_alloc, Py_tp_new, Py_tp_free
* Replace all `PyObject*` members in the `PyArrayObject` struct with `HPyField`
* Ensure that all the object references held by the ndarray can be traced by the GC

rlamy:
Thanks for the link, that will be useful. The `HPyField` mechanism does indeed require implementing cyclic GC support on CPython. However, the API is different since HPy attempts to abstract over the differences between CPython and non-refcounted implementations, so the C code can't be reused directly. OTOH, passing the tests from that PR is clearly a goal.

rlamy:
@thedrow Sorry for the late answer, I had to attend to urgent family matters this week. As I've started working on this, I expected to finish it on my own, and I don't think it can be split easily.

rlamy:
Dealing with arrays that contain objects, and particularly [structured arrays](https://numpy.org/doc/stable/user/basics.rec.html), raises some issues with HPyFields:

* To handle unaligned data, code like this needs to work:
```C
void setitem(HPyContext *ctx, char *data, HPy item) {
    HPyField f;
    memcpy(&f, data, sizeof(f));
    HPyField_Store(ctx, owner, &f, item);
    memcpy(data, &f, sizeof(f));
    return;
}
```
* Reading or writing to a field requires knowing the owning array. Many internal numpy functions pass around only the data pointer, which makes them hard to port - it's sometimes even unclear who the owner is.
* In some cases, upstream numpy copies a block of memory containing PyObjects and, later on, updates the refcounts by walking the copy. Emulating that with HPy requires code like:
```C
void fix_field_owner(HPyContext *ctx, HPy src_array, HPy dst_array, HPyField *f) {
    HPy item = HPy_Load(ctx, src_array, *f);
    *f = HPyField_NULL;  // break identification between source field and its copy 
    HPyField_Store(ctx, dst_array, f, item); // INCREFs item on CPython
    HPy_Close(ctx, item);
    return;
}
```
[...]

PR #302, opened 2022-03-10
Fix possibility to see stale objects in tp_finalize due to an issue in HPyField_Store

steve-s:
Already discussed in https://github.com/hpyproject/hpy/issues/252#issuecomment-1020989343 with @rlamy. I believe this is related to the reasons why CPython has the `Py_CLEAR` macro. The test could be also seen as a specification of HPy contract w.r.t. `HPyField_Store` and `tp_finalize` interaction.


steve-s:
Given that we agreed with @rlamy on the implementation already in https://github.com/hpyproject/hpy/issues/252#issuecomment-1020989343, the only thing left to review is the test, which I think is not such an issue to revert or fix ex-post, so I'll take the liberty to merge this.

PR #303, opened 2022-03-10
Debug mode: detect HPy_Close/return of context constants

steve-s:
Tags handles that are used as context constants (e.g., `h_None`) and adds checks that guards against `HPy_Close(ctx, ctx->h_None)` or against `return ctx->h_None`.


hpy/debug/src/debug_handles.c
```diff
@@ -75,7 +85,7 @@ void DHPy_invalid_handle(HPyContext *dctx, DHPy dh)
 {
     HPyDebugInfo *info = get_info(dctx);
     HPyContext *uctx = info->uctx;
-    assert(as_DebugHandle(dh)->is_closed);
+    assert(as_DebugHandle(dh)->is_closed || as_DebugHandle(dh)->is_context_constant);
     if (HPy_IsNull(info->uh_on_invalid_handle)) {
         // default behavior: print an error and abort
         HPy_FatalError(uctx, "Invalid usage of already closed handle");
```
steve-s, line 91: The error messages could be more specific (also for other scenarios) and we could pass some ID of the error to the user provided hook, but I am leaving that for followup work

steve-s:
Note: rebased on master, it seems the infer error was fixed on master meanwhile...

hpy/debug/src/_debugmod.c
```diff
@@ -15,6 +15,8 @@ HPyDef_METH(new_generation, "new_generation", new_generation_impl, HPyFunc_NOARG
 static UHPy new_generation_impl(HPyContext *uctx, UHPy self)
 {
     HPyContext *dctx = hpy_debug_get_ctx(uctx);
+    if (dctx == NULL)
```
steve-s, line 18: These checks were added after error report from infer. https://github.com/hpyproject/hpy/actions/runs/1968620957/jobs/2813414188

steve-s:
Rebased on master. @antocuni can we take a look at reviewing this? :-)

hpy/debug/src/_debugmod.c
steve-s: Maybe we should really abort in `hpy_debug_ctx_get` instead of returning `HPy_NULL`:
```HPyContext * hpy_debug_get_ctx(HPyContext *uctx)
{
    HPyContext *dctx = &g_debug_ctx;
    if (uctx == dctx) {
        HPy_FatalError(uctx, "hpy_debug_get_ctx: expected an universal ctx, "
                             "got a debug ctx");
    }
    if (hpy_debug_ctx_init(dctx, uctx) < 0) {
        HPyErr_SetString(uctx, uctx->h_SystemError, "Could not create debug context");
        return NULL;
    }
```

hpy/tools/autogen/debug.py
```diff
@@ -47,7 +47,7 @@ def generate(self):
         w('{')
         for var in self.api.variables:
             name = var.name
-            w(f'    dctx->{name} = DHPy_open(dctx, uctx->{name});')
+            w(f'    dctx->{name} = DHPy_open_context_constant(dctx, uctx->{name});')
```
antocuni, line 50: I really like the idea, but `context_constant` is a bit weird IMHO. What about `DHPy_open_immortal`?

antocuni (changes requested):
Brilliant idea!
I vote for calling them "immortal" handles, but apart from this, very good job :)

steve-s:
Rebased on master and changed the name to "immortal"

fangerer (approved):

PR #304, opened 2022-03-11
CPython 3.10 support

steve-s:
I kept only one flag in `tp_flags`. It's not ideal solution, but better than no 3.10 support :-)


auvipy (approved):

steve-s:
I've changed the bit used once more to avoid issues in 3.11a (there are however other problems in 3.11)

PR #309, opened 2022-03-24
Enable exact type checks with HPy_TypeCheck

fangerer:
In CPython, there are usually macros like:
#define PySomething_Check(op) PyObject_TypeCheck(op, &PySomething_Type)
#define PySomething_CheckExact(op) (Py_TYPE(op) == &PySomething_Type)
That translates to following in HPy:
// PySomething_Check
HPy_TypeCheck(ctx, h_obj, h_type)

// PySomething_CheckExact(op)
HPy_Is(ctx, HPy_Type(ctx, h_obj), h_type)
As it turns out, the exact type check is pretty verbose in HPy and having two API calls may also be a high performance drawback.

An alternative approach would be to have a new API for that. Something like `HPy_TypeCheckExact` (which would come with a new context function as well).

### DIFF
--- a/hpy/debug/src/_debugmod.c
+++ b/hpy/debug/src/_debugmod.c
@@ -230,7 +230,7 @@ HPyDef_SLOT(DebugHandle_cmp, DebugHandle_cmp_impl, HPy_tp_richcompare)
 static UHPy DebugHandle_cmp_impl(HPyContext *uctx, UHPy self, UHPy o, HPy_RichCmpOp op)
 {
     UHPy T = HPy_Type(uctx, self);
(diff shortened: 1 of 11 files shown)


steve-s:
Would it make sense to have `HPy_TypeCheck` and `HPy_TypeCheckExact` in the API, but keep the variant with the `exact` flag as a parameter as a context function to avoid adding another context function? From readability POV the third argument in snippets such as `HPy_TypeCheck(a, b, 0)` or `HPy_TypeCheck(a, b, false)` is not great, because it is not clear what it means unless you look at `HPy_TypeCheck` signature.

fangerer:
Right, let's add an inline helpers to improve readability.

PR #315, opened 2022-04-19
New test: calling TupleBuilder_Cancel after TupleBuilder_Build

steve-s:
I'd like to add this test to make the contract documented. We can also choose to make a different contract and update the test. No strong opinion.

### DIFF
--- a/test/test_hpytuple.py
+++ b/test/test_hpytuple.py
@@ -79,3 +79,19 @@ class TestTuple(HPyTest):
             @INIT
         """)
         assert mod.f("xy") == ("xy", True, -42)
+
+    def test_TupleBuilder_cancel_finished(self):
+        mod = self.make_module("""
+            HPyDef_METH(f, "f", f_impl, HPyFunc_O)
+            static HPy f_impl(HPyContext *ctx, HPy h_self, HPy h_arg)
+            {
+                HPyTupleBuilder builder = HPyTupleBuilder_New(ctx, 1);
+                HPyTupleBuilder_Set(ctx, builder, 0, h_arg);
+                HPy h_tuple = HPyTupleBuilder_Build(ctx, builder);
+                HPyTupleBuilder_Cancel(ctx);
+                return h_tuple;
+            }
+            @EXPORT(f)
+            @INIT
+        """)
+        assert mod.f("xy") == ("xy", True, -42)


hodgestar:
Is the current contract that `TupleBuilder_Cancel` must always be called even after `TupleBuilder_Build`? That seems a bit odd, but maybe it's easier to always do the clean up rather than have to only do it in error cases? Not sure.

hodgestar:
Either way -- maybe it's good to have the test with the current behaviour and the behaviour can be changed in a future PR if we decide to?

steve-s:
> Is the current contract that TupleBuilder_Cancel must always be called even after TupleBuilder_Build

No that was not the intention. The main question is if you can call `TupleBuilder_Cancel` even though you already successfully constructed the tuple using `TupleBuilder_Build`. It's something like `DECREF` vs `XDECREF` -- sometimes it leads to a simpler code if you can just put `TupleBuilder_Cancel` into some "cleanup" section, e.g.:

```
HPyTupleBuilder builder = HPyTupleBuilder_New(ctx, 1);
HPy result = HPy_NULL;
...
if (some_error_condition()) goto cleanup;
...
if (some_other_error_condition()) goto cleanup;
...
result = HPyTupleBuilder_Create(ctx, builder);

cleanup:
HPyTupleBuilder_Cancel(ctx, builder);
return result;
```

hodgestar:
That makes sense.

hodgestar (approved):
Test looks good once CI tests pass.

steve-s:
Ah, I thought it works so by just adding a test I would make it a contract. It worked only because I had an error in the test. :facepalm: 

Looking at the current implementation: the functions take `HPyTupleBuilder` by value, so resetting `_tup = 0` in 
```
ctx_TupleBuilder_Build(HPyContext *ctx, HPyTupleBuilder builder)
{
    PyObject *tup = (PyObject *)builder._tup;
    if (tup == NULL) {
        PyErr_NoMemory();
        return HPy_NULL;
    }
    builder._tup = 0;  // <====
    // ...
```
has no effect. Now it seem that the implementation of the contract I suggested would be unnecessarily complicated, so I am changing my mind and now I think that calling cancel after build should indeed have undefined behavior.

One more minor thing I noticed when looking at the API: the argument to `HPyTupleBuilder_New` is called `initial_size`. Before looking at the implementation I assumed that it is just a hint for the initial allocation, but that `HPyTupleBuilder_Set` would grow the tuple. I suggest renaming it to just `size`.

steve-s:
Declining this in the favor of https://github.com/hpyproject/hpy/issues/320

PR #327, opened 2022-06-03
Small autogen fixes and improvements.

fangerer:
* Fixes incompatibility with newest `pycparser`
* Add `requirements-autogen.txt` to install packages needed for autogen and testing.
* Sort declarations after parsing of `public_api.h` by line/col number to ensure deterministic autogen output.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -266,12 +266,9 @@ jobs:
       - name: Install/Upgrade Python dependencies
         run: python -m pip install --upgrade pip wheel
 
-      - name: Install dependencies
-        run: |
-          pip install pycparser==2.20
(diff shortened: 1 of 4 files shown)


steve-s (approved):

hodgestar (approved):
Looks good to me.

I'm not sure I understand how the ordering of the items in .functions, etc became non-deterministic in the first place. Does the parser not return things in the same order each time?

fangerer:
@hodgestar Not sure either. I _think_ it's if you use different `pycparser` versions. I tried to find some hints in the repo (https://github.com/eliben/pycparser) but wasn't successful.
I mostly observed the nondeterminism if autogen was triggered on different systems (e.g. on Stepan's box and on my box).

PR #328, opened 2022-06-03
RFC: Basic module state support

steve-s:
Adds support for specifying and using module state including the traverse and destroy slots. Module state can contain any plain C data and `HPyField`s. User can provide traverse and destroy slots like for types.

There is one thing missing: ability to query the module state from a class. In CPython this plays along with calling conventions that pass along defining class. HPy does not provide those calling conventions, but we can add them.

Introducing module state API similar to CPython will ease the migration to HPy for existing extensions that use CPython module state, and it adds one missing feature to complement the `HPyGlobal` and that a place for plain C per-interpreter state, e.g., some `int` counter cannot be stored in `HPyGlobal` (unless wrapped in Python object).

Before we introduce module state, we may consider some alternatives:

* https://github.com/hpyproject/hpy/blob/master/docs/module-state.txt
* instead of/along side calling conventions that pass along the defining class, we can add conventions that pass along the module state
* instead of module state, we can think about it as a module "context" similar to `HPyContext`, see below

We can also support module state and HPy specific alternative. If we merge this PR that would be the situation: we'd have more generic module state and specialized HPy specific concept of `HPyGlobal`.

Module context idea:

Like `HPyContext` it would be like and "argument", i.e., it would contain `HPy` handles directly and the user would have to `HPy_Dup` them if needed. They would be closed implicitly at the end of the "downcall".

This would allow to avoid the issue with `HPyGlobal` that to use a global, one has to do:
```
HPy type = HPyGlobal_Load(ctx, MyType);
HPy obj = HPy_New(ctx, type);
HPy_Close(ctx, type);
```
which is extra 2 calls. With something like a context it could be just `HPy_New(ctx, moduleCtx->myType)`, like it would be with some constant from `HPyContext`.

Alongside a calling convention that passes the module context as an argument, there would have to be a function to fetch the module context in places where it is not (easily) available and then probably some "close module context" function.

What is not clear is how the module context would be specified. Since it does not contain `HPyField`s, user cannot just assign `HPy` handles to it. Logically the module context would be created by the Python engine as an argument for each call (implementation may cache it).

### DIFF
--- a/hpy/debug/src/autogen_debug_ctx_init.h
+++ b/hpy/debug/src/autogen_debug_ctx_init.h
@@ -11,6 +11,7 @@
 */
 
 DHPy debug_ctx_Module_Create(HPyContext *dctx, HPyModuleDef *def);
+void *debug_ctx_Module_GetState(HPyContext *dctx, DHPy module);
 DHPy debug_ctx_Dup(HPyContext *dctx, DHPy h);
 void debug_ctx_Close(HPyContext *dctx, DHPy h);
 DHPy debug_ctx_Long_FromLong(HPyContext *dctx, long value);
@@ -243,6 +244,7 @@ static inline void debug_ctx_init_fields(HPyContext *dctx, HPyContext *uctx)
     dctx->h_TupleType = DHPy_open(dctx, uctx->h_TupleType);
     dctx->h_ListType = DHPy_open(dctx, uctx->h_ListType);
     dctx->ctx_Module_Create = &debug_ctx_Module_Create;
+    dctx->ctx_Module_GetState = &debug_ctx_Module_GetState;
     dctx->ctx_Dup = &debug_ctx_Dup;
     dctx->ctx_Close = &debug_ctx_Close;
     dctx->ctx_Long_FromLong = &debug_ctx_Long_FromLong;
--- a/hpy/debug/src/autogen_debug_wrappers.c
+++ b/hpy/debug/src/autogen_debug_wrappers.c
@@ -17,6 +17,11 @@ DHPy debug_ctx_Module_Create(HPyContext *dctx, HPyModuleDef *def)
     return DHPy_open(dctx, HPyModule_Create(get_info(dctx)->uctx, def));
 }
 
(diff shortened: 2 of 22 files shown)


steve-s:
One more thing to consider regarding NumPy port including its API. It exposes several global `PyObject*` variables, such as `PyArray_Type`. With `HPyGlobal` the conversion is straightforward (module the issue with compatible `HPyContext` struct layouts in the calling extensions and NumPy), with module state it does not seem so? AFAIK there is no way in CPython to get module state "out of thin air" and especially module state of another module.

mattip:
I think this API is used by [recordclass](https://pypi.org/project/recordclass/)

steve-s:
Update:
- in the dev call on 1.12.2022 we decided to go on with merging this as-is, I will just polish the code and fix the CI failures
- therefore: last call for reviews of the code/the idea as a whole
- we will need to add a new calling convention for the slots, such that users can access module state there (module is not passed to slots, see [CPython docs](https://docs.python.org/3/howto/isolating-extensions.html#module-state-access-from-regular-methods)) -- we may add the same calling convention as CPython added to support this, or we may add some more module state specific calling convention.

steve-s:
I would like to take back the call to just polish and merge this PR. I've read more on module state in CPython and I think we should figure out the whole thing first, especially how to access the module state in various other contexts:

* methods attached to a type (in CPython a new calling convention, can work for us fine)
* slots (in CPython new API that takes PyModuleDef, if we take that route, we would have to map `HPyModuleDef` to `PyModuleDef` -- doable, but bigger change worth upfront design)

On top of that: in the light of multi-phase module init, which allows to create multiple modules from one extension (in single interpreter), it seems to me that CPython's way of accessing module state from slots may have a fundamental flaw. Since module state is identified by `PyModuleDef` and that may be shared among multiple module instances.

With this understanding I am now also convinced that module state is necessary feature to complement or replace `HPyGlobal`, because `HPyGlobal` is fundamentally per extension and there's no simple way to make it per module instance. At the same time, I think there may be a value to `HPyGlobal`, because it provides more efficient and significantly simpler alternative for module state (no need to define traverse, etc.).

Issue #169, opened 2021-02-16
Add support for inheriting from built-in types other than PyObject (which is already supported)


hodgestar:
HPy supports two kinds of memory layouts in extension types:

* **Legacy** types have the same memory layout as in the old C API, and have structs that begin with PyObject_HEAD.
* **Pure** types do not include PyObject_HEAD.

When a legacy type inherits from another legacy type, or when a pure type inherits from another pure type, the situation is straight forward -- the inheriting type includes the full struct of the base type and optionally extends the struct with additional members.

Ideally HPy would simply forbid legacy types inheriting from pure types or vice versa -- having HPy magically add or remove PyObject_HEAD as needed by C slots or methods in the two types would be error prone and complex.

However, pure types may need to extend built-in types like PyLongObject, PyDictObject, or PyUnicodeObject. This is currently only partially supported as follows:

* **PyObject**: inheriting fully supported.
* **other built-in types**: supported, but only if no extra members are added to the struct, and if `HPy_AsStruct` is not used and and no deallocator (i.e. `HPy_tp_destroy`) is defined on the extending type.

Ideally pure types would treat other built-in types in the same way as `PyObject` -- i.e. all of their internal memory layout would be hidden from the C extension -- but the C Python implementation of HPy does not yet support this.

Support for inheriting from other built-in types in the same way as from `PyObject` should be added.

It might be possible to allow pure types to extend *any* legacy type in the same way (by making the struct of the existing type inaccessible to the slots and methods of the pure type) but this is a stretch goal.

hodgestar:
Things are further complicated because determining which type an object is inheriting its memory layout from is rather complex until after one has called `PyType_FromSpec`.

fangerer:
I think the main part of it is solved by PR #335 .
However, some ideas/points are still unresolved or untested:
* We don't test (and I'm not sure if that works): a legacy type inherits from a pure type. I'll investigate and open an issue if appropriate.
* Although it is possible for pure types to extend any legacy type (in the sense that at the time of type creation, we won't issue an error), it is actually not supported. Using the `<type_struct>_AsStruct` function will most likely return the wrong pointer. This is maybe something we should prevent by default. It might still make sense to allow it since one can use `_HPyType_GetBuiltinShape` to get the shape of the base type (which will be `HPyType_BuiltinShape_Legacy`)  and then it is possible to implement a custom `*_AsStruct` by manually adding the base's basicsize. I'll open a separate issue for that.

Issue #83, opened 2020-09-20
Redesign legacy types and HPy_CAST


antocuni:
After [my email](https://mail.python.org/archives/list/hpy-dev@python.org/thread/3TQ3AAY6AOUXZRAKWY42C77VL4AF54YC/) to hpy-dev, I and @arigo discussed a possible solution [on IRC](https://freenode.logbot.info/hpy/20200920#c5170940).
Here is a summary of the solution we found:

To talk more concretely, let's start from existing Python/C code like this:

```c
typedef struct {
    PyObject_HEAD
    int x;
    int y;
} PyPointObject;

// this is a method of Point
PyObject* Point_foo(PyObject *self, PyObject *arg)
{
    PyPointObject *p = (PyPointObject *)self;
    // ...
}

// this is an unrelated function which happens to cast a PyObject* into a
// PyPointObject*
void bar(PyObject *point)
{
    PyPointObject *p = (PyPointObject *)point;
    // ...
}
```

The idea is that an HPy type (as described by HPyType_Spec) can be:

1. Pure HPy:

    * `struct PyPointObject` does NOT contain any header. 
    
    * all methods and slots are written in HPy, and `.legacy_slots == NULL`.
    
    * if we obtain a `PyObject *` (e.g. via `HPy_AsPyObject()`), it is NOT possible to cast it to `PyPointObject *`.
   
2. Legacy type: 

    *  `struct PyPointObject` MUST start with `PyObject_HEAD`
    
    * `HPyType_Spec` MUST contain: `.legacy_headersize = offsetof(PyPointObject, x)`, (where `x` is the first field) and it is allowed to use `.legacy_slots`

    * It is possible to cast `PyObject *` to `PyPointObject *`

Properties of this solution:

1. We pay the space penalty for `PyObject_HEAD` only if it's explicitly requested
   
2. When we kill `PyObject_HEAD`, `.legacy_headersize` becomes automatically 0, and `HPyType_FromSpec` will immediately complain if we are still using `.legacy_slots`.
   
3. The C-casts inside legacy methods of the type (such as `Point_foo` above) are automatically valid, since if we kill `PyObject_HEAD` we can't have legacy methods (as explained above)

The remaining problem of this solution is how to "cast" an HPy handle into a PyPointObject* efficiently: Currently, we have `_HPy_Cast()`, but it is impossible to implement it efficiently, because we need to know whether the
[...]

PR #335, opened 2022-07-07
Support metaclasses in HPy.

### LINKED ISSUE #296: Support metaclass parameter in `HPyType_FromSpec`
This necessary for the numpy port. Numpy defines a metaclass of dtype class, with its own C struct, the gist is:
```
NPY_NO_EXPORT PyTypeObject PyArrayDTypeMeta_Type = {
    PyVarObject_HEAD_INIT(NULL, 0)
    .tp_name = "numpy._DTypeMeta",
    .tp_basicsize = sizeof(PyArray_DTypeMeta),
    // ...
}

typedef struct {
    PyHeapTypeObject super;
    // Numpy specific data:
    PyArray_Descr *singleton;
    int type_num;
    // ...
} PyArray_DTypeMeta;

NPY_NO_EXPORT PyArray_DTypeMeta PyArrayDescr_Type = {
    {{
        /* NULL represents `type`, this is set to DTypeMeta at import time */
        PyVarObject_HEAD_INIT(NULL, 0)
        .tp_name = "numpy.dtype",
        .tp_basicsize = sizeof(PyArray_Descr),
        // ...
    },},
    // values for the numpy specific fields:
    .type_num = -1,
    .singleton = NULL,
    // ...
};

// ... in module init:
Py_SET_TYPE(&PyArrayDescr_Type, &PyArrayDTypeMeta_Type);
if (PyType_Ready(&PyArrayDescr_Type) < 0) {
    goto err;
}
```

In order to support this, `HPyType_FromSpec` should take metaclass as an additional parameter. There is even a TODO for this in HPy. The HPy side of things seems to be clear.

However, the issue is that `PyType_FromSpecWithBases` does not support metaclasses (https://bugs.python.org/issue15870). I think that we will have to copy the logic of `PyType_FromSpecWithBases` from CPython and delegate back to CPython only at the point where the original code calls `PyType_Ready`. 

The problematic bit is that `PyType_FromSpecWithBases` calls `PyType_GenericAlloc(&PyType_Type, nmembers)`, it should call `tp_alloc` of the metaclass as suggested in https://bugs.python.org/issue15870.

fangerer:
This PR resolves #296 and is one of the big pieces that is required for migration of NumPy.

Goal of this PR: Support metaclasses for HPy types.

Before version 3.12, CPython does not support specifying metaclasses for heap types but only for static types.
We can still support that by creating a temporary heap type using `PyType_FromSpecWithBases` and then we manually allocate another heap type using the provided metaclass' `tp_alloc`. We then copy everything from the temporary heap type to the final one, reset some carefully picked slots and initialize it with `PyType_Ready`.
The temporary heap type is then trashed.

However, making that work with pure HPy types wasn't that trivial because a metaclass usually embeds the heap type struct like this:

```
typedef struct {
    PyHeapTypeObject super; // sizeof(PyHeapTypeObject) == 880
    int x;
    int y;
    // ....
} LegacyMetaClass;
```

The problem is that if we specify a pure HPy type and use it as metaclass, we would still just have following struct:
```
typedef struct {
    int x;
    int y;
    // ...
} MetaClass;
```
Since from an HPy point of view, the runtime's heap type is opaque, you cannot embed that in your structure and HPy assumes that it's just a bare `PyObject` and so the used offset is just `sizeof(PyObject)`.

However, each metaclass needs to inherit from `PyType_Type` (directly or indirectly). I use that fact and iterate over the bases to determine the largest basicsize. That is then used as an offset.
Furthermore, we now don't have a fixed offset for the pure HPy data pointer and so we need to store that in the HPy type's extra info.
This makes the `*_AsStruct` calls slightly more expensive because we need to load the offset from the object type's extra info.
I think, this change is not only a fix for metaclasses but in general a fix to be able to inherit from any opaque type. For example, an HPy extension could define a type that inherits from a foreign HPy extension's type which also specifies a basicsize.
Furthermore, as far as I know, the situation is similar in the C API when using the limited API. The structs are opaque and if one wants to inherit from a type, he needs to add the base type's basicsize.

### DIFF
--- a/hpy/devel/src/runtime/ctx_type.c
+++ b/hpy/devel/src/runtime/ctx_type.c
@@ -1105,6 +1105,8 @@ ctx_Type_FromSpec(HPyContext *ctx, HPyType_Spec *hpyspec,
     /* On Python 3.12 an newer, we can just use 'PyType_FromMetaclass'. */
     PyObject *result = PyType_FromMetaclass(meta, NULL, spec, bases);
 #else
+    /* On Python 3.11 an older, we need to use our own
+       '_PyType_FromMetaclass'. */
     PyObject *result = _PyType_FromMetaclass(spec, bases, metatype);
 #endif
 
--- a/hpy/devel/src/runtime/ctx_type.c
+++ b/hpy/devel/src/runtime/ctx_type.c
@@ -985,11 +985,13 @@ _PyType_FromMetaclass(PyType_Spec *spec, PyObject *bases,
         if (PyType_Ready(tp) < 0)
             goto fail;
 
-        /* The following is the tail of 'PyType_FromSpecWithBases'. */
-
-        if (tp->tp_dictoffset) {
-            ht->ht_cached_keys = _PyDict_NewKeysForClass();
+        /* Restore 'ht_cached_keys' after call to 'PyType_Ready' */
+        if (temp_ht->ht_cached_keys) {
+            Py_INCREF(temp_ht->ht_cached_keys);
+            ht->ht_cached_keys = temp_ht->ht_cached_keys;
         }
+
+        /* The following is the tail of 'PyType_FromSpecWithBases'. */
 
         /* Set type.__module__ */
         s = strrchr(spec->name, '.');
--- a/hpy/devel/src/runtime/ctx_type.c
+++ b/hpy/devel/src/runtime/ctx_type.c
@@ -874,6 +874,134 @@ _HPy_HIDDEN struct _typeobject *get_metatype(HPyType_SpecParam *params) {
     return &PyType_Type;
 }
 
+static PyObject *
+_PyType_FromMetaclass(PyType_Spec *spec, PyObject *bases, struct _typeobject *meta)
+{
+#if PY_VERSION_HEX >= 0x030C0000
+    /* On Python 3.11 an newer, we can just use 'PyType_FromMetaclass'. */
+    return PyType_FromMetaclass(meta, NULL, spec, bases);
+#else
+    /* On older Python versions, we need to workaround the missing support for
+       metaclasses. We create a temporary heap type using
+       'PyType_FromSpecWithBases' and if a metaclass was provided, we use it to
+       allocate the appropriate type object and memcpy most of the contents
+       from the heap type to the manually allocated one. Then we clear some key
+       slots and call 'PyType_Ready' on it to re-initialize everything. The
+       temporary heap type is then expired. */
+
+    PyObject *temp = PyType_FromSpecWithBases(spec, bases);
+    if (!temp)
+        return NULL;
+
+    if (meta)
+    {
+        PyHeapTypeObject *temp_ht = (PyHeapTypeObject *) temp;
+        PyTypeObject *temp_tp = &temp_ht->ht_type;
+
+        Py_INCREF(temp_ht->ht_name);
+        Py_INCREF(temp_ht->ht_qualname);
+        Py_XINCREF(temp_ht->ht_slots);
+        Py_INCREF(temp_tp->tp_base);
+
+        /* Count the members as 'PyType_FromSpecWithBases' does such that we
+           can properly allocate the size later when allocating the type. */
+        Py_ssize_t nmembers = 0;
+        PyType_Slot *slot;
+        PyMemberDef *memb;
+        for (slot = spec->slots; slot->slot; slot++) {
+            if (slot->slot == Py_tp_members) {
+                nmembers = 0;
+                for (memb = slot->pfunc; memb->name != NULL; memb++) {
+                    nmembers++;
+                }
+            }
+        }
+
+        PyObject *result = PyType_GenericAlloc(meta, nmembers);
+        if (!result)
+            goto fail;
+
+        PyHeapTypeObject *ht = (PyHeapTypeObject *) result;
+        PyTypeObject *tp = &ht->ht_type;
+
+        /* Set the type name and qualname */
+        const char *s = strrchr(spec->name, '.');
+        if (s == NULL)
+            s = (char*)spec->name;
+        else
+            s++;
+
+        memcpy(ht, temp_ht, sizeof(PyHeapTypeObject));
+
+        tp->ob_base.ob_base.ob_type = meta;
+        tp->ob_base.ob_base.ob_refcnt = 1;
+        tp->ob_base.ob_size = 0;
+        tp->tp_as_async = &ht->as_async;
+        tp->tp_as_number = &ht->as_number;
+        tp->tp_as_sequence = &ht->as_sequence;
+        tp->tp_as_mapping = &ht->as_mapping;
+        tp->tp_as_buffer = &ht->as_buffer;
+        tp->tp_flags = spec->flags | Py_TPFLAGS_HEAPTYPE;
+
+        tp->tp_dict = NULL;
+        tp->tp_bases = NULL;
+        tp->tp_mro = NULL;
+        tp->tp_cache = NULL;
+        tp->tp_subclasses = NULL;
+        tp->tp_weaklist = NULL;
+        ht->ht_cached_keys = NULL;
+        tp->tp_version_tag = 0;
+
+        /* Refresh 'tp_doc'. This is necessary because
+           'PyType_FromSpecWithBases' allocates its own buffer which will be
+           free'd. */
+        if (temp_tp->tp_doc)
+        {
+            size_t len = strlen(temp_tp->tp_doc)+1;
+            char *tp_doc = PyObject_MALLOC(len);
+            if (!tp_doc)
+                goto fail;
+            memcpy(tp_doc, temp_tp->tp_doc, len);
+            tp->tp_doc = tp_doc;
+        }
+
+        /* Sanity check: GC objects need to provide 'tp_traverse' and
+           'tp_clear'. */
+        assert(!PyType_IS_GC(tp) || tp->tp_traverse != NULL || tp->tp_clear != NULL);
+
+        PyType_Ready(tp);
+
+        /* The following is the tail of 'PyType_FromSpecWithBases'. */
+
+        if (tp->tp_dictoffset) {
+            ht->ht_cached_keys = _PyDict_NewKeysForClass();
+        }
+
+        /* Set type.__module__ */
+        s = strrchr(spec->name, '.');
+        if (s != NULL) {
+            int err;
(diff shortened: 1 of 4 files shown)


hpy/devel/src/runtime/ctx_type.c
```diff
@@ -715,18 +934,41 @@ ctx_Type_FromSpec(HPyContext *ctx, HPyType_Spec *hpyspec,
...
-            basicsize = hpyspec->basicsize + _HPy_PyObject_HEAD_SIZE;
-            base_member_offset = _HPy_PyObject_HEAD_SIZE;
+            if (bases) {
+                /* If base classes were specified, we need to compute the
+                   largest basicsize of all bases. We
+                   will then use the base's basicsize as offset for the pure
+                   HPy type's struct. In most cases, that will just be
+                   PyObject_HEAD but it could, e.g., also be PyHeapTypeObject
+                   in case of metaclasses. */
+                base_member_offset = compute_largest_basicsize(bases);
```
steve-s, line 959: I think there could be in theory an issue with memory alignment here that's why the other branch is using `_HPy_PyObject_HEAD_SIZE`, which is `sizeof` of some struct padded with extra field with a union of all possible types. We can probably change that union to a `max_align_t` field. In this branch we can maybe at least make the `base_member_offset` be divisible by `sizeof(max_align_t)`? Not sure if there is any more portable/less hacky solution.

steve-s:
From my POV this looks good modulo the failing CI jobs and is an important step forward for NumPy porting effort. It would be great if someone else could take a look too (@antocuni, @hodgestar, ... :-))

fangerer:
Yes, I'm sorry. MSVC is a bit more pedantic than `clang` or `gcc`.
I took the opportunity to write a test which is not straight forward since we should not make too many assumptions.

antocuni (commented):
wow, this was a tremendous amount of work, good job @fangerer and @steve-s 👍.
The PR looks good in general, but before merging I would like to explore whether there are other ways to avoid paying the penalty in `HPy_AsStruct`.

`HPy_AsStruct` will be called *a lot*, and we should try at all costs to avoid making it unnecessarily slow, especially considering that it's a penalty that we are going to pay also in CPython mode, which in theory should be 0 costs.
I also have the feeling that we are going in circles, because when we proposed #83, we did it specifically to avoid paying the cost of this kind of "dynamic cast".

I *think* there is a possible solution which came to my mind while reading this PR, so let me explain it here although it might be wrong.

If I understand correctly, the problem comes when subclassing a builtin type which is not `PyObject`. But the key realization is that there is only a finite number of those: `PyHeapTypeObject`, `PyLongObject`, `PyDictObject` and probably a handful more. The relevant issue is #169.

Currently, we distinguish between "pure types" and "legacy types". I propose to treat specially all built-in type that can be subclassed (or at least, the ones for which we care about performance). So, we will have:
```c
HPy_AsStruct(); // works only for pure types
HPy_AsStruct_Generic(); // this is the "slow" version, which dynamically looks up the payload_offset. Maybe we should call it HPy_AsStruct_Dynamic? Or _Slow?
HPy_AsStruct_Legacy_PyObject();
HPy_AsStruct_Legacy_PyHeapTypeObject();
...
```
`HPyType_DEFINE_HELPERS` will be able to use the most appropriate version, and probably we can teach the debug mode to detect the case in which we are using a specialized version which does not match what's written in `payload_offset`.

Note that we still need the "generic" slow version for cases in which we don't know upfront the type, but this is already the case in the current master: we have `_pyobj_as_struct` which does exactly that and it's called from a couple of internal places:
https://github.com/hpyproject/hpy/blob/d262f43520b02c1e7244a0aaee97a4178fd7e7d5/hpy/devel/src/runtime/ctx_type.c#L73-L81

Also, if we decide to go down that route, we can probably remove the `is_pure` flag:
```c
typedef struct {
    uint16_t magic;
    HPyFunc_traverseproc tp_traverse_impl;
    HPyFunc_destroyfunc tp_destroy_impl;
    bool is_pure;
    HPy_ssize_t payload_offset;
    char name[];
} HPyType_Extra_t;
```

I _think_ that `payload_offset` should be enough: a type is pure IIF `payload_offset == 0`.

steve-s:
> If I understand correctly, the problem comes when subclassing a builtin type which is not `PyObject`. But the key realization is that there is only a finite number of those: `PyHeapTypeObject`, `PyLongObject`, `PyDictObject` and probably a handful more. The relevant issue is #169.

I though the problem here was that we have also custom subclasses. However, now I realized we do not have this problem here. At the same time, we still need to know the right offset, and even for pure types (or only for pure types actually). Let's have some example for discussion:

```
struct {
   PyHeapTypeObject super;
   int data;
} DTypeMeta;

static HPyType_Spec DTypeMeta_spec = {
  .basicsize = sizeof(DTypeMeta)
  // ...
};

struct {
   PyObject_HEAD
   int other_data;
} DType;

static HPyType_Spec DType_spec = {
  .basicsize = sizeof(DType)
  // ...
};

// inicialization:
HPyType_SpecParam p1[] = { { HPyType_SpecParam_Base, ctx->h_TypeType }, 0 };
HPy dtype_meta = HPyType_FromSpec(ctx, DTypeMeta_spec, p1);

HPyType_SpecParam p2[] = { { HPyType_SpecParam_Metaclass, metaclass }, 0 };
HPy dtype = HPyType_FromSpec(ctx, DType_spec, p2);
```

To recap what is going on there: `DTypeMeta` subclasses builtin `PyHeapTypeObject`, `dtype` is an instance of `DTypeMeta`, so `HPy_AsStruct(ctx, dtype)` should ideally give us `DTypeMeta*`. 

With HPy pure types the `PyHeapTypeObject super` field would disappear. Since it should be opaque anyway, we do not have anything to embed into our `DTypeMeta`. We just specify that the base class is `ctx->h_Type`. The issue is that when in HPy we call `PyType_FromSpecWithBases` we need to set basicsize to `sizeof(PyHeapTypeObject) + sizeof(DTypeMeta)` (ignoring alignment for now) for this to work on CPython. In `HPy_AsStruct` when we have a `PyObject*` at hand that is actually `PyHeapTypeObject*` in order to get the right offset we need to know that it's in fact `PyHeapTypeObject*` (or know directly the offset to use).

The suggested solution looks good if we really fear that fishing the offset from the type would be unacceptably slow. However, I think we'll need to have `AsStruct` variant for all combinations of builtin type x legacy/non-legacy, so in concrete terms:
```
HPy_AsStruct
HPy_AsStruct_Legacy
HPyType_AsStruct
HPyType_AsStruct_Legacy
HPyLong_AsStruct
HPyLong_AsStruct_Legacy
etc.
```
in fact, I think that the legacy ones are not necessary, because there the user should already embed the right struct inside their struct, so we do not have to calculate any offset, but such a difference between pure and legacy types would make porting harder imo. For the helpers: one would have to specify the base type as an argument of the macro, or we'd have different macros:
```
HPyType_HELPERS
HPyTypeType_HELPERS?
HPyLongType_HELPERS?
```
For "dynamic" `HPy_AsStruct` variant -- is that actually useful? Because if I do not know what is coming from `HPy_AsStruct`, I cannot make any sensible use of it? If I do not know what type of object my handle points to, I need to do some type check and after that I know that it's, for example, some subclass of Long, and so I know I need to use `HPyLong_AsStruct`.

---
Completely different thing is inheritance of custom user defined types (and their structs). I believe that inheritance of custom types must be actually treated differently from builtin types, because builtin types must stay opaque, while custom types should be accessible to be actually useful. There I'd agree with what we discussed on the last call, that the user should actually embed structs, example:
```
struct {
   int data;
} MyBaseType;

static HPyType_Spec MyBaseType_spec = {
  .basicsize = sizeof(MyBaseType)
  // ...
};

struct {
   MyBaseType base;
   int other_data;
} MyType;

static HPyType_Spec MyType_spec = {
  .basicsize = sizeof(MyType)
  // ...
};

// init:
HPy base = HPyType_FromSpec(ctx, MyBaseType_spec, { 0 });

HPyType_SpecParam p2[] = { { HPyType_SpecParam_Base, base }, 0 };
HPy my_type = HPyType_FromSpec(ctx, MyType_spec, p2);
```
here we will need to distinguish that in this case `basicsize` of `MyType` already contains the embedded `MyBaseType` and we should not calculate the `basicsize` to be passed to `PyType_FromSpecWithBases` as `basicsize` of `MyBaseType` plus our `basicsize`. This will make the implementation bit cumbersome, but it sounds like it could be implemented fine?

Last situation that comes to my mind is inheriting from a type from another extension, which is however not exposed via some shared header file or such, so its struct cannot be embedded. In that case I would say that setting `basicsize` to anything else than `0`, i.e., requesting some memory for a custom struct, should be just prohibited. It seems that this is something that couldn't work before with either static or heap types on CPython, anyway?

steve-s:
This is a relevant thread that proposes similar APIs for C API: https://mail.python.org/archives/list/capi-sig@python.org/thread/SIP3VP7JU4OBWP62KBOYGOYCVIOTXEFH/

encukou:
FWIW, turning that proposal into s PEP is now almost on top of my CPython TODO list (after https://github.com/python/cpython/pull/95992 and a bit of non-CPython work)

hpy/devel/src/runtime/ctx_type.c
```diff
@@ -692,7 +739,205 @@ static PyObject *build_bases_from_params(HPyType_SpecParam *params)
...
+
+/* On older Python versions, we need to workaround the missing support for
+   metaclasses. We create a temporary heap type using
+   'PyType_FromSpecWithBases' and if a metaclass was provided, we use it to
+   allocate the appropriate type object and memcpy most of the contents from
+   the heap type to the manually allocated one. Then we clear some key slots
+   and call 'PyType_Ready' on it to re-initialize everything. The temporary
+   heap type is then expired. */
+static PyObject*
+_PyType_FromMetaclass(PyType_Spec *spec, PyObject *bases,
```
timfel, line 814: Add `#ifdef PY_VERSION_HEX` around this

timfel:
Discussion at sprint: simplify this to not dynamically calculate the payload, but instead just solve the issue specifically for `PyHeapTypeObject_HEAD` and thus statically. /cc @fangerer

encukou:
I've proposed [PEP 697](https://peps.python.org/pep-0697). Could you please check it out, and [comment on Discourse](https://discuss.python.org/t/19743)?

fangerer:
> I've proposed [PEP 697](https://peps.python.org/pep-0697). Could you please check it out, and [comment on Discourse](https://discuss.python.org/t/19743)?

Will do so.

PR #336, opened 2022-07-18
update docs

koubaa:
Here I propose some minor changes to the docs

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -79,7 +79,7 @@ Becomes using HPy API:
 Calling any HPy function on a closed handle is an error. Calling
 ``HPy_Close()`` on the same handle twice is an error. Forgetting to call
 ``HPy_Close()`` on a handle results in a memory leak. When running in
-:ref:`debug-mode:debug mode`, HPy actively checks that you that you don't
+:ref:`debug-mode:debug mode`, HPy actively checks that you don't
 close a handle twice and that you don't forget to close any. This is
 possible because handles are unique - the handle used to reference an
 object is valid only in the context of the operation that produced it.
@@ -91,7 +91,7 @@ object is valid only in the context of the operation that produced it.
   If you find a memory leak on CPython, you know that you are missing a
   ``Py_DECREF`` somewhere but the only way to find the corresponding
   ``Py_INCREF`` is to manually and carefully study the source code.
-  On the other hand, if you forget to call ``HPy_Close()``, Debug mode
+  On the other hand, if you forget to call ``HPy_Close()``, debug mode
   is able to identify the precise code location which created the unclosed
   handle. Similarly, if you try to operate on a closed handle, it will
   identify the precise code locations which created and closed it.
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -48,7 +48,7 @@ Handles vs ``PyObject *``
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 
 In order to fully understand the way HPy handles work, it is useful to discuss
-the Python/C API ``Pyobject *`` pointer. In the latter, these pointers always
+the Python/C API ``Pyobject *`` pointer. These pointers always
 point to the same object, and a python object's identity is completely given
 by its address in memory, and two pointers with the same address can
 be passed to Python/C API functions interchangeably. As a result, ``Py_INCREF``
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -129,7 +129,7 @@ identity, you can use ``HPy_Is()``:
    Unlike CPython, PyPy does not use reference counting to manage memory:
    instead, it uses a *moving GC*, which means that the address of an object
    might change during its lifetime, and this makes it hard to implement
-   semantics like ``PyObject *``'s where the address *idenfies* the object,
+   semantics like ``PyObject *``'s where the address *identifies* the object,
    and this is directly exposed to the user.  HPy solves this problem: on
    PyPy, handles are integers which represent indices into a list, which
    is itself managed by the GC. When an address changes, the GC edits the
@@ -150,8 +150,8 @@ future-proof: it is conceivable to use it to hold the interpreter or the
 thread state in the future, in particular when there will be support for
 sub-interpreters.  Another possible usage could be to embed different versions
 or implementations of Python inside the same process. Its likely that
-future extensions to rune and extend the behavior of HPy would happen here,
-which would not break anything built against HPy.
+future additions to HPy would happen here, which would not break anything
+built against the current ``HPyContext``.
 
 Moreover, ``HPyContext`` is used by the :term:`HPy Universal ABI` to contain a
 sort of virtual function table which is used by the C extensions to call back
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -80,10 +80,7 @@ Calling any HPy function on a closed handle is an error. Calling
 ``HPy_Close()`` on the same handle twice is an error. Forgetting to call
 ``HPy_Close()`` on a handle results in a memory leak. When running in
 :ref:`debug-mode:debug mode`, HPy actively checks that you don't
-close a handle twice and that you don't forget to close any. This is
-possible because handles are associated with operations on objects,
-meaning that if a handle leaks, it is possible to identify exactly
-the operation which produced it.
+close a handle twice and that you don't forget to close any.
 
 
 .. note::
@@ -95,7 +92,10 @@ the operation which produced it.
   On the other hand, if you forget to call ``HPy_Close()``, debug mode
   is able to identify the precise code location which created the unclosed
   handle. Similarly, if you try to operate on a closed handle, it will
-  identify the precise code locations which created and closed it.
+  identify the precise code locations which created and closed it. This is
+  possible because handles are associated with a single call to a C/API
+  function. As a result, given a handle that is leaked or used after freeing,
+  it is possible to identify exactly the C/API function that producted it.
 
 
 Remember that Python/C guarantees that multiple references to the same
@@ -149,8 +149,9 @@ One of the reasons to include ``HPyContext`` from the day one is to be
 future-proof: it is conceivable to use it to hold the interpreter or the
 thread state in the future, in particular when there will be support for
 sub-interpreters.  Another possible usage could be to embed different versions
-or implementations of Python inside the same process. In addition, the ``HPyContext`` may
-also be extended by adding new functions to the end without breaking any extensions built against the current ``HPyContext``.
+or implementations of Python inside the same process. In addition, the
+``HPyContext`` may also be extended by adding new functions to the end without
+breaking any extensions built against the current ``HPyContext``.
 
 Moreover, ``HPyContext`` is used by the :term:`HPy Universal ABI` to contain a
 sort of virtual function table which is used by the C extensions to call back
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -81,8 +81,9 @@ Calling any HPy function on a closed handle is an error. Calling
 ``HPy_Close()`` on a handle results in a memory leak. When running in
 :ref:`debug-mode:debug mode`, HPy actively checks that you don't
 close a handle twice and that you don't forget to close any. This is
-possible because handles are unique - the handle used to reference an
-object is valid only in the context of the operation that produced it.
+possible because handles are associated with operations on objects,
+meaning that if a handle leaks, it is possible to identify exactly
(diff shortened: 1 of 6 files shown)


docs/api.rst
```diff
@@ -80,26 +79,28 @@ Becomes using HPy API:
 Calling any HPy function on a closed handle is an error. Calling
 ``HPy_Close()`` on the same handle twice is an error. Forgetting to call
 ``HPy_Close()`` on a handle results in a memory leak. When running in
-:ref:`debug-mode:debug mode`, HPy actively checks that you that you don't close a handle
-twice and that you don't forget to close any.
+:ref:`debug-mode:debug mode`, HPy actively checks that you that you don't
```
steve-s, line 82: "that you" is repeated twice

docs/api.rst
```diff
@@ -80,26 +79,28 @@ Becomes using HPy API:
...
-point to the same object::
+  Debug mode is a good example of how powerful it is to decouple the
+  identity and therefore the lifetime of handles and those of objects.
+  If you find a memory leak on CPython, you know that you are missing a
+  ``Py_DECREF`` somewhere but the only way to find the corresponding
+  ``Py_INCREF`` is to manually and carefully study the source code.
+  On the other hand, if you forget to call ``HPy_Close()``, Debug mode
+  is able to identify the precise code location which created the unclosed
+  handle. Similarly, if you try to operate on a closed handle, it will
+  identify the precise code locations which created and closed it.
```
steve-s, line 97: I like this explanation :+1:

docs/api.rst
```diff
@@ -147,7 +148,9 @@ One of the reasons to include ``HPyContext`` from the day one is to be
 future-proof: it is conceivable to use it to hold the interpreter or the
 thread state in the future, in particular when there will be support for
 sub-interpreters.  Another possible usage could be to embed different versions
-or implementations of Python inside the same process.
+or implementations of Python inside the same process. Its likely that
+future extensions to rune and extend the behavior of HPy would happen here,
```
steve-s, line 152: "to rune" is a typo or I don't understand it?

docs/overview.rst
```diff
@@ -181,23 +176,32 @@ or HPy Hybrid ABIs require installing the ``hpy.universal`` package on
...
     many common mistakes are checked and reported automatically.
-
-  - Universal binaries: you can choose to distribute only Universal ABI
-    binaries. This comes with a small speed penalty on CPython, but for
-    non-performance critical libraries it might still be a good tradeoff.
+  - Universal binaries: libraries can choose to distribute only Universal ABI
+    binaries. By doing so, they can support all versions of CPython for which
+    an hpy loader exists, including versions of CPython that have not yet been
+    developed. This currently comes with a small speed penalty on CPython, but
+    for non-performance critical libraries it might still be a good tradeoff.
```
steve-s, line 195: The universal binaries also work on any other Python, like PyPy or GraalPython. Even the original text is advertising that clearly, but maybe a chance to improve that.

docs/overview.rst
```diff
@@ -181,23 +176,32 @@ or HPy Hybrid ABIs require installing the ``hpy.universal`` package on
...
+  - Implementation-switching Python environments: Users can currently create
+    Python environments for a given implementation with a set of library
+    requirements. With universal binaries, users can create Python environments
+    with any permutation of both library requirements and choice of
+    implementation. This allows users to choose the best implementation for
+    their workload.
```
steve-s, line 201: I needed few seconds to parse this. It also follows from the previous point, right?, but maybe it can be viewed as such an advantage that it is worth pointing out explicitly. From my POV it seems bit complex, too abstract and obscure here, but I am open to be convinced otherwise

docs/api.rst
```diff
@@ -80,26 +79,28 @@ Becomes using HPy API:
 Calling any HPy function on a closed handle is an error. Calling
 ``HPy_Close()`` on the same handle twice is an error. Forgetting to call
 ``HPy_Close()`` on a handle results in a memory leak. When running in
-:ref:`debug-mode:debug mode`, HPy actively checks that you that you don't close a handle
-twice and that you don't forget to close any.
+:ref:`debug-mode:debug mode`, HPy actively checks that you that you don't
+close a handle twice and that you don't forget to close any. This is
+possible because handles are unique - the handle used to reference an
+object is valid only in the context of the operation that produced it.
```
steve-s, line 85: I don't know if `in the context of the operation that produced it` isn't simplifying it bit too much or if it cannot be bit misleading?

steve-s (commented):
Overall this looks very good to me. Thank you for your contribution.

docs/api.rst
```diff
@@ -118,20 +119,20 @@ identity, you can use ``HPy_Is()``:
...
+   The main benefit of opaque handle semantics is that implementations are
+   allowed to use very different models of memory management.  On CPython,
+   implementing handles is trivial because ``HPy`` is basically ``PyObject *``
+   in disguise, and ``HPy_Dup()`` and ``HPy_Close()`` are just aliases for
+   ``Py_INCREF`` and ``Py_DECREF``.
+
+   Unlike CPython, PyPy does not use reference counting to manage memory:
+   instead, it uses a *moving GC*, which means that the address of an object
+   might change during its lifetime, and this makes it hard to implement
+   semantics like ``PyObject *``'s where the address *idenfies* the object,
```
fangerer, line 131: typo: *idenfies* -> *identifies*

docs/api.rst
```diff
@@ -80,26 +79,28 @@ Becomes using HPy API:
...
-The other important difference is that Python/C guarantees that multiple
-references to the same object results in the very same ``PyObject *`` pointer.
-Thus, it is possible to compare C pointers by equality to check whether they
-point to the same object::
+  Debug mode is a good example of how powerful it is to decouple the
+  identity and therefore the lifetime of handles and those of objects.
+  If you find a memory leak on CPython, you know that you are missing a
+  ``Py_DECREF`` somewhere but the only way to find the corresponding
+  ``Py_INCREF`` is to manually and carefully study the source code.
+  On the other hand, if you forget to call ``HPy_Close()``, Debug mode
```
fangerer, line 94: minor: You use `Debug mode` (upper-case `D` as it would be a name) but we don't do that in other locations. Maybe just use `debug mode`?

docs/overview.rst
```diff
@@ -181,23 +176,32 @@ or HPy Hybrid ABIs require installing the ``hpy.universal`` package on
...
+    written in Python/C.
   - Improved debugging: when you load extensions in :ref:`debug-mode:debug mode`,
     many common mistakes are checked and reported automatically.
-
-  - Universal binaries: you can choose to distribute only Universal ABI
-    binaries. This comes with a small speed penalty on CPython, but for
-    non-performance critical libraries it might still be a good tradeoff.
+  - Universal binaries: libraries can choose to distribute only Universal ABI
+    binaries. By doing so, they can support all versions of CPython for which
+    an hpy loader exists, including versions of CPython that have not yet been
```
fangerer, line 193: `hpy` -> `HPy`

fangerer (commented):
Thank you for the docs update.

docs/api.rst
paugier, line 82: `that you that you`

docs/overview.rst
koubaa: Elsewhere in the documentation we highlight the advantages to Python implementations. Before my change here - the advantages discussed are for authors of CPython, which follows from the old title of this section.

I renamed the title and added this point (and the below one) to show that HPy offers advantages not only to CPython extension authors, but to the ecosystem as a whole.  This includes users.  As a Python user, if I get python from python.org nothing changes.  But if I use a higher-level python environment system (such as conda) - it is possible for a hypothetical python environment system - let's call it honda - to do things like this:

```
honda create-env Variant1 --python pypy:latest --requirements requirements1.txt
honda create-env Variant2 --python CPython:3.12 --requirements requirements2.txt
honda create-env Variant3 --python graal:latest --requirements requirements1.txt
honda create-env Variant4 --python cinder:20 --requirements requirements3.txt
honda create-env Variant5 --python nogil-python:11 --requirements requirements4.txt
honda create-env Variant6 --python gc-nogil-python:22 --requirements requirements2.txt
honda create-env Variant7 --python my-experimental:5 --requirements requirements2.txt
```

where python interpreters use a docker-like convention for name:version.  Then, users could benchmark one against another to try things out. This could be interesting for python users, who could switch implementations easily, letting `honda` figure out where the packages in the requirements file are supported for a given implementation.

steve-s: Thanks for the explanation. I get the idea. I still wonder if this isn't too abstract for an overview, adding an example or being more concrete would help, but then it would be too long/complex topic for an overview imo. Maybe a concrete example how it can help specifically conda users would be more explanatory, but still short? At the same time, it's definitely not a big issue. Unless anyone else also shares my concerns, I would not sink too much time into this. From my POV this is not a blocker.

koubaa: I reworded it.

koubaa:
This is ready to merge.  I'm willing to follow up with more doc improvements - please point out which areas you feel need the most attention.

fangerer (approved):
LGTM but @hodgestar announced in the IRC channel that he will also have a look at this PR. Let's just wait a little bit longer for his review before we merge.

docs/api.rst
```diff
@@ -47,15 +47,14 @@ Unix file descriptors, where you have ``dup()`` and ``close()``, and Windows'
...
-   right now. Please rephrase/rewrite :)
-
-In the old Python/C API, multiple ``PyObject *`` references to the same object
-are completely equivalent to each other. Therefore they can be passed to Python/C
-API functions interchangeably. As a result, ``Py_INCREF`` an ``Py_DECREF`` can
-be called with any reference to an object as long as the total number of calls
-of `incref` is equal to the number of calls of `decref` at the end of the object
-lifetime.
+In order to fully understand the way HPy handles work, it is useful to discuss
+the Python/C API ``Pyobject *`` pointer. In the latter, these pointers always
```
hodgestar, line 51:
```suggestion
the Python/C API ``Pyobject *`` pointer. These pointers always
```
There is no "latter" to reference since there is only a single case (i.e. the pointer).

docs/api.rst
```diff
@@ -80,26 +79,29 @@ Becomes using HPy API:
...
+  On the other hand, if you forget to call ``HPy_Close()``, debug mode
+  is able to identify the precise code location which created the unclosed
+  handle. Similarly, if you try to operate on a closed handle, it will
+  identify the precise code locations which created and closed it.
+
+
+Remember that Python/C guarantees that multiple references to the same
+object results in the very same ``PyObject *`` pointer. Thus, it is
+possible to compare the pointer addresses to check whether they refer
+to the same object::
```
hodgestar, line 104: I think the reason that tracking handle usage on HPy is easier is much simpler:

In HPy each handle is created (dupped) and closed or returned exactly once within one call to a C extension function.

In the old C API, references may be incremented and decremented across all of the running code over the lifetime of the entire Python interpreter.

docs/api.rst
```diff
@@ -147,7 +149,9 @@ One of the reasons to include ``HPyContext`` from the day one is to be
 future-proof: it is conceivable to use it to hold the interpreter or the
 thread state in the future, in particular when there will be support for
 sub-interpreters.  Another possible usage could be to embed different versions
-or implementations of Python inside the same process.
+or implementations of Python inside the same process. Its likely that
+future additions to HPy would happen here, which would not break anything
+built against the current ``HPyContext``.
```
hodgestar, line 154:
```suggestion
or implementations of Python inside the same process. In addition, the ``HPyContext`` may
also be extended by adding new functions to the end without breaking any extensions built against the current ``HPyContext``.
```

hodgestar (changes requested):
Left a few comments but looks great!

hodgestar (approved):

hodgestar:
@koubaa It looks like the documentation build failed with
```
.../hpy/docs/overview.rst:390:Explicit markup ends without a blank line; unexpected unindent.
```
Could you fix that and then hopefully we can merge?

koubaa:
> @koubaa It looks like the documentation build failed with
> 
> ```
> .../hpy/docs/overview.rst:390:Explicit markup ends without a blank line; unexpected unindent.
> ```
> 
> Could you fix that and then hopefully we can merge?

fixed

fangerer:
I'm merging this PR. Thank you again!

antocuni:
A very late review admittedly, but I like the changes! Thank you @koubaa !

PR #344, opened 2022-09-09
Backport changes needed to support PyPy

mattip:
The changes are
- disallow setting `on_invalid_handle` hook in debug mode, since setting the hook will fail on non-cpython in accordance with [this comment](https://github.com/hpyproject/hpy/commit/51c3212), which should be in a more permanent place
- add a gc.collect in a test
- use relative imports to allow vendoring the test suite
- adapt the `SUPPORTS_MEM_PROTECTION` checks to PyPy

### DIFF
--- a/hpy/devel/include/hpy/hpydef.h
+++ b/hpy/devel/include/hpy/hpydef.h
@@ -10,7 +10,7 @@ extern "C" {
 #include "hpy/autogen_hpyslot.h"
 #include "hpy/cpy_types.h"
 
-typedef void* (*HPyCFunction)();
+typedef void* (*HPyCFunction)(void);
 
 typedef struct {
     HPySlot_Slot slot;     // The slot to fill
--- a/test/debug/test_charptr.py
+++ b/test/debug/test_charptr.py
@@ -1,5 +1,6 @@
-from test.support import SUPPORTS_SYS_EXECUTABLE
+from ..support import SUPPORTS_SYS_EXECUTABLE
 import os
+import sys
 import pytest
 
 # Tests detection of usage of char pointers associated with invalid already
@@ -46,7 +47,9 @@ def test_charptr_use_after_implicit_arg_handle_close(compiler, python_subprocess
     result = python_subprocess.run(mod, code)
     assert result.returncode != 0
     assert result.stdout == b""
-    if SUPPORTS_MEM_PROTECTION:
+    if sys.implementation.name == 'pypy':
+        assert b"RPython" in result.stderr
+    elif SUPPORTS_MEM_PROTECTION:
         assert result.stderr == b""
     else:
         # The garbage we override the data with will cause this error
@@ -76,7 +79,9 @@ def test_charptr_use_after_handle_close(compiler, python_subprocess):
     result = python_subprocess.run(mod, code)
     assert result.returncode != 0
     assert result.stdout == b""
-    if SUPPORTS_MEM_PROTECTION:
+    if sys.implementation.name == 'pypy':
+        assert b"RPython" in result.stderr
+    elif SUPPORTS_MEM_PROTECTION:
         assert result.stderr == b""
     else:
         # The garbage we override the data with will cause this error
@@ -103,7 +108,10 @@ def test_charptr_write_ptr(compiler, python_subprocess):
         @INIT
     """)
     result = python_subprocess.run(mod, "mod.f('try writing me!');")
-    assert result.returncode != 0
+    if sys.implementation.name == 'pypy':
+        assert result.returncode == 0
+    else:
+        assert result.returncode != 0
     assert result.stdout == b""
     assert result.stderr == b""
 
--- a/test/hpy_devel/test_distutils.py
+++ b/test/hpy_devel/test_distutils.py
@@ -90,7 +90,7 @@ class TestDistutils:
             proc = subprocess.run(cmd)
             out = None
         if proc.returncode != 0:
-            raise Exception(f"Command {exe} failed")
+            raise Exception(f"Command {cmd} failed")
         return out
 
 
--- a/test/test_slots.py
+++ b/test/test_slots.py
@@ -354,6 +354,7 @@ class TestSlots(HPyTest):
     def test_buffer(self):
         import pytest
         import sys
+        import gc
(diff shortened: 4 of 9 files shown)


hpy/devel/include/hpy/hpydef.h
```diff
@@ -10,7 +10,7 @@ extern "C" {
 #include "hpy/autogen_hpyslot.h"
 #include "hpy/cpy_types.h"
 
-typedef void* (*HPyCFunction)();
+typedef void* (*HPyCFunction)(void);
```
mattip, line 13: Am I using a different compiler to everyone else? My default compiler (gcc 9.4) complained about this ...

test/debug/test_charptr.py
```diff
@@ -1,5 +1,6 @@
-from test.support import SUPPORTS_SYS_EXECUTABLE
+from ..support import SUPPORTS_SYS_EXECUTABLE
```
mattip, line 1: The relative import needs the empty `__init__.py`

test/debug/test_charptr.py
```diff
@@ -235,4 +243,4 @@ def clear_raw_data_in_closed_handles():
             assert h.raw_data_size == -1
     finally:
         _debug.set_protected_raw_data_max_size(old_raw_data_max_size)
-        _debug.set_closed_handles_queue_max_size(old_closed_handles_max_size)
\ No newline at end of file
+        _debug.set_closed_handles_queue_max_size(old_closed_handles_max_size)
```
mattip, line 246: I am not sure what my editor did here: was an empty line missing?

test/hpy_devel/test_distutils.py
```diff
@@ -90,7 +90,7 @@ def python(self, *args, capture=False):
             proc = subprocess.run(cmd)
             out = None
         if proc.returncode != 0:
-            raise Exception(f"Command {exe} failed")
+            raise Exception(f"Command {cmd} failed")
```
mattip, line 93: This is a fix for something not connected to the PR. I don't understand why these tests are failing.

test/debug/test_charptr.py
```diff
@@ -46,7 +47,9 @@ def test_charptr_use_after_implicit_arg_handle_close(compiler, python_subprocess
     result = python_subprocess.run(mod, code)
     assert result.returncode != 0
     assert result.stdout == b""
-    if SUPPORTS_MEM_PROTECTION:
+    if sys.implementation.name == 'pypy':
+        assert b"RPython" in result.stderr
+    elif SUPPORTS_MEM_PROTECTION:
```
steve-s, line 52: Out of curiosity: so what does this test give on PyPy?

hpy/devel/include/hpy/hpydef.h
fangerer: Could be (I'm on Mac and compiling with LLVM).
However, that is semantically not the same: specifying `void` really means that the function doesn't have arguments while leaving the argument list empty means that it is not specified at this point. I hope that doesn't make problems elsewhere.

fangerer (approved):
We just discussed that PR at the HPy sprint and we were unsure about the changes in `test_charptr.py`. However, we (Antonio and I) think this is fine for now but will change the debug mode a bit to be more specific if mem protection is not available.

test/debug/test_charptr.py
ambv: Yup.

test/hpy_devel/test_distutils.py
ambv: Fixed in master.

test/debug/test_charptr.py
mattip: It seems to be different as I change things to support hpy-0.0.4, so this line may need tweeking moving forward.

hpy/devel/include/hpy/hpydef.h
mattip: The function declaration should match the function definition. Why can't we specify the argument list when we know what it is meant to be?

mattip:
The `test_charptr.py` tests crash on PyPy and on CPython if there is no mem protection. I don't think the exact error message is important or even guaranteed to be reproducible since the code is writing to arbitrary memory.

fangerer:
> The `test_charptr.py` tests crash on PyPy and on CPython if there is no mem protection. I don't think the exact error message is important or even guaranteed to be reproducible since the code is writing to arbitrary memory.

They crash? I think that's unexpected. Do you know why? 
If we don't have mem protection, we write some carefully picked bytes into the buffer (i.e. the *poison* which is `char poison[] = {0xBA, 0xD0, 0xDA, 0x7A};`) that will prevent any useful usage in a unicode object.

mattip:
I think PyPy does not like the random unicode string.
On the hpy-0.0.4 branch, PyPy exits with 1 and prints 
```
b'RPython traceback:
  File "pypy_interpreter.c", line 48519, in BuiltinCodePassThroughArguments1_funcrun_obj
  File "pypy_module__hpy_universal.c", line 18123, in W_ExtensionFunction_d_descr_call
  File "pypy_module__hpy_universal.c", line 18936, in W_ExtensionFunction_d_call_noargs
  File "pypy_module__hpy_universal.c", line 49901, in ctx_HPyUnicode_FromString__star_2
  File "pypy_module__hpy_universal.c", line 59473, in HPyUnicode_FromString
  File "pypy_module__hpy_universal_1.c", line 17855, in _maybe_utf8_to_w
  File "rpython_rlib.c", line 5324, in check_utf8
Traceback (most recent call last):
  File "<string>", line 1, in <module>
SystemError: unexpected internal exception (please report a bug): <CheckError object at 0x7f2e5210f640>; internal traceback was dumped to stderr'
```

CPython exits with -11 and prints `b''`.

test/test_slots.py
```diff
@@ -419,6 +420,7 @@ def test_buffer(self):
                 assert sys.getrefcount(arr) == init_refcount + 1
             for i in range(12):
                 assert mv[i] == i
+        gc.collect()
```
mattip, line 423: @hodgestar Note I needed to add this in 0.4 as well

mattip:
Closing in favor of #455

PR #346, opened 2022-09-19
Support stacktrace on macOS

ambv:
Turns out Darwin ships `backtrace()` since 10.5.

This change also makes HPy require setuptools 64.0.0+ and removes a flakiness of parallel (xdist) tests creating and deleting temporary wheels in the same directory.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,7 @@ jobs:
 
       - name: Run tests
         run: |
-          python -m pip install pytest pytest-xdist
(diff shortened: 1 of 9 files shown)


ambv:
The last failure here is a Windows-specific one:

```
C:\Users\runneradmin\AppData\
   Local\Temp\pytest-of-runneradmin\pytest-0\
   popen-gw1\test_hpymod_build_platlib0\venv\lib\
   site-packages\hpy\devel\src\
   runtime\argparse.c : fatal error C1083: Cannot open compiler generated file: '': Invalid argument
```

PR #347, opened 2022-09-19
Support object members

timfel:
We already define `HPyMember_OBJECT`, but using it in
```c
HPyDef_MEMBER(Foo_OBJECT_member, "OBJECT_member", HPyMember_OBJECT, offsetof(FooObject, OBJECT_member))
```
fails in universal mode, as the included test demonstrates, since the embedded member must an `HPyField` in this mode. This should work, since it's annoying to have to define getters/setters manually here.

### DIFF
--- a/hpy/devel/src/runtime/ctx_type.c
+++ b/hpy/devel/src/runtime/ctx_type.c
@@ -361,7 +361,7 @@ create_member_defs(HPyDef *hpydefs[], PyMemberDef *legacy_members, HPy_ssize_t b
                 while ((*getsets)[getsetcnt].name) {
                     getsetcnt++;
                 }
-                *getsets = (PyGetSetDef*)PyMem_Realloc(*getsets, (getsetcnt + 1) * sizeof(PyGetSetDef));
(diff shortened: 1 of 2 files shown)


hodgestar:
What happens during porting a C extension? Presumably initially the member is a `PyObject *`. Will compilation in universal mode then fail? And once the member is been made an `HPyField`, will compilation work in both CPython and universal mode?

I am wondering whether we need a legacy_defines or similar? Or at least some way to make it clear "these object members should use HPyField and those should use PyObject*".

timfel:
> What happens during porting a C extension? Presumably initially the member is a `PyObject *`. Will compilation in universal mode then fail? And once the member is been made an `HPyField`, will compilation work in both CPython and universal mode?

As long as it is a PyObject*, compilation in universal mode just fails. After it is an HPyField, it works in both CPython C ABI mode and HPy universal mode. It even works to have an HPyField and still use the legacy definition for members, since then you have to compile in C ABI mode and the HPyField is binary equivalent to PyObject* in that mode. So afaics, nothing can go wrong even if you just partially port.

PR #351, opened 2022-09-21
Use Python 3.10 for g++ tests, POC tests and documentation tests

ambv:
Also, stop ignoring failures for all tests except for main (as Windows is currently failing).


ambv:
As discussed with @antocuni at the sprint, I'm landing this.

PR #356, opened 2022-09-22
Add HPy ABI version in the ABI and check it when loading a module

### LINKED ISSUE #374: Add HPy ABI version in the ABI and check it when loading a module
This should provide the ability to eventually break the ABI of `HPyContext` and other structs that are part of ABI (`HPyMethodDef`, etc.), the only thing that will be set in stone forever will be that the extension exposes two symbols with minor and major HPy ABI version.

steve-s:
Fixes #374 

The loader now looks up two new symbols when loading an extension ([link to the code](https://github.com/hpyproject/hpy/pull/356#discussion_r1036943570)):
* `required_hpy_major_version_##modname`
* `required_hpy_minor_version_##modname`

These capture the HPy ABI versions of the HPy that was used to compile the module. The HPy that is loading the module is going to check them and:
* create the right version/layout of `HPyContext` for the ABI major version or fail if given major version is not supported
* check that it can support the requested minor version

The idea is that major ABI versions have different layout/semantics. Minor ABI versions are only ABI compatible additions at the end of the `HPyContext` structure (and other structures that are part of the ABI). The changes in major versions are not expected for now, but this makes the ABI future proof for forward compatibility. 

This feature duplicates the ABI version tag added in https://github.com/hpyproject/hpy/pull/371

Note: the ABI version is not meant to be used by humans. I suggest we still use `HPY_VERSION` macro that provides human friendly version. Maybe we can/should enforce some consistency between the two.

### DIFF
--- a/hpy/devel/include/hpy/hpymodule.h
+++ b/hpy/devel/include/hpy/hpymodule.h
@@ -40,25 +40,29 @@ typedef struct {
 #ifdef HPY_ABI_CPYTHON
 
 // module initialization in the CPython case
-#define HPy_MODINIT(modname)                                   \
-    static HPy init_##modname##_impl(HPyContext *ctx);         \
-    PyMODINIT_FUNC                                             \
-    PyInit_##modname(void)                                     \
-    {                                                          \
-        return _h2py(init_##modname##_impl(_HPyGetContext())); \
+#define HPy_MODINIT(modname)                                      \
+    static HPy init_##modname##_impl(HPyContext *ctx);            \
+    PyMODINIT_FUNC                                                \
+    PyInit_##modname(void)                                        \
+    {                                                             \
+        return _h2py(init_##modname##_impl(_HPyGetContext()));    \
     }
 
 #else // HPY_ABI_CPYTHON
 
 // module initialization in the universal and hybrid case
-#define HPy_MODINIT(modname)                                   \
-    _HPy_CTX_MODIFIER HPyContext *_ctx_for_trampolines;        \
-    static HPy init_##modname##_impl(HPyContext *ctx);         \
-    HPyMODINIT_FUNC                                            \
-    HPyInit_##modname(HPyContext *ctx)                         \
-    {                                                          \
-        _ctx_for_trampolines = ctx;                            \
-        return init_##modname##_impl(ctx);                     \
+#define HPy_MODINIT(modname)                                      \
+    HPy_EXPORTED_SYMBOL uint32_t                                  \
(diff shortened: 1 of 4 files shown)


timfel:
This looks fine to me.

fangerer:
> The idea is that major ABI versions have different layout/semantics. Minor ABI versions are only ABI compatible additions at the end of the `HPyContext` structure. The changes in minor versions are not expected for now, but this makes the ABI future proof for forward compatibility.

This reads like a compatibility policy. IIRC, we discussed that we should define something like this. We could just formalize that and write it to the readme or `public_api.h` or in some other appropriate place.

hpy/devel/include/hpy/hpymodule.h
```diff
@@ -34,14 +34,18 @@ typedef struct {
...
-    HPyMODINIT_FUNC                                         \
-    HPyInit_##modname(HPyContext *ctx)                     \
-    {                                                          \
-        _ctx_for_trampolines = ctx;                            \
-        return init_##modname##_impl(ctx);                     \
+#define HPy_MODINIT(modname)                                      \
+    Py_EXPORTED_SYMBOL uint32_t                                   \
+    required_hpy_major_version_##modname = HPY_ABI_VERSION_MAJOR; \
+    Py_EXPORTED_SYMBOL uint32_t                                   \
+    required_hpy_minor_version_##modname = HPY_ABI_VERSION_MINOR; \
```
fangerer, line 41: If I understood it correctly, those variables are read-only, right? In that case, I think they should be `const`.

steve-s: fixed

steve-s: Hm, `const` has a different semantics in C++ and gives compiler error in this context. We could use `#if defined(__cplusplus)`, but I was reading through the docs on what `const` actually means and implies and now it seems safer and easier to me to just not use it at all.

hpy/universal/src/hpymodule.c
```diff
@@ -105,6 +106,38 @@ static PyObject *do_load(PyObject *name_unicode, PyObject *path, int debug)
         goto error;
     }
 
+    char minor_version_symbol_name[258];
```
steve-s, line 109: This is the place where we load and check the ABI versions. We do not check the ABI tag here. We can get the ABI tag from the filename and use that instead or alongside of the symbols that this PR adds.

steve-s:
I am closing in favor of opening a new fresh discussion. This was stale for too long and some important things changed since then.

PR #361, opened 2022-09-23
Fix microbenchmarks

antocuni:
Microbenchmarks got slightly out of date and didn't build, fix it and improve the readme

### DIFF
--- a/microbench/src/hpy_simple.c
+++ b/microbench/src/hpy_simple.c
@@ -56,9 +56,7 @@ static HPy allocate_int_impl(HPyContext *ctx, HPy self)
 HPyDef_METH(allocate_tuple, "allocate_tuple", HPyFunc_NOARGS)
 static HPy allocate_tuple_impl(HPyContext *ctx, HPy self)
 {
-    //return Py_BuildValue("ii", 2048, 2049);
-    HPyErr_SetString(ctx, ctx->h_Exception, "HPy_BuildValue not implemented yet");
-    return HPy_NULL;
+    return HPy_BuildValue(ctx, "ii", 2048, 2049);
 }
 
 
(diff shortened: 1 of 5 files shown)


hodgestar (approved):

hodgestar:
Should we add running the microbenchmarks to CI so that they don't break in future?

microbench/src/hpy_simple.c
```diff
@@ -65,7 +65,8 @@ static HPy allocate_tuple_impl(HPyContext *ctx, HPy self)
 /* Foo type */
 
 typedef struct {
-    PyObject_HEAD
+    long x;
+    long y;
 } FooObject;
```
fangerer, line 70: Why do we even need `FooObject`? An HPy type can also have `.basicsize = 0`.

fangerer (approved):
I think this micro benchmark could use a further update.
Function `allocate_tuple_impl` says:
```
HPyErr_SetString(ctx, ctx->h_Exception, "HPy_BuildValue not implemented yet");
```
but that's no longer true. We already have it.
But I'm already fine with merging that.

microbench/src/hpy_simple.c
antocuni: it's because we wanted to have non-0 allocations in `test_allocate_*`, they are more likely to be representative.

antocuni:
> I think this micro benchmark could use a further update. Function `allocate_tuple_impl` says:
> 
> ```
> HPyErr_SetString(ctx, ctx->h_Exception, "HPy_BuildValue not implemented yet");
> ```
> 
> but that's no longer true. We already have it. But I'm already fine with merging that.

ah good point. I'll try to fix that

antocuni:
> Should we add running the microbenchmarks to CI so that they don't break in future?

yes, good idea

PR #350, opened 2022-09-21
debug mode context check

NiRit100:
[steve-s](https://github.com/steve-s) worked with me on this with during the PyPy sprint in Düsseldorf.
It's work in progress, just a semi-working hacky test.

PR #363, opened 2022-09-23
Debug mode: check correct HPyContext usage

steve-s:
Based on WIP in https://github.com/hpyproject/hpy/pull/350 done with @NiRit100 at the HPy sprint

* the private data of debug context now contains a flag `is_valid` and all the context functions check this flag as the first thing.
* the private data of debug context is split to "shared" part and the `is_valid` flag
* we have a circular buffer of multiple copies of the debug context
* the trampoline (`debug_ctx_CallRealFunctionFromTrampoline`) takes next context from the buffer, sets `is_valid=false` for the previous context and `is_valid=true` for the context to be used (after the call the flags are reverted back)

### DIFF
--- a/hpy/debug/src/debug_ctx_cpython.c
+++ b/hpy/debug/src/debug_ctx_cpython.c
@@ -71,6 +71,24 @@ static void _buffer_py2h(HPyContext *dctx, const Py_buffer *src, HPy_buffer *des
     dest->internal = src->internal;
 }
 
+static HPyContext* _switch_to_next_dctx_from_cache(HPyContext *current_dctx) {
+    HPyContext *next_dctx = hpy_debug_get_next_dctx_from_cache(current_dctx);
+    if (next_dctx == NULL) {
+        HPyErr_NoMemory(current_dctx);
+        get_ctx_info(current_dctx)->is_valid = false;
+        get_ctx_info(next_dctx)->is_valid = true;
+        return NULL;
+    }
+    get_ctx_info(current_dctx)->is_valid = false;
+    get_ctx_info(next_dctx)->is_valid = true;
+    return next_dctx;
+}
+
+static void _switch_back_to_original_dctx(HPyContext *original_dctx, HPyContext *next_dctx) {
+    get_ctx_info(next_dctx)->is_valid = false;
+    get_ctx_info(original_dctx)->is_valid = true;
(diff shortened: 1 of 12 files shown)


steve-s:
I've rebased the PR and it's ready for a review now.

hpy/debug/src/debug_ctx_cpython.c
```diff
@@ -78,8 +78,24 @@ void debug_ctx_CallRealFunctionFromTrampoline(HPyContext *dctx,
         HPyFunc_noargs f = (HPyFunc_noargs)func;
         _HPyFunc_args_NOARGS *a = (_HPyFunc_args_NOARGS*)args;
         DHPy dh_self = _py2dh(dctx, a->self);
-        DHPy dh_result = f(dctx, dh_self);
-        DHPy_close_and_check(dctx, dh_self);
+
+        HPyContext *next_dctx = hpy_debug_get_next_dctx_from_cache(dctx);
```
fangerer, line 82: That's a lot of code duplication. I think we could pull the _temporary debug context setup and tear down code_ out of the switch if we use `break` instead of `return` at the end of the cases.
We then only need to use `next_dctx` for opening and closing the argument handles like `dh_self`.
Or do I miss something?

fangerer (commented):
Besides my concern about code duplication (see comment below), looks like a very good debug mode feature.

hpy/debug/src/debug_ctx_cpython.c
steve-s: Fixed. We cannot really pull out the whole thing, because in `a->result` variable `a` is a different struct every time, but I pulled as much as I could. Also used the new helpers in the autogenerated code.

fangerer: lgtm

fangerer (approved):

steve-s:
@antocuni would you have some spare cycles for a review :-)?

steve-s:
Proactively rebased on top of https://github.com/hpyproject/hpy/pull/402 to streamline merging of both PRs

mattip:
Could you describe under what cases the checks might fail? What is the scenario that would cause someone to get the wrong context?

mattip:
Thanks @steve-s

PR #364, opened 2022-09-30
Add support for native subclasses of built-in types.

fangerer:
This may deserve some explanation:
Until now, it was already possible to inherit from built-in classes **IF** you had `basicsize == 0`. Because in that case, it would just inherit everything from the base class.
Before this change, if you have `basicsize > 0`, the only way to do that was by using a legacy type like this:
```
typedef struct {
    PyLongObject ob_base;
    int member0;
    // ...
} SomeObject;
```

However, since `PyLongObject` is actually an internal data structure and is opaque on non-CPythons, we need to provide a different approach.

After several discussion with Antonio we (for now) settled on the term `shape`.

This PR now introduces enum `HPyType_BuiltinShape` which abstracts the internal _shape_ of built-in types. It can be seen as a generalization over the `legacy` approach.
So, pure HPy types now have `HPyType_BuiltinShape_(Object|Type|Long|Float|Unicode|Tuple|List)` and legacy types have shape `HPyType_BuiltinShape_Legacy`.

The main usage for the shape is then to compute the offset for the HPy type's native data pointer (statically). So, at runtime, we will still just do something like `((char *)obj) + HEADER_OFFSET` when calling `HPy_AsStruct_*`. This was the main reason for this solution.

Additionally, we are storing the header offset into the HPy type's extra info. That enables us to have a generic but slower `HPy_AsStruct_Slow` that will always pick the right header offset from the extra info.
We also check (in debug mode) if the right `HPy_AsStruct_*` function is used by comparing to the result of `HPy_AsStruct_Slow`.

**Still missing in this PR** (I'll add that after the first review round):

* I was up to write documentation about that in `hpy/docs/api.rst` and realized that we did not document how to create HPy types at all. So, I started to write that but before I spend more time on it, I'll wait for the reviews since things may change.
* More tests (in particular: HPy types inheriting from types with the non-default shape).


hpy/devel/src/runtime/ctx_type.c
```diff
@@ -931,17 +1023,60 @@ ctx_Type_GenericNew(HPyContext *ctx, HPy h_type, HPy *args, HPy_ssize_t nargs, H
...
-ctx_AsStructLegacy(HPyContext *ctx, HPy h)
+ctx_AsStruct_Legacy(HPyContext *ctx, HPy h)
 {
     return _h2py(h);
 }
 
+_HPy_HIDDEN void*
+ctx_AsStruct_Type(HPyContext *ctx, HPy h)
+{
+    return _HPy_Payload(_h2py(h), HPyType_BuiltinShape_Type);
```
fangerer, line 1040: btw. I verified (by disassembling in the debugger) that those functions really just do the offset computation if compiled with `-O3`:
```
(lldb) di -f
mytest.cpython-39-darwin.so`ctx_AsStruct_Object:
    0x1073497a0 <+0>: pushq  %rbp
    0x1073497a1 <+1>: movq   %rsp, %rbp
->  0x1073497a4 <+4>: leaq   0x10(%rsi), %rax
    0x1073497a8 <+8>: popq   %rbp
    0x1073497a9 <+9>: retq
    ```

File: hpy/tools/autogen/public_api.h [hunk 2]
```diff
@@ -311,9 +311,23 @@ HPy_ID(168)
...
+HPy_ID(231)
+void* _HPy_AsStruct_Float(HPyContext *ctx, HPy h);
+HPy_ID(232)
+void* _HPy_AsStruct_Unicode(HPyContext *ctx, HPy h);
+HPy_ID(233)
+void* _HPy_AsStruct_Tuple(HPyContext *ctx, HPy h);
+HPy_ID(234)
+void* _HPy_AsStruct_List(HPyContext *ctx, HPy h);
+HPy_ID(235)
+void* _HPy_AsStruct_Slow(HPyContext *ctx, HPy h);
```
steve-s, line 330: Do we have a use-case for `_HPy_AsStruct_Slow`? If not then maybe we should not expose it yet?

steve-s: Hm, the debug context needs it. I see. The debug context does not have an access to the "internals" of the universal context it wraps, so we need to expose something for it. Ok, then we probably have no other option.

TeamSpen210:
Could this also be used for exception types? Most of them just have the same layout as `BaseException`, so only a few shapes would need to be added for things like `UnicodeError`.

hpy/tools/autogen/public_api.h
antocuni: yes, good point. So far we need something similar here, which is called by tp_traverse and tp_destroy.
https://github.com/hpyproject/hpy/blob/79f30c03cfaa4fdc3fa660b936a7ac172c3572aa/hpy/devel/src/runtime/ctx_type.c#L73-L81

+1 for keeping it private as long as there is no real need outside.

FWIW I just realized that in theory we could avoid it altogether by writing several versions of tp_traverse and tp_destroy, one for each shape. Not sure if it's worth it though (probably not)

hpy/debug/src/debug_ctx.c
```diff
@@ -181,6 +181,36 @@ DHPy debug_ctx_Type_FromSpec(HPyContext *dctx, HPyType_Spec *spec, HPyType_SpecP
     return DHPy_open(dctx, HPyType_FromSpec(get_info(dctx)->uctx, spec, NULL));
 }
 
+#define DEBUG_AS_STRUCT(SHAPE) \
```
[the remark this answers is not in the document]
antocuni: just a suggestion, but when I started reading it it took me a while to realize what was it for. I think that the new name makes it clearer

hpy/debug/src/debug_ctx.c
```diff
@@ -181,6 +181,36 @@ DHPy debug_ctx_Type_FromSpec(HPyContext *dctx, HPyType_Spec *spec, HPyType_SpecP
...
+    void *debug_ctx_AsStruct_##SHAPE(HPyContext *dctx, DHPy dh) \
+    { \
+        HPyContext *uctx = get_info(dctx)->uctx; \
+        UHPy uh = DHPy_unwrap(dctx, dh); \
+        void *actual_data_ptr = _HPy_AsStruct_##SHAPE(uctx, uh); \
+        void *expected_data_ptr = _HPy_AsStruct_Slow(uctx, uh); \
+        if (actual_data_ptr != expected_data_ptr) { \
+            HPy_FatalError(uctx, "Invalid usage of _HPy_AsStruct_" #SHAPE \
+                ". The object's type has a different shape."); \
+        } \
```
antocuni, line 194: maybe we can be even more precise? With this logic, if two shapes happen to have the size `basicsize`, this check would pass. So ideally I would like to check that the actual shape stored in `extra->shape` matches the one which we expect.
I don't know how hard it is to access `extra->shape` from here though (if it's too hard, I'm fine with leaving this logic, but maybe with a comment which explains why it's sub-optimal and why it's too hard to do the right thing)

hpy/devel/include/hpy/hpytype.h
```diff
@@ -8,36 +8,45 @@
...
-       Legacy types are available to allow gradual porting of existing CPython
-       extensions.
-
-       A type with .legacy_slots not NULL is required to have .legacy = true and to
-       include PyObject_HEAD at the start of its struct. It would be easy to
-       relax this requirement on CPython (where the PyObject_HEAD fields are
-       always present) but a large burden on other implementations (e.g. PyPy,
-       GraalPython) where a struct starting with PyObject_HEAD might not exist.
+       A type whose struct starts with PyObject_HEAD is a legacy type. A legacy
+       type must set .builtin_shape = HPyType_BuiltinShape_Legacy in its
```
antocuni, line 29: `A type whose struct starts with PyObject_HEAD is a legacy type.`

this is not strictly correct. I think that after this PR we can write something like this, can't we:
```c
struct MyMetaClass {
    PyTypeObject base;
    ...
};
```
i.e. it is a legacy type even if it doesn't literally start with `PyObject_HEAD`.
One quick fix could be "a type whose struct starts with PyObject_HEAD or equivalent is a legacy type"?
Do we have any better wording?

hpy/devel/include/hpy/hpytype.h
```diff
@@ -76,43 +85,66 @@ typedef struct {
...
-#define _HPyType_GENERIC_HELPERS(TYPE, CAST, IS_LEGACY)              \
-                                                                     \
-enum { TYPE##_IS_LEGACY = IS_LEGACY };                               \
-                                                                     \
-HPyAPI_UNUSED TYPE *                                                 \
-TYPE##_AsStruct(HPyContext *ctx, HPy h)                              \
-{                                                                    \
-    return (TYPE*) CAST(ctx, h);                                     \
+    _HPyType_GENERIC_HELPERS(TYPE, HPyType_BuiltinShape_Legacy)
+
```
antocuni, line 129: I wonder whether we can improve/simplify the usage and have only `HPyType_HELPERS`, whose first param is optional. I.e.:
```c
HPyType_HELPERS(PointObject); // common case
HPyType_HELPERS(PointObject, HPyType_BuiltinShape_Object); // equivalent to the above
HPyType_HELPERS(PointObject, HPyType_BuiltinShape_Legacy);
HPyType_HELPERS(PointObject, HPyType_BuiltinShape_Tuple);
```

hpy/devel/src/runtime/ctx_type.c
```diff
@@ -788,32 +873,31 @@ ctx_Type_FromSpec(HPyContext *ctx, HPyType_Spec *hpyspec,
         return HPy_NULL;
     }
     int basicsize;
-    HPy_ssize_t base_member_offset;
+    int base_member_offset;
```
antocuni, line 876: why `int`?

hpy/devel/include/hpy/hpytype.h
```diff
@@ -8,36 +8,45 @@
 #define HPyAPI_UNUSED static inline
 #endif /* __GNUC__ */
 
+typedef enum {
```
antocuni, line 11:
```suggestion
// NOTE: HPyType_BuiltinShape_Object == 0, which means it's the default
// if it's not specified in HPyType_spec
typedef enum {
```

hpy/devel/src/runtime/ctx_type.c
```diff
@@ -788,32 +873,31 @@ ctx_Type_FromSpec(HPyContext *ctx, HPyType_Spec *hpyspec,
...
 
-    bool is_pure;
-    if (hpyspec->legacy != 0) {
-        basicsize = hpyspec->basicsize;
-        base_member_offset = 0;
-        is_pure = false;
+    int head_size = _HPy_GetHeaderSize(hpyspec->builtin_shape);
+    if (head_size < 0) {
+        // an invalid shape was specified
+        PyErr_Format(PyExc_ValueError, "invalid shape: %d", hpyspec->builtin_shape);
```
antocuni, line 882:
```suggestion
        PyErr_Format(PyExc_ValueError, "invalid .builtin_shape: %d", hpyspec->builtin_shape);
```

hpy/devel/src/runtime/ctx_type.c
```diff
@@ -822,7 +906,7 @@ ctx_Type_FromSpec(HPyContext *ctx, HPyType_Spec *hpyspec,
     spec->basicsize = basicsize;
     spec->flags = flags | HPy_TPFLAGS_INTERNAL_IS_HPY_TYPE;
     spec->itemsize = hpyspec->itemsize;
-    spec->slots = create_slot_defs(hpyspec, base_member_offset, extra);
+    spec->slots = create_slot_defs(hpyspec, (HPy_ssize_t)base_member_offset, extra);
```
antocuni, line 909: same question as above, I think the cast here is because you turned it into an `int`

test/debug/test_as_struct.py
```diff
@@ -0,0 +1,40 @@
...
+
+        @EXPORT_TYPE("Dummy", Dummy_spec)
+        @EXPORT(f)
+        @INIT
+    """)
+    code = "assert mod.f(mod.Dummy()) == 0"
+    result = python_subprocess.run(mod, code)
+    assert result.returncode != 0
+    assert "Invalid usage of _HPy_AsStruct_Object" in result.stderr.decode("utf-8")
+
```
antocuni, line 39: Brilliant test, I love it 😍

antocuni:
> Could this also be used for exception types? Most of them just have the same layout as `BaseException`, so only a few shapes would need to be added for things like `UnicodeError`.

Good point. I don't remember how I and @fangerer came up with the initial list of builtin shapes, but it seems that exceptions should be included as well.

hpy/devel/include/hpy/hpytype.h
fangerer: I think it is possible and I tried to do that but it became unreadable so I gave up on it.
The problem is: we cannot have simply two macro definitions like:
```
#define HPyType_HELPERS(TYPE)
#define HPyType_HELPERS(TYPE, SHAPE)
```
So, we need to do something like:
```
#define HPyType_HELPERS(TYPE, ...)
```
If you do it this way, you quickly run into problems if `__VA_ARGS__` is empty. In particular, if you pass them on and have a preceding comma like:
```
DO_SOMETHING(TYPE, __VA_ARGS__)
```
So, you need to do
```
#define HPyType_HELPERS(...)
```
such that `__VA_ARGS__` is never empty but then selecting the single arguments out of the `__VA_ARGS__` is nasty.

Long story short: it is possible (I think) but the macro definition becomes ugly.
I'll give it another try ...

hpy/devel/src/runtime/ctx_type.c
fangerer: I switched to `int` mostly because `HPyType_Spec.basicsize` also is an `int`. But I can use `HPy_ssize_t` here everywhere. This is, for sure, safer.

fangerer: Ah, no, I now remember:
If we use `HPy_ssize_t` for `basicsize` and `base_member_offset`, we need to cast from `HPy_ssize_t` to `int` at some point which is unsafe. In this way, we just need to cast from `int` to `HPy_ssize_t` which should be safe.

hpy/devel/src/runtime/ctx_type.c
fangerer: see: https://github.com/hpyproject/hpy/pull/364#discussion_r989106377

hpy/debug/src/debug_ctx.c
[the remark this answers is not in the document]
fangerer: done

hpy/debug/src/debug_ctx.c
fangerer: Okay, I've removed `_HPyType_AsStruct_Slow` and introduced `_HPyType_GetBuiltinShape`. This is similar to CPython's function `PyObject_GetTypeDataSize` as proposed in [PEP-697](https://peps.python.org/pep-0697/).
We could also add some `_HPyType_AsStruct_FromShape(HPyContext *, HPy, HPyType_BuiltinShape)` if we need it (but that's not the case so far).
The debug mode now uses that and can do an accurate check.

hpy/devel/include/hpy/hpytype.h
fangerer: done

hpy/devel/include/hpy/hpytype.h
fangerer: As promised, I gave it another try and it works well.
The only drawback is that we need to expose more helper macros that like `_HPyType_HELPER_DEFINE_SHAPE`. But I'm not opposed to that.

That means: I've removed `HPyType_BUILTIN_HELPERS` and you can do:
```
HPyType_HELPERS(SomeObject)
HPyType_HELPERS(SomeObject, HPyType_BuiltinShape_Something)
```

hpy/devel/src/runtime/ctx_type.c
fangerer: done

hpy/devel/src/runtime/ctx_type.c
antocuni: > we need to cast from HPy_ssize_t to int at some point which is unsafe. 

I guess that this "some point' is inside the CPython implementation of HPy, i.e. when you need to fill CPython's `PyType_Spec.basicsize`, is it correct?

If so, I don't think it is unsaf(er) than what we have now.
In C, the return type of `offsetof` and `sizeof` is `size_t`, so potentially they might not fit into an `int` (veeeeery potentially, I admit 😅). So, in one case the "unsafe cast" between `size_t` and `int` is done by the user, in the other it's done by us to comply with a (broken, IMHO) CPython API.

So, I think I would prefer to do the "clean & correct" thing on the user-facing side and to the "unsafe necessary cast" as an implementation detail (and possibly also throw `HPy_FatalError` in case it doesn't fit).

That said, I don't think it changes anything in practice at the moment of course (it works on CPython, so it's battle-tested by thousands of extensions and many compilers)

hpy/debug/src/debug_ctx.c
```diff
@@ -181,35 +181,64 @@ DHPy debug_ctx_Type_FromSpec(HPyContext *dctx, HPyType_Spec *spec, HPyType_SpecP
...
 
-#define DEBUG_AS_STRUCT(SHAPE) \
+static const char *get_builtin_shape_name(HPyType_BuiltinShape shape)
+{
+#define SHAPE_NAME(SHAPE) \
+    case SHAPE:           \
+        return #SHAPE;    \
+
+    /* Note: we use a switch here because then the compiler will warn us about
+       missing cases if shapes are added to enum 'HPyType_BuiltinShape' */
```
antocuni, line 191: brilliant! I didn't know this trick :)

antocuni (approved):
thank you, I like it! :)

hpy/devel/include/hpy/hpytype.h
```diff
@@ -76,43 +88,68 @@ typedef struct {
...
      is associated with an object that is not an instance of PointObject.
+     However, debug mode will catch an incorrect usage.
 
-   * `PointObject_IS_LEGACY`: an enum value set to 0 so that in the
-     HPyType_Spec for PointObject one can write
-     `.legacy = PointObject_IS_LEGACY` and not have to remember to
-     update the spec when the helpers used changes.
+   * `SHAPE(PointObject)`: a macro that is meant to be used as static
+     initializer in the corresponding HPyType_Spec. It is recommended to write
+     `.builtin_shape = SHPAE(PointObject)` such that you don't have to remember
```
hodgestar, line 109:
```suggestion
     `.builtin_shape = SHAPE(PointObject)` such that you don't have to remember
```

fangerer: btw. the reason why I now introduced `SHAPE(...)` is because (1) `g++` doesn't like if we assign an anonymous enum to `HPyType_BuiltinShape` and (2) because `MSVC` doesn't like `static const HPyType_BuiltinShape var` in static initializers.

hpy/devel/src/runtime/ctx_type.c
fangerer: Okay, you convinced me. I switched back to `HPy_ssize_t`.

hpy/devel/__init__.py
```diff
@@ -295,6 +296,12 @@ def _finalize_hpy_ext(self, ext):
         ext.include_dirs += self.hpydevel.get_extra_include_dirs()
         ext.sources += self.hpydevel.get_extra_sources()
         ext.define_macros.append(('HPY', None))
+        if self.compiler == 'msvc' or get_default_compiler() == 'msvc':
+            # If using MSVC, we need to enable the new preprocessor otherwise
+            # our variable macros (in particular HPyType_HELPERS) won't
+            # properly work. See also:
+            # https://learn.microsoft.com/en-us/cpp/preprocessor/preprocessor-experimental-overview?view=msvc-170
+            ext.extra_compile_args.append('/Zc:preprocessor')
```
fangerer, line 304: @antocuni @hodgestar Before I merge, I ask you to have a look on this.
The MSVC preprocessor isn't conforming (by default) to the C standard. This hurts us in the variadic macro `HPyType_HELPERS`. Consider this example:
```
#define BAR(A, B, ...) A
#define FOO(...) BAR(__VA_ARGS__, "default")

FOO("hello", "world")
```
Result on MSVC: `"hello", "world"`
Result on GCC/LLVM: `"hello"`
MSVC puts the whole `__VA_ARGS__` into the first arg of the called function-like macro.

This is basically a shortcoming in MSVC which can be solved by using argument `/Zc:preprocessor`.
So, we need to pass this argument to all HPy extensions but only for compiler `msvc`.

I used condition `self.compiler == 'msvc' or get_default_compiler() == 'msvc'` because if you specify `--compiler=something`, this will end up in `self.compiler`. If nothing is specifed, the default compiler will be used.

Do you have a better idea how to do that?

mattip: How do others work around this? For instance, there is `cpython/Python/getargs.c` which uses varargs but does not special-case MSVC

fangerer: > How do others work around this?

I don't know. It depends on how exactly you use it.

> For instance, there is cpython/Python/getargs.c which uses varargs but does not special-case MSVC

I just grepped through the CPython code base for `__VA_ARGS__` and didn't find a single usage. I think what you were referring to are the C functions using varargs. But that's completely different.

fangerer: Ooops, I used the wrong search scope: Anyway, I still just found very few usages like this in `Objects/stringlib/fastsearch.h`
```
# define LOG(...) printf(__VA_ARGS__)
```
In this case, it doesn't matter since the `__VA_ARGS__` are just passed on as a whole.

mattip: https://stackoverflow.com/questions/32399191/va-args-expansion-using-msvc

PR #365, opened 2022-10-03
Add dump mode to HPy tests.

fangerer:
We've discussed that we could create examples out of HPy's unit tests.
Since I'm usually looking at tests if I don't exactly remember how to do things in HPy, I found it is time to have that feature.

How does it work:
You just need to specify `--dump-dir=<dir>` like this:
```
hpyproject/hpy$ python3 -m pytest --dump-dir=test_sources test/
```
This will write all test sources passed to the compiler (in particular, to function `test.support.compile_module`) and then skip the test.
This mode will create one folder per test function (since a test function may have several sources).
I plan to automatically add the generated test sources also to the docs as examples but I'll do that in a separate step.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -2,6 +2,7 @@ import pytest
 from .support import ExtensionCompiler, DefaultExtensionTemplate,\
     PythonSubprocessRunner, HPyDebugCapture
 from hpy.debug.leakdetector import LeakDetector
+from pathlib import Path
 
 IS_VALGRIND_RUN = False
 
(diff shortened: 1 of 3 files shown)


antocuni (approved):
brilliant idea, I like it!
My only very small concern is that it doesn't have a test, so there is a non-zero risk that someone will break it by chance.
But I'll leave the decision to you, since it's already useful as-is.

hodgestar (approved):

fangerer:
> My only very small concern is that it doesn't have a test, so there is a non-zero risk that someone will break it by chance.

You are right, I'll add a test such that it is not accidentally broken.

Issue #359, opened 2022-09-23
Add two argument versions of common HPyDef_ macros


hodgestar:
It would be nice to have to repeat oneself less when defining slots and methods. E.g. instead of writing

```HPyDef_SLOT(Point_traverse, Point_traverse_impl, HPy_tp_traverse)```

it would be nice to write just:

```HPyDef_SLOT(Point_traverse, HPy_tp_traverse)``` and have the implementation be Point_traverse_impl by convention.

We can then keep the three argument versions for corner cases, or maybe even just get rid of them eventually? If one really needs a different name one can use a `#define`? This last paragraph is up for discussion.

antocuni:
sorry, I see it only now. I'd vote for killing they three argument versions: in the common case you use the short version, and if you need more control over the names you just write an `HPyDef` manually.

fangerer:
Afaik, this issue is resolved.

PR #367, opened 2022-10-10
Added shortcut for PyDef_* Macros

PurityLake:
Refers to #359 

Added 
* `HPyDef_SLOT2`
* `HPyDef_METH2`
* `HPyDef_GET2`
* `HPyDef_SET2`
* `HPyDef_GETSET2`

I couldn't think of a better name and perhaps renaming the original would create a breaking change. I made an assumption with the implied `IMPL` values passed to the original macros.


hodgestar:
Looks good! A few thoughts:

- There should be tests that use these macros (probably alongside the corresponding existing tests for these features)
- It would be good to have better names than `HPyDef_SLOT2`. One option is to name the new macros `HPyDef_SLOT` (i.e. change the default behaviour) and to name the old macros `HPyDef_SLOT_IMPL` (so that it's clear that it takes precisely an extra impl argument).

Changing the default behaviour might be annoying for existing ports, but if we like this better we should change it now while it's still possible. Hopefully others will chime in whether they like my suggested names.

PurityLake:
Ok huge new update, made each change a new commit so that its easier to see what I did

hpy/devel/include/hpy/hpydef.h
```diff
@@ -114,12 +114,12 @@ typedef struct {
    more detailed explanation in the comments around HPyFunc_DECLARE in
    hpyfunc.h
 */
-#define HPyDef_SLOT(SYM, IMPL, SLOT)                            \
+#define HPyDef_SLOT_IMPL(SYM, IMPL, SLOT)                            \
```
fangerer, line 117: The philosophy we followed so far was that we only provide a macro for the very common case.
I'm not sure if we should keep the `HPyDef_*_IMPL` macros since you can always manually fill your `HPyDef` struct.

hodgestar: My own view is that manually filling the HPyDef struct is quite a big jump in complexity for users. One has to suddenly understand all the details of the struct.

fangerer: I agree since I cannot fill an `HPyDef` from the top of my head myself.
The question is more: where should we make the cut. I remember that I once asked if we should have some variant for `HPyDef_METH` that covers the _ujson_ case (same C function; two Python methods) and we decided to not introduce a macro for that.
Anyway, I'm not really objected to keep the `HPyDev_*_IMPL` macros. I would also agree to add more. We just need to start writing docs about that.

PurityLake: `HPyDev_*_IMPL` can be good if you want to implement a slot with one of HPy's generic functions. I saw one in a test case

docs/examples/simple-example/simple.c
```diff
@@ -8,16 +8,27 @@
...
 {
     return HPy_Absolute(ctx, arg);
 }
 // END: myabs
 
+// BEGIN: thisabs
+#include "hpy.h"
+
+HPyDef_METH_IMPL(thisabs, "thisabs", thisabs_impl HPyFunc_O)
+static HPy myabs_impl(HPyContext *ctx, HPy self, HPy arg)
```
steve-s, line 22: shouldn't this be `thisabs_impl(HPyContext *ctx, ...)`?

steve-s: Maybe it should have a completely different name to better illustrate what the `_IMPL` variant of the macro is for?

PurityLake: Yes, apologies, will fix that in a minute

PurityLake:
The tests failed? Was full sure they passed for me, I'll look into this

PurityLake:
Ok, I believe the tests are passing now, sorry for the mess up

Issue #129, opened 2020-12-04
Argument clinic-like way to generate the argument parsing logic


antocuni:
This is an attempt to start a discussion on how to implement function calls and function definitions. In the following, "functions"  and "methods" are used interchangeably.

First, we need to identify who are the interested players and use cases:

- authors of C extensions who write C functions. Let's call them C-writers.

- Authors of C extensions who call Python functions. Let's call them C-callers.

- Generators of C extensions. Let's call them "cython" but it applies also to others.

- Python implementations with a JIT. Let's call them "pypy", but it applies also to others.

The basic idea is that we want to allow C-writers to write functions with a C-level signature, and generate the logic to do argument parsing automatically. E.g.:

```c
long my_add_impl(long x, long y) { return x+y; };
AUTOMAGICALLY_GENERATE_PYTHON_FUNCTION("my_add", my_add_impl);
```

The precise details of how to implement ``AUTOMAGICALLY_GENERATE_PYTHON_FUNCTION`` are the scope of this discussion.

Goals & constraints
====================

1. Writing functions in this way is optional. It will always be possible to write functions with the usual calling conventions such as   ``HPyFunc_VARARGS`` and to the argument parsing manually.

2.  There must be a way to quickly check whether a Python callable supports a given C-level signature and to get the underlying function pointer. "pypy" can use this to generate code which completely bypasses the python argument parsing logic, and cython" could use it to emit a fast-path in case it statically know the C types of the arguments.
  
3. Ideally, this should be integrated with the C API to *call* functions. E.g., if a C-caller calls ``HPy_Call(my_add, "ll", 4, 5)``, an implementation should be able to bypass argument parsing and call ``my_add_impl`` directly.
  
4. Bonus point if we find a way to implement goal 3 also on CPython.

5. It should be possible to do things manually: i.e., a C-writer could write its own argument parsing code **and** be able to declare a C-level signature.  In that case, he needs to ensure that its own argument parsing code does the "correct things".  In particular, when you do call from Python like ``my_add(4, 5)``, the implementation should be free to decide whether to call the generic version or the C-specialized overload.

6. To be discussed: we need to decide whether we want to support "overloads" or not. I.e., a given Python functions could in principle support many different C-level signatures. I think that all the following options are reasonable:
   
     * support at most one C-level signature. Functions which can't be encoded this way needs to be written in the "old style" and parse arguments manually.

[...]

PR #368, opened 2022-10-11
Use CPython FASTCALL to optimise hpy function call

ankith26:
Hello!
I'm new to the project and just want to experiment with using FASTCALL internally when HPy is running under cpython. In my understanding, this should give a decent performance boost while calling functions implemented in HPy from the python end, especially for short-running functions called many times the internal tuple creation overhead is not very negligible.

Currently I've only implemented `FASTCALL` for `VARARGS` and not `KEYWORDS` since the former is comparatively easier. The latter is not technically impossible, but would need larger changes (The main issue here is that HPy takes in the keywords as a dict, while python FASTCALL with KEYWORDS gives a tuple of keyword names, and the actual arguments still come via the first argument)

This PR should probably be postponed to when HPy drops CPython 3.6 support, because FASTCALL was an internal CPython implementation detail back then and not public API

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -288,10 +288,10 @@ which adds two integers:
 There are a few things to note:
 
   * The C signature is different than the corresponding ``Python.h``
-    ``METH_VARARGS``: in particular, instead of taking a ``PyObject *args``,
-    we take an array of ``HPy`` and its size.  This allows e.g. PyPy to do a
-    call more efficiently, because you don't need to create a tuple just to
-    pass the arguments.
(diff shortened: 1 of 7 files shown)


fangerer (approved):
Looks good to me.
For the mentioned future work (i.e. using `METH_FASTCALL | METH_KEYWORDS` for `HPyFunc_KEYWORDS`), I want to point to related issue #129

fangerer:
@ankith26 I'm sorry that this PR wasn't merged yet. As I wrote, it looks good to me. I was just waiting to give others a change for a review. Anyway, I took the freedom to rebase on current master and if all tests pass, I'll merge it.

Issue #366, opened 2022-10-05
Consider let build fail when hpy-abi=universal and `Py*` symbols are discovered


timfel:
This could happen when the stub loader is created, since the `.so` exists at that point. It would ensure that people don't accidentally build hpy universal binaries that only work because our implementations currently have cpyext (or are CPython)

mattip:
Another option is to set up compilation so that any use of `Python.h` will fail. This could be done by setting the include macro and creating a "fake" `Python.h`. Maybe both methods are needed to actually prove the binary is truly universal.

mattip:
I think this can be closed now that #371 is merged. Please reopen if I misunderstood.

PR #371, opened 2022-10-16
Introduce the Hybrid ABI

### LINKED ISSUE #288: Unable to compile on Python 3.11 alpha 5
When running `python3.11 -m pip wheel hpy --use-pep517` the output is as follows:

```console
Collecting hpy
  Downloading hpy-0.0.3.tar.gz (172 kB)
     |████████████████████████████████| 172 kB 9.5 MB/s 
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Installing backend dependencies ... done
    Preparing wheel metadata ... done
Building wheels for collected packages: hpy
  Building wheel for hpy (PEP 517) ... error
  ERROR: Command errored out with exit status 1:
   command: /home/domdf/.local/bin/python3.11 /home/domdf/python311/lib/python3.11/site-packages/pip/_vendor/pep517/in_process/_in_process.py build_wheel /tmp/tmp4xeykvcx
       cwd: /tmp/pip-wheel-9n1000e1/hpy_bf00928872e2496a9d116bd1dd260d92
  Complete output (37 lines):
  fatal: not a git repository (or any of the parent directories): .git
  running bdist_wheel
  running build
  running build_py
  running egg_info
  writing hpy.egg-info/PKG-INFO
  writing dependency_links to hpy.egg-info/dependency_links.txt
  writing entry points to hpy.egg-info/entry_points.txt
  writing requirements to hpy.egg-info/requires.txt
  writing top-level names to hpy.egg-info/top_level.txt
  listing git files failed - pretending there aren't any
  reading manifest file 'hpy.egg-info/SOURCES.txt'
  reading manifest template 'MANIFEST.in'
  adding license file 'LICENSE'
  running build_ext
  building 'hpy.universal' extension
  hpy/debug/src/debug_ctx_cpython.c: In function ‘debug_ctx_CallRealFunctionFromTrampoline’:
  hpy/debug/src/debug_ctx_cpython.c:162:20: error: dereferencing pointer to incomplete type ‘cpy_Py_buffer’ {aka ‘struct bufferinfo’}
    162 |             a->view->obj = NULL;
        |                    ^~
  hpy/debug/src/debug_ctx_cpython.c:165:36: warning: passing argument 3 of ‘_buffer_h2py’ from incompatible pointer type [-Wincompatible-pointer-types]
    165 |         _buffer_h2py(dctx, &hbuf, a->view);
        |                                   ~^~~~~~
        |                                    |
        |                                    cpy_Py_buffer * {aka struct bufferinfo *}
  hpy/debug/src/debug_ctx_cpython.c:41:78: note: expected ‘Py_buffer *’ {aka ‘struct <anonymous> *’} but argument is of type ‘cpy_Py_buffer *’ {aka ‘struct bufferinfo *’}
     41 | static void _buffer_h2py(HPyContext *dctx, const HPy_buffer *src, Py_buffer *dest)
        |                                                                   ~~~~~~~~~~~^~~~
  hpy/debug/src/debug_ctx_cpython.c:173:29: warning: passing argument 2 of ‘_buffer_py2h’ from incompatible pointer type [-Wincompatible-pointer-types]
    173 |         _buffer_py2h(dctx, a->view, &hbuf);
        |                            ~^~~~~~
        |                             |
        |                             cpy_Py_buffer * {aka struct bufferinfo *}
  hpy/debug/src/debug_ctx_cpython.c:56:61: note: expected ‘const Py_buffer *’ {aka ‘const struct <anonymous> *’} but argument is of type ‘cpy_Py_buffer *’ {aka ‘struct bufferinfo *’}
     56 | static void _buffer_py2h(HPyContext *dctx, const Py_buffer *src, HPy_buffer *dest)
        |                                            ~~~~~~~~~~~~~~~~~^~~
  error: command '/usr/bin/gcc' failed with exit code 1
  ----------------------------------------
  ERROR: Failed building wheel for hpy
Failed to build hpy
ERROR: Failed to build one or more wheels

```

The output is more or less the same when building from git, with some of the line numbers changing.

-----

It looks like `bufferinfo` was removed in https://github.com/python/cpython/pull/29991, which made it into alpha 5.

It's used here:

https://github.com/hpyproject/hpy/blob/dd702f55ce391df43cbf5dab1ce7d4b1be23110b/hpy/devel/include/hpy/cpy_types.h#L12

antocuni:
This PR introduces the `hybrid` ABI:
  - extensions which targets `universal` cannot `#include <Python.h>`, cannot use legacy features and produce a file called e.g. `simple.hpy0.so`.
  - extensions which targets `hybrid` can use `Python.h` and produce a file called e.g. `simply.hpy0-cpy38.so`.

See the updated docs for more info.

**Filename**
For universal, I choose to use `.hpy0.so` because we should use a versioned scheme. I didn't want to use `.hpy1.so` for now, because we might still want to change the ABI before becoming "official"

For hybrid, we want to include both the hpy version and the cpython version. I chose `hpy0-cp38.so`, which is the result of many sub-choices.
To start with, let's see what is the filename that you get on "normal" extensions:
- linux: `foo.cpython-38-x86_64-linux-gnu.so`.
- linux stable ABI: `foo.abi3.so`
- mac: `foo.cpython-310-darwin.so` (or `.dylib`? I think I've seen both)
- win: `foo.cp38-win_amd64.pyd`
- pypy: `foo.pypy38-pp73-x86_64-linux-gnu.so`
- graalpy: `foo.graalpy-38-native-x86_64-darwin.dylib`

It varies wildly: 
  - cpython uses both `cpython-38` and `cp38`. 
  - the platform tag sometimes includes the CPU, sometimes not
  - the stable ABI extensions doesn't include the platform tag at all

So for hpy hybrid I chose to:
  - use `cp38` instead of `cpython-38`: this is shorter, and most importantly it's the same tag which goes in the wheel names
  - don't include the platform tag: if it's not needed for the stable ABI, it means it's not needed for us as well

**Compile time errors**
In universal mode, you are not allowed to use `Python.h`: we try hard to display a nice compile time error by putting our own `Python.h` in the include dirs. This might not work in all setups, but it should be good enough for most cases.

In universal mode, you cannot call `HPy_AsPyObject` and `HPy_FromPyObject`: I managed to get a compile-time error on gcc, but not on old clangs nor in MSVC. Again, too bad but it's better than nothing:
https://github.com/hpyproject/hpy/blob/4cde71792d754cc0445b39a6795e68536e877a04/hpy/devel/include/hpy/cpy_types.h#L24-L27
https://github.com/hpyproject/hpy/blob/4cde71792d754cc0445b39a6795e68536e877a04/hpy/devel/include/hpy.h#L76-L95

The other legacy features that you cannot use are `.legacy_methods` and `.legacy_slots`: but these are now impossible to use, because in order to pass anything meaningful you have to instantiate e.g. a `PyMethodDef` struct, which you can't because you cannot include `Python.h`

We still need to reference some `cpy_*` types, e.g. to declare the `HPyModule_Spec.legacy_methods` field. This is done here:
https://github.com/hpyproject/hpy/blob/4cde71792d754cc0445b39a6795e68536e877a04/hpy/devel/include/hpy/cpy_types.h#L4-L36

in cpython and hybrid modes, we use the real types from `Python.h`. In universal mode, we use "fake" types, but note that they are only forward declarations and they cannot be instantiated.

This should also fix #288.

**Future improvements**
There are many things which we can to to improve the situation, but I think they can be done in future PRs:
  - post-compilation sanity check: after compiling to `universal`, we could run a sanity check to actually ensure that we don't link against any `Py*` symbol
  - runtime check: it would be nice to have two different context for universal and hybrid ABI: the universal ctx would fail cleanly if you try to call `HPy_{As,From}PyObject` or use `.legacy_methods`, etc.
  - better compile time errors on different compilers


hpy/devel/include/hpy.h
```diff
@@ -37,6 +72,28 @@ extern "C" {
 #  define _HPy_NO_RETURN
 #endif
 
+
+// clang and gcc supports __has_attribute, MSVC doesn't. This should be enough
```
fangerer, line 76: What if we surround the auto-generated trampolines with some `#ifdef HPY_ABI_UNIVERSAL` ?

hpy/devel/include/hpy/cpy_types.h
```diff
@@ -1,15 +1,42 @@
...
+// #error if they are used
+HPyAPI_FUNC _HPY_LEGACY cpy_PyObject *HPy_AsPyObject(HPyContext *ctx, HPy h);
+HPyAPI_FUNC _HPY_LEGACY HPy HPy_FromPyObject(HPyContext *ctx, cpy_PyObject *obj);
+
+#else
+
+// Python.h has already been included by the main hpy.h
+typedef PyObject cpy_PyObject;
+typedef PyMethodDef cpy_PyMethodDef;
+typedef Py_buffer cpy_Py_buffer;
```
fangerer, line 34: Ah, that's great and will fix our `CPython >=3.11` compatibility problem.
Maybe you could mention issue #288 in this PR.

hpy/devel/include/hpy/hpymodule.h
```diff
@@ -31,30 +31,30 @@ typedef struct {
...
-        _ctx_for_trampolines = ctx;                            \
-        return init_##modname##_impl(ctx);                     \
+        return _h2py(init_##modname##_impl(_HPyGetContext())); \
     }
 
-#else // HPY_UNIVERSAL_ABI
+#else // HPY_ABI_CPYTHON
 
-// module initialization in the CPython case
+// module initialization in the universal case
```
fangerer, line 47:
```suggestion
// module initialization in the universal and hybrid case
```

hpy/universal/src/hpymodule.c
```diff
@@ -155,7 +155,7 @@ static PyObject *get_version(PyObject *self, PyObject *ignored)
 }
 
 static PyMethodDef HPyMethods[] = {
-    {"load", (PyCFunction)load, METH_VARARGS | METH_KEYWORDS, "Load a .hpy.so"},
+    {"load", (PyCFunction)load, METH_VARARGS | METH_KEYWORDS, "Load a .hpy0.so"},
```
fangerer, line 158: We are also using this in `hpy/devel/abitag.py`. I think these two usages should have a strong connection. Can we do what we do with the version? Define it in a C header and expose it in Python.

test/conftest.py
```diff
@@ -19,6 +19,10 @@ def pytest_addoption(parser):
         help="Enables dump mode and specifies where to write generated test "
              "sources. This will then only generate the sources and skip "
              "evaluation of the tests.")
+    parser.addoption(
+        '--reuse-venv', action="store_true",
```
fangerer, line 23: 👍

test/test_legacy_forbidden.py
```diff
@@ -0,0 +1,76 @@
...
+from .support import HPyTest, make_hpy_abi_fixture
+
+# this is not strictly correct, we should check whether the actualy compiler
+# is GCC. But for the CI and most cases, it's enough to assume that if we are
+# on linux we are using GCC.
+#
+# We need this because some of the nice compilation errors (such as the ones
+# causes by _HPY_LEGACY) are triggered only by gcc. Would be nice to have them
+# also for other compilers
+ONLY_GCC = pytest.mark.skipif(sys.platform!='linux', reason='GCC only')
```
fangerer, line 17: Should that maybe go into `support.py` ?

fangerer (approved):
Great work, thank you!
I added some comments but I don't see blockers.

hpy/devel/abitag.py
```diff
@@ -0,0 +1,63 @@
...
+    # platform:
+    #   - linux:   '.cpython-310-x86_64-linux-gnu.so'
+    #   - mac:     '.cpython-310-darwin.so'
+    #   - win:     '.cp310-win_amd64.pyd'
+    #   - pypy:    '.pypy38-pp73-x86_64-linux-gnu.so'
+    #   - graalpy: '.graalpy-38-native-x86_64-darwin.dylib'
+    assert ext_suffix[0] == '.'
+    _, soabi, ext = ext_suffix.split('.')
+
+    # this is very ad-hoc but couldn't find a better way to do it
```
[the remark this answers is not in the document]
mattip: See also [this PR](https://github.com/pypa/packaging/pull/607/files#diff-e2a5f8198f944bfcd050534d68367fd50e3f9f28afe1bb4b54c72afccd8fc0f6R228) to packaging/tags to improve the generic abi parsing there.

fangerer:
@steve-s You are right, I forgot about that effort. Would it make sense to introduce some generic ABI tag?

fangerer:
> What exactly do you mean by that? That we'd drop `hpyX` tag as suggested in this PR and use just generic `hpy`?

Yes, that's what I meant.

I do also think that a generic ABI tag does not make a lot of sense. I agree that we can have both.

hpy/devel/abitag.py
```diff
@@ -0,0 +1,63 @@
...
+        n = 2
+    elif parts[0].startswith('graalpy'):    # graalpy
+        n = 3
+    else:                                   # none of the above, keep all parts
+        n = len(parts)
+
+    abi_tag = '-'.join(parts[:n])
+    # on cpython linux/mac, abi is now cpython-310: shorten it ot cp310, like
+    # on windows and on wheel tags
+    abi_tag = abi_tag.replace('cpython-', 'cp')
```
mattip, line 43: Is it possible to change the value graalpython returns here or has that ship sailed? I think graalpy-38-native is a bit long. This is the name used in the wheel filename, so will wheels be called `numpy-1.23.4-graalpy38-graalpy-38-native-manylinux_2_17_x86_64.whl ` ?

mattip:
Is there more to do here or should we merge this as-is and incrementally improve things toward a release?

mattip:
Closes #366

steve-s:
I also had a cursory look over the changes and I vote for merging it and incrementally improve things toward a release

hpy/universal/src/hpymodule.c
fangerer: One little more thing: there is an `int ctx_version` field in `HPyContext`. Matti brought to my attention, that this field basically is the ABI version, right? So, I think we should set `ctx_version` appropriately.

hpy/devel/abitag.py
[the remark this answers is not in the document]
antocuni: > We can just have our own scheme

This is possible for universal modules (and it's essentially what we do) but not for hybrid: for hybrid, we also want to include enough info to be able to distinguish the host interpreter, and so the only reliable source of information is the host interpreter itself.
For example, on CPython we need to distinguish between e.g. `cp38` and `cp38d`.
Basically, my idea is to start from EXT_SUFFIX and carefully remove unnecessary stuff (such as the platform), while keeping all the rest (which I treat as mostly opaque).

> See also [this PR](https://github.com/pypa/packaging/pull/607/files#diff-e2a5f8198f944bfcd050534d68367fd50e3f9f28afe1bb4b54c72afccd8fc0f6R228) to packaging/tags

yes, it looks very similar. I wonder whether it's a good idea to just reuse it?

hpy/devel/include/hpy/cpy_types.h
antocuni: right! I edited the description to mention it

hpy/devel/include/hpy.h
fangerer: Not to this particular line but to the block from line 78 to 82.
What I meant is: `_HPY_LEGACY` is used to mark legacy functions such that the extension author gets compilation error if they are called and compiled in universal ABI, right?
My suggestion was that instead of this attribute, we could just define those functions in a way that the produce a compilation error. But I'm fine with this approach.

mattip:
after the capsule changes, this PR has conflicts

hpy/devel/include/hpy.h
antocuni: ah ok I understand what I mean. I think it would be worse, because in that case the C compiler would assume that e.g. `HPy_AsPyobject` has an implicit declaration: it would emit a warning and happily compile, which is even worse.

hpy/universal/src/hpymodule.c
antocuni: yes, good point. It turns out that sharing the same definition in C and Python is a mess as usual, so I opted for defining it in two places (hpy.h and abitag.py) and add a test which checks that they are on sync.
Also, I renamed `ctx->ctx_version` into `ctx->abi_version` and set its value to `HPY_ABI_TAG`, see commits 71d2a7d and 1f70e65.

test/test_legacy_forbidden.py
antocuni: fixed by 4907ee8

antocuni:
> If we include the ABI version in the tag, I am wondering if we need #356. At the same time I like the approach in #356, because it will be "fool-proof" -- even if you have wrong tag for whatever reason it'll still check and choose the right ABI for your extension. So right now I am in favor of keeping both the ABI tag and #356 even though they are redundant.

if think we can have both: basically, the logic of this PR tries to ensure that you cannot use CPython symbols by mistake, and #356 would become a sanity check to ensure that it works well.
Not sure if it's high-priority, but it surely doesn't hurt.

antocuni:
@mattip @fangerer @steve-s I think I addressed all the comments and fixed the conflicts.
If you don't have anything against, I'll merge the PR as soon as the tests are green

fangerer:
LGTM. Please go ahead and merge.
That's great work. Thank you!

mattip:
Yay!

PR #376, opened 2022-11-08
Add built-in types/objects complex, bytes, memoryview, capsule, slice and builtins to HPyContext.

fangerer:
This adds a few more built-in types/objects to the context.
The `h_Builtins` is of particular interest because it allows to do many operations that are available in Python but HPy does not yet have dedicated API for.

Further, we now also use Autogen to generate the (smaller) `HPyContext` struct for the CPython ABI.


.github/workflows/ci.yml
```diff
@@ -363,7 +363,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v2
```
mattip, line 364:
```suggestion
        uses: actions/setup-python@v4
```

mattip (approved):
LGTM. Is there somewhere that a version is bumped?

fangerer:
> LGTM. Is there somewhere that a version is bumped?

Thanks for the review. 
Sorry, I don't get your question. Do you mean if GitHub or Ubuntu documented that it upgraded to Python 3.11 ?
I quickly tried to find out more. According to https://packages.ubuntu.com/jammy/python/python3 or https://packages.ubuntu.com/jammy-updates/python/ the used Python version is still `3.10.7`.
GitHub Action `setup-python` says that it will use the version from `PATH` if no version was explicitly specified (see https://github.com/actions/setup-python). But I think it is because we used `python-version: '3.x'` and according to https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md, this will use the latest stable version. And since 3.11 is the latest stable one, this would make sense to me.

mattip:
I was referring to updating the ABI tag since this changes the ctx struct, but that seems more appropiate as we approach a release.

PR #378, opened 2022-11-22
Improve API/implementation for HPyCapsule_Destructor

### LINKED ISSUE #377: Improve the API/implementation for HPyCapsule_Destructor
this is best explained by this comment:
https://github.com/hpyproject/hpy/pull/308#issuecomment-1322131574

fangerer:
This resolved #377. 
In this PR, I'm introducing macro `HPyDef_CAPSULE_DESTRUCTOR` (much like `HPyDev_METH` macro) and removing `HPyCapsule_GetDestructor`. With this, we can significantly simplify our capsule implementation because we never need to hand out the HPy capsule destructor function while the new macro will generate an appropriate CPython trampoline.

Removing `HPyCapsule_GetDestructor` should be fine since I didn't find a single usage of `PyCapsule_GetDestructor` in the C code of the top4000 packages.

### DIFF
--- a/test/test_capsule.py
+++ b/test/test_capsule.py
@@ -75,7 +75,7 @@ class CapsuleTemplate(DefaultExtensionTemplate):
                         HPyErr_SetString(ctx, ctx->h_MemoryError, "out of memory");
                         return HPy_NULL;
                     }
-                    return HPyCapsule_New(ctx, ptr, CAPSULE_NAME, (HPyCapsule_Destructor) %s);
+                    return HPyCapsule_New(ctx, ptr, CAPSULE_NAME, (HPyCapsule_Destructor *) %s);
                 }
                 /* just for error case testing */
                 return HPyCapsule_New(ctx, NULL, CAPSULE_NAME, NULL);
@@ -167,7 +167,6 @@ class TestHPyCapsule(HPyTest):
         try:
             assert mod.capsule_getname(p) == "some_capsule"
         finally:
-            # since HPy's capsule API does not allow a destructor, we need to
             # manually free the payload to avoid a memleak
             mod.payload_free(p)
         with pytest.raises(ValueError):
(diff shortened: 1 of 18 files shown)


mattip:
LGTM, but will defer to @antocuni

test/test_capsule.py
```diff
@@ -75,7 +76,11 @@ def DEFINE_Capsule_New(self, destructor="NULL"):
                         HPyErr_SetString(ctx, ctx->h_MemoryError, "out of memory");
                         return HPy_NULL;
                     }
-                    return HPyCapsule_New(ctx, ptr, CAPSULE_NAME, (HPyCapsule_Destructor) %s);
+                    res = HPyCapsule_New(ctx, ptr, CAPSULE_NAME, (HPyCapsule_Destructor *) %s);
```
antocuni, line 79: is the cast to `(HPyCapsule_Destructor *)` actually needed?

antocuni (changes requested):
LGTM, apart the name of the macro: the idea of the various existing `HPyDef_*` macros is that each of them defines a static variable whose C type is actually `HPyDef`.

`HPyDef_CAPSULE_DESTRUCTOR` does something different, so it should be named differently; it belongs to the `HPyCapsule` "namespace", so what about `HPyCapsule_DESTRUCTOR`?

test/test_capsule.py
fangerer: I think it was required since `g++` issued a warning but I'll verify.

fangerer:
> LGTM, apart the name of the macro: the idea of the various existing `HPyDef_*` macros is that each of them defines a static variable whose C type is actually `HPyDef`.
> 
> `HPyDef_CAPSULE_DESTRUCTOR` does something different, so it should be named differently; it belongs to the `HPyCapsule` "namespace", so what about `HPyCapsule_DESTRUCTOR`?

Okay, I'm not objected to rename it. I did that for consistency but didn't have in mind that the `HPyDef_*` prefix is because of the `HPyDef` type.

test/test_capsule.py
fangerer: No, the cast is not necessary (I already removed it). I guess it was previously necessary when `HPyCapsule_Destructor` was a function pointer.

antocuni (approved):
LGTM, thanks

Issue #310, opened 2022-04-07
Do not compile helpers with extension by default.


fangerer:
Right now, we are adding source files `argparse.c`, `buildvalue.c`, and `helpers.c` to the source file list of the HPy extension to be built and then they are compiled together with the extension's sources.

We do so because the goal is that the helpers binary code should be in the HPy extension's library.
While this works in most cases, we recently discovered a problem with that strategy on Mac (using LLVM toolchain). The extension was written in C++ and added:
```
extra_compile_args=['-std=c++11', '-stdlib=libc++']
extra_link_args=['-stdlib=libc++']
```
The extension failed to build `argparse.c` with following error:
```
gcc -Wno-unused-result -Wsign-compare -g -O0 -Wall -DHPY -DHPY_UNIVERSAL_ABI -I. -I/path/to/lib/python3.8/site-packages/hpy/devel/include -I/path/to/include -I/Users/fa/work/repos/Python-3.8.5/Include -I/Users/fa/work/repos/Python-3.8.5 -c /path/to/lib/python3.8/site-packages/hpy/devel/src/runtime/argparse.c -o build/temp.macosx-12.0-x86_64-3.8-pydebug/path/to/lib/python3.8/site-packages/hpy/devel/src/runtime/argparse.o -std=c++11 -stdlib=libc++
error: invalid argument '-std=c++11' not allowed with 'C'
error: command 'gcc' failed with exit status 1
```

IMO, the underlying issue is that we are building these sources with the flags of the target. I think we should build them in the same way as we do with the rest of HPy. So, I suggest that we build an archive for the helpers and statically link them into the extension.

Optionally, if someone really wants to compile the helpers with the extension, we can provide an option to do so.

hodgestar:
Would adding
```c
    #ifdef __cplusplus
    extern "C" {
    #endif
```
blocks to the relevant `.c` files help at all?

fangerer:
Unfortunately, this does not help (just tried it out).

hodgestar:
> Unfortunately, this does not help (just tried it out).

Pity! Thanks for trying it out.

steve-s:
We should perhaps remove the `extern "C"` blocks in `ctx_*.c` files. I searched around and afaics it just disables names mangling, but the C++ semantics stays the same. I'd say that compiling those files with C++ compiler is just not supported. We should try to keep the header files, especially the inline functions, in some common subset of C and C++ if possible though.

PR #379, opened 2022-11-23
Build static libraries of HPy devel sources for testing.

fangerer:
This PR resolves issue #310.

As described in the linked issue, we are currently adding our sources (i.e. `argparse.c`, `buildvalue.c`, `helpers.c` for universal ABI and additionally all `hpy/devel/src/runtime/ctx_*.c` for CPython ABI) to every HPy extension.
This may cause problems for the extension author since he maybe wants to use compilation flags or even compilers that are not compatible with our source.

Further, our tests are also slowed down extremely because we compile those sources over and over for every single test case.

### Approach

With this PR, we will build static libraries in HPy's `setup.py` (for _universal_ and _cpython_ ABI). The main problem to do so was that setuptools doesn't really provide a facility to do that and I didn't want to use a lot of distutils internals since distutils is deprecated and subject to be removed.

I decided to build the static libraries using the `build_clib` command. This is actually what it is good for *BUT* with the assumption that such libraries are just temporary build aritfacts and that they will immediately after be used in `build_ext`.
I therefore customized `build_clib`and such that static libraries are emitted to the same output directory as `build_ext` uses but into sub-directory `lib/<abi>`.
I've defined two static libraries: `hpyextra` for _universal ABI_ and `hpyctx` for _ cpython ABI_ (see variables `HPY_EXTRA_SOURCES` and `HPY_CTX_SOURCES`, respectively).

I tried hard to minimize dependencies on deprecated distutils features. However, the introduced `build_clib_hpy` command overrides methods `get_library_names` and `build_libraries` which is not officially supported, I think (see https://setuptools.pypa.io/en/latest/userguide/extension.html#setuptools.command.build.SubCommand).

Unfortunately, we need some extra treatment to be able to use that for the HPy tests. Although we install HPy before we run the tests, we run them from the HPy source directory which means that we don't use the `hpy.devel` from the installed distribution but from the source tree.
I therefore also implemented an _inplace_ mode for `build_clib_hpy` (the option is _inherited_ from `build_ext`). That is used to output build artifacts into the source tree such that we can use them in the tests.

### Numbers

On my Linux box, this improves test duration a lot:
```
before: 8m 28s
after: 1m 23s
speedup: 6.12x
```
The numbers on my Windows VM were not that good (something like 30% faster) but let's see how this goes in the GitHub CI.

### DIFF
--- a/hpy/devel/__init__.py
+++ b/hpy/devel/__init__.py
@@ -329,7 +329,7 @@ class build_ext_hpy_mixin:
         if static_libs:
             static_libs = self.hpydevel.get_static_libs(ext.hpy_abi)
             if static_libs is None or len(static_libs) != 1:
-                raise DistutilsError('Expected exactly one static library for'
+                raise DistutilsError('Expected exactly one static library for '
                                      'ABI "%s" but got: %r' %
                                      (ext.hpy_abi, static_libs))
 
--- a/hpy/devel/__init__.py
+++ b/hpy/devel/__init__.py
@@ -178,7 +178,7 @@ def handle_hpy_ext_modules(dist, attr, hpy_ext_modules):
 
     # add a global option --hpy-abi to setup.py
     dist.__class__.hpy_abi = DEFAULT_HPY_ABI
-    dist.__class__.hpy_no_static_libs = False
+    dist.__class__.hpy_use_static_libs = False
     dist.__class__.global_options += [
         ('hpy-abi=', None, 'Specify the HPy ABI mode (default: %s)' % DEFAULT_HPY_ABI),
         ('hpy-no-static-libs', None, 'Compile context and extra sources with extension (default: False)')
@@ -322,11 +322,11 @@ class build_ext_hpy_mixin:
         ext.name = HPyExtensionName(ext.name)
         ext.hpy_abi = self.distribution.hpy_abi
         ext.include_dirs += self.hpydevel.get_extra_include_dirs()
-        # If we may use static libs (default), then add all available libs (for
+        # If static libs should be used, then add all available libs (for
         # the given ABI) to the extra objects. The libs will then just be added
         # in the linking phase but nothing will be compiled in addition.
-        use_static_libs = not self.distribution.hpy_no_static_libs
-        if use_static_libs:
+        static_libs = self.distribution.hpy_use_static_libs
+        if static_libs:
             static_libs = self.hpydevel.get_static_libs(ext.hpy_abi)
         else:
             static_libs = None
--- a/hpy/devel/__init__.py
+++ b/hpy/devel/__init__.py
@@ -328,8 +328,10 @@ class build_ext_hpy_mixin:
         static_libs = self.distribution.hpy_use_static_libs
         if static_libs:
             static_libs = self.hpydevel.get_static_libs(ext.hpy_abi)
-        else:
-            static_libs = None
+            if static_libs is None or len(static_libs) != 1:
+                raise DistutilsError('Expected exactly one static library for'
+                                     'ABI "%s" but got: %r' %
+                                     (ext.hpy_abi, static_libs))
 
         if static_libs:
             ext.extra_objects += static_libs
--- a/hpy/devel/__init__.py
+++ b/hpy/devel/__init__.py
@@ -52,6 +52,7 @@ class HPyDevel:
         self.base_dir = Path(base_dir)
         self.include_dir = self.base_dir.joinpath('include')
         self.src_dir = self.base_dir.joinpath('src', 'runtime')
+        self._available_static_libs = None
 
     def get_extra_include_dirs(self):
         """ Extra include directories needed by extensions in both CPython and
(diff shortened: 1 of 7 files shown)


mattip:
> This may cause problems for the extension author since he maybe wants to use compilation flags or even compilers that are not compatible with our source.

Compilers that are not compatible will have a hard time linking to the static libraries. Maybe you were thinking of dynamic libraries that export the needed functions? 

I think the testing improvements are significant and are enough reason for this PR. We might also want to explore using precompiled headers, that speeds up PyPy translation quite a bit.

fangerer:
> Compilers that are not compatible will have a hard time linking to the static libraries. 

You are basically right but I forgot to mention that HPy extension authors may disable this _feature_ by passing `--hpy-no-static-libs` to `setup.py`. I think I need to add some docs.

> Maybe you were thinking of dynamic libraries that export the needed functions?

No, definitively not. Previously, we added the sources to the HPy extension which means that the code would statically be compiled into the extension's binary. This PR just tries to avoid the repeated compilation (w/o linking) of the helpers and context sources. Anyway, I was thinking that a dynamic lib could make sense for context impl (in CPython ABI).

> I think the testing improvements are significant and are enough reason for this PR.

If we are unsure about this, we could disable this feature by default and just enable it for testing.

mattip:
> This PR just tries to avoid the repeated compilation (w/o linking) of the helpers and context sources

Other than tests, how common is the repetition? I think default disabled and enabling for tests makes sense

hpy/devel/__init__.py
```diff
@@ -293,10 +327,28 @@ def _finalize_hpy_ext(self, ext):
...
+        else:
+            static_libs = None
+
+        if static_libs:
+            ext.extra_objects += static_libs
+        else:
+            # If we should not use (pre-compiled) static libs or if they are
+            # not available, we just add the sources of the helpers to the
+            # extension. They are then compiler with the extension.
+            ext.sources += self.hpydevel.get_extra_sources()
```
antocuni, line 345: so, if I understand correctly the logic is the following:
1. by default we use the `.a` file
2. if you specify `--hpy-no-static-libs`, we use `*.c` files
3. if you try to use the `.a` file but you don't find it, you fallback to (2)

I think that (1) and (2) are reasonable, but what is the use case for (3)?
In which case do we expect it to be useful?

If we don't have a concrete use case and the logic is there "just in case", I vote for removing it and emit a clean error instead (something like "Cannot find libhpy.a"): basically I'm worried about the following scenario:
1. at some point the future we make a mistake when building `libhpy*.a.` (e.g. by putting it in the wrong directory or something like that) 
2. we don't notice because things still work as usual (but using *.c instead of the static lib)
3. at some point someone will try to compile a C++ lib again and they will be very confused

setup.py
```diff
@@ -93,48 +94,136 @@ def get_scm_config():
...
+]
+
+HPY_INCLUDE_DIRS = [
+    'hpy/devel/include',
+    'hpy/universal/src',
+    'hpy/debug/src/include',
+]
+
+HPY_EXTRA_LIB_NAME = "hpyextra"
+HPY_CTX_LIB_NAME = "hpyctx"
```
antocuni, line 124: if I understand correctly, this is a cpython-specifc library, isn't it? Then, I propose to call it `hpy-extra-cpython` or something like that, which makes it crystal clear what is the intended usage

setup.py
```diff
@@ -93,48 +94,136 @@ def get_scm_config():
...
+            finally:
+                self.build_temp = orig_build_temp
+                self.build_clib = orig_build_clib
+
+
+STATIC_LIBS = [(HPY_EXTRA_LIB_NAME,
+                {'sources': HPY_EXTRA_SOURCES,
+                 'include_dirs': get_hpy_runtime_includes(),
+                 'abi': 'universal',
+                 'macros': [('HPY_UNIVERSAL_ABI', None)]}),
```
antocuni, line 197: if I understand correctly `'abi'` is not a distutils thing but it's some custom setting which is used only by us, right?
Then I propose to call it `hpy_abi` (which is also consistent to how it is called everywhere else in the code)

test/support.py
```diff
@@ -467,7 +467,11 @@ def _build(tmpdir, ext, hpy_devel, hpy_abi, compiler_verbose=0, debug=None):
 
     # this is the equivalent of passing --hpy-abi from setup.py's command line
     dist.hpy_abi = hpy_abi
+    dist.hpy_no_static_libs = False
```
antocuni, line 470: why do you need to set it explicitly here? Isn't it the default?

antocuni (commented):
LGTM, good job!
I think we are missing a test for `--hpy-no-static-libs` though. Probably it should be put in `test_distutils.py`

hpy/devel/__init__.py
antocuni: To elaborate more: basically, instead of "let's search from random *.a files in the given lib", I'd change the logic to "let's find hpy/devel/lib/libhpy-extra.a" (or libhpy-extra-cpython.a, accordingly).

fangerer: Sorry that I never answered your comments. I was side-tracked and just forgot about it.

> so, if I understand correctly the logic is the following:
> - by default we use the .a file
> - if you specify --hpy-no-static-libs, we use *.c files
> - if you try to use the .a file but you don't find it, you fallback to (2)

Yes, right now, it tries to find `*.a` (or `*.lib` on Windows) files for the given HPy ABI.
That can be disabled using `--hpy-no-static-libs`.
If the libs are not available but we want use static libs (your point (3) ), we fallback and include/compile sources with the extension as before.
Reason: I tried to be defensive and didn't want to make things worse as before. It is IMO even necessary because we only ship libraries for our well-known ABIs. If someone has another ABI than `universal` or `cpython`, then we cannot ship the static libs.

But I also agree, it might be confusing. Anyway, I think I'll go a step back with that PR and will just use this for the tests.
I realized that there is a severe conceptual problem: shipping the static libs means that they are built in the CI. But the whole point of the extra sources was that they are compiled on the *client* side for portability (because of varargs etc).

> To elaborate more: basically, instead of "let's search from random *.a files in the given lib", I'd change the logic to a more explicit "let's find hpy/devel/lib/libhpy-extra.a" (or libhpy-extra-cpython.a, accordingly).

I tried to avoid looking for a specific file like `libhpy-extra.a` because setuptools' support for that is poor. We would need to make assumptions about the naming scheme and I tried to avoid that by just looking into a well-known folder (which is ABI-specific) and take whatever is there.

setup.py
fangerer: Right. `hpyextra` is right now just for `universal` and `hpyctx` (which contains the extra sources as well) is just used for `cpython`.

setup.py
fangerer: Yes, `abi` is not a distutils thing. Okay, I'll rename.

test/support.py
fangerer: We need to set the attribute here otherwise it just doesn't exist on the `Distribution` object. The tests are not going through `handle_hpy_ext_modules`. This is all done manually here (e.g. we call `fix_distribution`in line 475).

mattip (approved):
Let's give this a try

PR #385, opened 2022-11-29
Enable slots HPy_tp_str and HPy_tp_hash.

### LINKED ISSUE #384: Enable slots HPy_tp_str and HPy_tp_hash.
Slots `tp_str` and `tp_hash` are used in Numpy and thus required to be able to port Numpy to HPy.

fangerer:
Resolves #384 .

Slots `HPy_tp_str` and `HPy_tp_hash` are required for the NumPy/HPy port.

### DIFF
--- a/test/test_slots.py
+++ b/test/test_slots.py
@@ -423,6 +423,44 @@ class TestSlots(HPyTest):
             assert sys.getrefcount(arr) == init_refcount
         mv2 = memoryview(arr)  # doesn't raise
 
+    def test_tp_repr_and_tp_str(self):
+        mod = self.make_module("""
+            #include <stdio.h>
+
+            #define BUF_SIZE 128
+
+            @DEFINE_PointObject
+            @DEFINE_Point_new
+
+            static HPy
+            point_str_repr(HPyContext *ctx, HPy h, int str)
+            {
+                char buf[BUF_SIZE];
+                PointObject *p = PointObject_AsStruct(ctx, h);
+                snprintf(buf, BUF_SIZE, "%s(Point(%ld, %ld))",
+                            (str ? "str" : "repr"), p->x, p->y);
+                return HPyUnicode_FromString(ctx, buf);
+            }
+
+            HPyDef_SLOT(Point_repr, HPy_tp_repr)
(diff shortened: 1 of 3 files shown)


mattip (approved):
These already worked as `__hash__` and `__repr__` but not as `tp_*` slots? Yeah, LGTM.

test/test_slots.py
```diff
@@ -423,6 +423,72 @@ def test_buffer(self):
...
+
+            @EXPORT_POINT_TYPE(&Point_new, &Point_hash)
+            @INIT
+        """)
+        p = mod.Point(1, 10)
+        assert p.__hash__() == 11
+        assert hash(p) == 11
+        # We expect that the slot wrapper accepts hash code -1 without
+        # complaining. This is not the case for built-in function 'hash'.
+        assert mod.Point(0, -1).__hash__() == -1
```
hodgestar, line 488: This seems like a straight up bug in the CPython implementation of  `__hash__`. The C code set and exception and returned the documented `-1` error value, and the exception was simply dropped.

Should we report it? Or is this a known CPython weirdness?

hodgestar: P.S. I'm happy for this to be merged as is -- I was just very surprised.

fangerer: No, I don't think this is a bug. Sorry, the test may be a bit confusing. `Point_hash_impl` only raises an exception if `x < 0`. This is not the case in line 488. No exception is suppressed.
The test aims to ensure that other runtimes actually do the same check, i.e., `if (result == -1 && PyErr_Occurred()) /* error */`. See https://github.com/python/cpython/blob/74d5f61ebd1cb14907bf7dae1ad9c1e676707bc5/Objects/typeobject.c#L7500.
However, if we would use `hash(mod.Point(0, -1))` we would get a `SystemError` because of https://github.com/python/cpython/blob/5bbf8ed8fbca0d68144c7b317ea8550420ac2594/Python/bltinmodule.c#L1567.

So, it is rather a *known* CPython weirdness.

hodgestar: Oh, I see. Only `y` is set to `-1`. So the hash returns `-1` with no error set. Please ignore my comment. :)

mattip:
Thanks @fangerer

PR #388, opened 2022-12-06
Fix GitHub actions: fix Ubuntu version for py27 compat and cpp-check

steve-s:
The latest Ubuntu does not provide Python 2.7 and also cpp-check seems to have some issues with it (runs, but fails with some preprocessor configuration issues).

From https://github.com/actions/setup-python/issues/543 it seems that we need to stick to older Ubuntu.

Note: EOL of Ubuntu 20.04 should be in 2025

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -316,7 +316,7 @@ jobs:
 
   check_py27_compat:
     name: Check Python 2.7 compatibility
-    runs-on: 'ubuntu-latest'
+    runs-on: 'ubuntu-20.04'
     steps:
       - uses: actions/checkout@v2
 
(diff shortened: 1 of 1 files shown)


fangerer (approved):

mattip:
On the one hand we could add support for older python versions with [deadsnake](https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa). On the other, at what point do we drop 2.7 language compatibility?

steve-s:
> what point do we drop 2.7 language compatibility?

I am not sure I follow, we did not drop it, right? (For the tests, my understanding was that PyPy test driver needs that, otherwise HPy itself was never 2.7 compatible?)

PR #391, opened 2022-12-20
Split valgrind tests.

fangerer:
This introduces a [reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows) for the valgrind tests.
I'm using `pytest-portion` to split the tests. This is controlled by env variable `HPY_TEST_PORTION` which is honored by the Makefile.

### DIFF
--- a/.github/workflows/valgrind-tests.yml
+++ b/.github/workflows/valgrind-tests.yml
@@ -1,0 +1,29 @@
+on:
+  workflow_call:
+    inputs:
+      portion:
+        required: true
+        type: string
+
+jobs:
+  valgrind_tests:
+    # name: Valgrind tests
+    runs-on: 'ubuntu-latest'
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install / Upgrade system dependencies
+        run: sudo apt update && sudo apt install -y valgrind
+
+      - name: Install / Upgrade Python dependencies
+        run: python -m pip install --upgrade pip wheel
+
+      - name: Build
+        run: python -m pip install .
+
+      - name: Run tests
+        env:
+          HPY_TEST_PORTION: ${{ inputs.portion }}
+        run: |
+          python -m pip install pytest pytest-valgrind pytest-portion filelock
+          make valgrind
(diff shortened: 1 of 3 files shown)


.github/workflows/valgrind-tests.yml
```diff
@@ -0,0 +1,29 @@
...
+      portion:
+        required: true
+        type: string
+
+jobs:
+  valgrind_tests:
+    # name: Valgrind tests
+    runs-on: 'ubuntu-latest'
+    steps:
+      - uses: actions/checkout@v2
```
hodgestar, line 13: Is there something that prevents using checkout v3 here?

.github/workflows/valgrind-tests.yml
```diff
@@ -0,0 +1,29 @@
+on:
+  workflow_call:
+    inputs:
+      portion:
+        required: true
```
hodgestar, line 5: I would drop the required keyword and instead add a default and a description, like so:
```
description: "Select portion of tests to run under Valgrind (default runs all):"
default: ""
```

hodgestar: (assuming description and default work for workflow_call)

hodgestar: We could instead have the Valgrind workflow be triggered by the successful run of the `ci.yml` workflow. It would still allow splitting the portions, but also skip Valgrind entirely if other tests fail.

.github/workflows/valgrind-tests.yml
fangerer: no, just forgot about it since I moved that to a separate file

.github/workflows/valgrind-tests.yml
fangerer: okay, sounds good to me

fangerer: > We could instead have the Valgrind workflow be triggered by the successful run of the ci.yml workflow. It would still allow splitting the portions, but also skip Valgrind entirely if other tests fail.

Let me do that in a separate PR. I first need to read through GH actions doc to learn how to do it.

hodgestar (approved):

PR #393, opened 2022-12-21
Multiphase module initialization

steve-s:
**I think the best way to review is to start from the changes in the docs/examples and tests.**

---
Make the multiphase module initialization the only way to initialize HPy modules, abandons the previous single phase initialization.

---
TL;DR: how module initialization looks like with this new API:

* empty module: no user provided initialization function, no call to `HPyModule_Create`, just "register" the `HPyModuleDef` in `HPy_MODINIT`.
```
HPyModuleDef mod = { ... }
HPy_MODINIT(my_extension_name, mod)
```

* when some module post-initialization is needed: use "exec" slot:
```
HPyDef_SLOT(exec1, HPy_mod_exec)
static int exec1_impl(HPyContext *ctx, HPy mod) {
    HPy_SetAttr_s(ctx, mod, "data", ctx->h_None);
    return 0;
}

static HPyDef *moduledefs[] = { &exec1, NULL };

HPyModuleDef mod = {
    .defines = moduledefs,
    ...
}

HPy_MODINIT(my_extension_name, mod)
```

---
HPy multiphase module initialization is a restricted version of what CPython API has. Currently both the CPython ABI and CPython HPy universal impl implement this via CPython's multiphase module initialization, but to some extent it can be emulated on top of CPython single phase module initialization if the need is (only custom `HPy_mod_create` slot would be disallowed in such case).

`HPy_MODINIT` takes name of the extension, so that it can generate `PyInit_XXX` for CPython, and `HPyModuleDef`. There is no user provided init function anymore. The runtime initializes the module and if the user wants to populate the module, or do any other initialization, they can define one or more `HPy_mod_exec` slots that are called by the runtime right after the module is created. The slots are defined in the same way as slots for types: by using HPyDef_SLOT macro.

As a consequence, `HPy_MODINIT(myextension, hpydef)` is all that is necessary if the module only exposes module functions.

Users can also define `HPy_mod_create` slot, similar to CPython's `Py_mod_create`, but in HPy this slot can return only objects that are not of the builtin module type.

The reasoning is that builtin module objects can be created by the runtime from `HPyModuleDef` and, for the time being, there is no use-case for manually creating builtin module objects in the `HPy_mod_create` slot or anywhere else. If such a use-case emerges, support for it can be easily added to HPy. (CPython unit tests create `SimpleNamespace` in the "create" slot and I could not find any other use of that slot anywhere on the internet...)

If user specifies the `HPy_mod_create slot`, specifying anything else (setting it to a non-default value) in `HPyModuleDef` is a runtime error (likely a bug or misunderstanding on the user side).

Additionally, this PR removes `HPyModule_Create` API/ABI function. The CPython equivalent `PyModule_Create` works only for single phase module initialization. We may add equivalents of `PyModule_FromDefAndSpec` to allow manual module creation if there is a use-case. Otherwise only the runtime can create a builtin module object and only from the provided `HPyModuleDef`.

Additionally, removes the CPython semantics of negative `m_size`. HPy will add other more explicit means to communicate subinterpreters support -- I think we should just have explicit fields/flags field in `HPyModuleDef` for that, I will add it in a followup PR.


steve-s:
rebased on master and fixed conflicts

docs/porting-example/steps/step_01_hpy_legacy.c
```diff
@@ -143,36 +143,47 @@ static HPyType_Spec Point_Type_spec = {
...
 static HPyDef *module_defines[] = {
+    &module_exec,
     NULL
 };
 
 static HPyModuleDef moduledef = {
-    .name = "step_01_hpy_legacy",
+    // .name = "step_01_hpy_legacy",
+    // ^-- .name is not needed for multiphase module initialization,
+    // it is always taken from the ModuleSpec
```
mattip, line 178: What happens if `.name` is provided anyway? Is this suggestion true?
```suggestion
    // it is always taken from the ModuleSpec. It can optionally be provided
    // here but must match
```

docs/porting-example/steps/step_01_hpy_legacy.c
```diff
@@ -143,36 +143,47 @@ static HPyType_Spec Point_Type_spec = {
...
     .doc = "Point module (Step 1; All legacy methods)",
-    .size = -1,
+    .size = 0,
```
mattip, line 180: Why this change?

hpy/devel/include/hpy/hpymodule.h
```diff
@@ -19,9 +19,18 @@ extern HPyContext *_ctx_for_trampolines;
...
+ * The name if always taken from the ModuleSpec (PEP 451).
+ */
 typedef struct {
-    const char* name;
     const char* doc;
+    /**
+     * Size of the module state structure. If set to zero, then the module
+     * will not get allocated and assigned any HPy module state. Negative
+     * size, unlike in CPython API, does not have any specific meaning and
+     * will produce a runtime error.
```
mattip, line 32: Ahh, this is the answer to my question above. Is this rendered somewhere in the html documentation?

mattip (commented):
I don't have much to say about the code changes, they all seem to follow from the desired new outcome. Is there any way to banchmark this to see if it changes anything?

docs/porting-example/steps/step_01_hpy_legacy.c
steve-s: There's no `name` field in the struct anymore in HPy. CPython always ignores `name` in `PyModuleDef` in multi-phase init. I assume they keep it there only for backwards binary compatibility (and single phase init).

hpy/devel/include/hpy/hpymodule.h
steve-s: Yes that's good point. I did cursory review of the docs and fixed all examples that are compiled/tested, but I should do more thorough pass over it! However, I suggest I do that on top of @fangerer current documentation improvements PR (and merge this PR as-is w.r.t. documentation) to reduce conflicts if that sounds good?

mattip: Sounds good, thanks

docs/porting-example/steps/step_01_hpy_legacy.c
mattip: In that case maybe this should state "it can be provided here but will be ignored"?

mattip:
I merged the documentation PR, and as predicted now this has conflicts, sorry.

steve-s:
> Is there any way to banchmark this to see if it changes anything?

This should have an effect only on module creation/initialization, which I assumed is typically not on the hot-path, and it should not make it significantly more complex algorithmically, so it did not seem to me this needs extra performance check.

> I merged the documentation PR, and as predicted now this has conflicts, sorry.

np, it's actually better. In the end, the conflicts pointed to exactly the things that I'd have to update anyway.

I've rebased the PR and updated the documentation. I tried to keep the already reviewed changes in one commit and added the new docs improvements in the new separate commit.

mattip (commented):
LGTM, thanks. Since this changes the public API in a non-backward compatible way, it should increment the major API number, right?

steve-s:
>  Since this changes the public API in a non-backward compatible way, it should increment the major API number, right?

Yes, that's a good question. My take was that until we finish the first milestone and release it, we allow ourselves to break binary compatibility.

fangerer:
IMO, we should just increase the *major API version* when releasing.
I don't think that we need or even should give compatibility guarantees on `master`.

mattip:
Thanks @steve-s

Issue #355, opened 2022-09-22
Clarify encoding used for C strings in the API.


markshannon:
It is not clear what encoding is used for C strings in the API. Are they ASCII, latin-1, or utf-8?

Either make this clear in the docs (maybe it is, but my cursory inspection didn't find it), or better still in the API itself.
A simple typedef would suffice.
```C
typedef const char *Utf8String;
```
This would also make it clearer what is text data, and what is just bytes.

In `HPyBytes_FromStringAndSize(HPyContext *ctx, const char *v, HPy_ssize_t len)`, `v` is just an array of bytes,
but in `HPyImport_ImportModule(HPyContext *ctx, const char *name)` `name` is a name.
The name gives a clue, but the encoding is hidden.
`HPyImport_ImportModule(HPyContext *ctx, Utf8String name)` is much clearer.

timfel:
I agree, I think [utf8everywhere](https://utf8everywhere.org) is the right approach and we should make that clear in the name

fangerer:
We decided in [today's dev call](https://github.com/hpyproject/hpy/wiki/dev-call-20221201) that we will just do that.

PR #392, opened 2022-12-20
Test upgrade to macos-12.

fangerer:
**DO NOT MERGE**

According to [this GitHub announcenment](https://github.com/actions/runner-images/issues/6384), `macos-latest` will switch to MacOS 12 (Monterey) next year. I just want to test if we run into problems.

PR #394, opened 2022-12-29
Use fixed-width integers and clarify encoding of C strings.

### LINKED ISSUE #343: Use fixed width integers (at least in ABI)
What is the thinking w.r.t. types such as `int`, `long`, which can vary in size even between different compilers even on the same system? Shouldn't the ABI be defined in terms of unambiguous fixed width types such as `int32_t`?

The question is then what the API should expose. I am not sure if switching everything to fixed width integers would be good porting experience, maybe it's fine? Without switching to fixed width integers in user code, people would see compiler warnings for conversions if they use (with `-Wconversion`), but maybe that's a good thing and it would guide them to fix them.

Alternative: the API can still expose generic `int` etc. and the trampolines would do the conversions, but then what if the ABI result does not fit into the API type (i.e., ABI returns `int32_t`, API returns `int` and `sizeof(int) == 16` for some compiler). We could fail at compilation time in such situation.

Example of the conversion warnings:
```
$ cat /tmp/test.c 
#include <stdint.h>
int32_t do_int() { return 42; }
int64_t do_int64() { return 42; }
int main() {
	int res1 = do_int();
	int res2 = do_int64();
	int16_t res3 = do_int64();
	return res1 + res2 + res3;
}

$ gcc -Wconversion /tmp/test.c 
/tmp/test.c: In function ‘main’:
/tmp/test.c:6:13: warning: conversion from ‘int64_t’ {aka ‘long int’} to ‘int’ may change value [-Wconversion]
    6 |  int res2 = do_int64();
      |             ^~~~~~~~
/tmp/test.c:7:17: warning: conversion from ‘int64_t’ {aka ‘long int’} to ‘int16_t’ {aka ‘short int’} may change value [-Wconversion]
    7 |  int16_t res3 = do_int64();
      |                 ^~~~~~~~
```

fangerer:
Resolves #343 and #355 .

As we decided in the [last HPy dev call](https://github.com/hpyproject/hpy/wiki/dev-call-20221201), we want to use fixed-width integers (where appropriate) instead of `int`/`long` in the API/ABI because the sizes of `int`/`long` could vary depending on several properties.

Since `int` is actually mostly used as a flag, status code, or small enum values (like in `HPy_RichCompare` argument `int op`), this PR is mainly about `HPyLong_From(Long|LongLong)`.

I did some research about the guarantees and found this: https://en.cppreference.com/w/c/language/arithmetic_types

Based on this document, I assume that `int` has at least 2 bytes, `long` at least 4 bytes, and `long long` at least 8 bytes.
I came to the conclusion that instead of `long` and `long long`, I'll use `int32_t` and `int64_t` because a smaller type could still be beneficial in some cases.

I then implemented the previous APIs like `HPyLong_FromLong` as inline helpers delegating to `HPyLong_From(Int32|Int64)` depending on `sizeof(long)`. Since the size of a type is compile-time constant, the dispatching overhead should be zero. 
In case of CPython, `HPyLong_From(Int32|Int64)` again delegates to `PyLong_From(Long|LongLong)` depending on the type sizes (again, compile-time constant).

The second change in this PR makes the encoding of C strings in the HPy API clear. We use UTF-8 everywhere. I checked if the `const char *` arguments are then really consumed by some UTF-8 conversion which is in most cases `HPyUnicode_FromString`.


mattip:
On what systems are the macros `#if SIZEOF_LONG >= SIZEOF_INT32` and `#if SIZEOF_LONG_LONG >= SIZEOF_INT64` ? Or in other words, where will CPython go via the slower c-api calls?

fangerer:
> On what systems are the macros `#if SIZEOF_LONG >= SIZEOF_INT32` and `#if SIZEOF_LONG_LONG >= SIZEOF_INT64` true? Or in other words, where will CPython go via the slower c-api calls?

You mean: on what system are conditions `SIZEOF_LONG >= SIZEOF_INT32` and `SIZEOF_LONG_LONG >= SIZEOF_INT64` **false** ? Because they are true on, e.g. my system since `sizeof(long) == sizeof(long long) == 8` these days.

TBH, I don't know. I just tried to cover all possible cases according to the minimal sizes guaranteed by https://en.cppreference.com/w/c/language/arithmetic_types. The slower C API calls will go to e.g. `ctx_Long_FromUInt32_t` (in `ctx_long.c`) and currently, it will also just delegate to `PyLong_FromLong` but the difference is: if `SIZEOF_LONG >= SIZEOF_INT32` is **false**, then you can't even build this function. We could, ofc, also do a range check and that case but I thought that can be future work if we ever need it.

mattip:
OK, thanks. Even though this is a big renaming, I like it. The names are now much more self-explanatory.

hpy/tools/autogen/public_api.h
```diff
@@ -115,30 +115,30 @@ HPy_ID(79)
...
-HPy HPyLong_FromUnsignedLong(HPyContext *ctx, unsigned long value);
+HPy HPyLong_FromUInt32_t(HPyContext *ctx, uint32_t value);
 HPy_ID(82)
-HPy HPyLong_FromLongLong(HPyContext *ctx, long long v);
+HPy HPyLong_FromInt64_t(HPyContext *ctx, int64_t v);
 HPy_ID(83)
-HPy HPyLong_FromUnsignedLongLong(HPyContext *ctx, unsigned long long v);
+HPy HPyLong_FromUInt64_t(HPyContext *ctx, uint64_t v);
 HPy_ID(84)
 HPy HPyLong_FromSize_t(HPyContext *ctx, size_t value);
```
steve-s, line 126: Since `size_t` is implementation-defined, do we want to do the same for it as for `int`/`long`?

hpy/tools/autogen/public_api.h
```diff
@@ -115,30 +115,30 @@ HPy_ID(79)
...
 HPy_ID(85)
 HPy HPyLong_FromSsize_t(HPyContext *ctx, HPy_ssize_t value);
```
steve-s, line 128: Maybe unlike CPython we should define `HPy_ssize_t` to be always an alias to, e.g., `uint64_t`?

hpy/tools/autogen/public_api.h
```diff
@@ -154,7 +154,7 @@ HPy_ID(97)
 double HPyFloat_AsDouble(HPyContext *ctx, HPy h);
 
 HPy_ID(98)
-HPy HPyBool_FromLong(HPyContext *ctx, long v);
+HPy HPyBool_FromBool(HPyContext *ctx, bool v);
```
steve-s, line 157: I wonder if we should have some `HPy_bool_t` to also remove the compiler-specific `bool`. Not sure how far we want to take this -- do we want to remove all sources of potential compile-specifics from ABI?

hpy/tools/autogen/public_api.h
mattip: I think we should do this in pieces. Since int/long and ascii/utf-8 are the most common places for confusion, let's just stick with those for this PR.

mattip:
There is a failure in cppcheck, which I think is causing the other CI runs to abort. Maybe we should be using `fail-fast: false` in all the non-essential CI runs.

hpy/tools/autogen/public_api.h
fangerer: That's a point and we probably should. This basically means that we would remove `ctx_Long_FromSize_t` from the context and just have an inline helper for it (which, I think, is fine).

hpy/tools/autogen/public_api.h
fangerer: > Maybe unlike CPython we should define `HPy_ssize_t` to be always an alias to, e.g., `uint64_t`?

This would be fine for me but just to consider the original reason for having `Py_ssize_t` in CPython (https://peps.python.org/pep-0353/): As far as I understand, there are two reasons. (1) CPython previously used `int` for indexing which is just too small on 64-bit systems, and (2) because the signed integral type `ssize_t` is not available everywhere.

From my point of view, point (1) implies that the type used for indexing should fit into one register (certainly for performance reasons). Consequently, we should maybe also have such a type but as far as I know, we don't run any tests on 32-bit systems right now. I doubt that HPy even works on 32-bit systems. This indicates that we should probably just remove `ctx_Long_FromSsize_t` from the context and provide an inline helper.

The more interesting part then is: how should we deal with all the other API functions that use `HPy_ssize_t`?
Anyway, this looks like a larger topic which I wouldn't tackle in this PR.

hpy/tools/autogen/public_api.h
fangerer: > the compiler-specific `bool`

As far as I know, the Boolean type (i.e. `_Bool` or `bool`) is part of the C standard since C99 (see, e.g.,  https://en.cppreference.com/w/c/types). That's why I thought that we could just have it in the ABI, in particular because `true` and `false` are now also defined to expand to `1` and `0`, respectively.

What we should definitively do is to explicitly make C99 the minimum required standard for HPy (right now, this is just an implicit requirement).

fangerer:
> There is a failure in cppcheck, which I think is causing the other CI runs to abort. Maybe we should be using `fail-fast: false` in all the non-essential CI runs.

@mattip: I just don't understand why cppcheck is failing. I tried to reproduce locally but it doesn't print me any useful message. There are a lot of messages with level `information` and then there is a message that not all includes could be resolved. Running with `--check-config` (as suggested by cppcheck) doesn't give any useful output. I already played around and added include paths and such but nothing helped.
Do you know how to debug that?

mattip:
> I just don't understand why cppcheck is failing

I have no idea. I suggest that we do not try to debug it, rather we make passing cppcheck optional since it is flaky and difficult to control. It is meant to help us, but turns out obstructing progress.

hpy/tools/autogen/public_api.h
thedrow: I agree that we should specify C99 as a minimum required standard.

thedrow:
> > I just don't understand why cppcheck is failing
> 
> I have no idea. I suggest that we do not try to debug it, rather we make passing cppcheck optional since it is flaky and difficult to control. It is meant to help us, but turns out obstructing progress.

#395 attempts to fix the CPPCheck failures.
We can disable it if it's not useful but it did catch a few bugs related to varargs before so I think it is somewhat useful at the very least.

fangerer:
> #395 attempts to fix the CPPCheck failures.

Well, that's my PR where I made it pass on my local box but it still fails in the GitHub CI. Hence, #395 is not (yet) the solution. I'll do some more investigations ...

mattip:
I would strongly suggest not spending the very limited time we have on the cppcheck CI run. Let it fail, and let the rest of CI continue. The problem right now is that failing cppcheck run stops the rest of the CI, which then means we have to fix cppcheck before we can do anything else. By adding `fail-fast: false` to all the non-essential checks, we can make faster progress.

fangerer:
Agreed. I'll do that.

mattip:
Thanks, now CI runs to completion. The failures do not seem to be related to this PR. Are they failing elsewhere as well?

fangerer:
> Thanks, now CI runs to completion. The failures do not seem to be related to this PR. Are they failing elsewhere as well?

Yes, in #392 . I've seen those problems on my local machine (Mac OS 12). I thought that the GitHub upgrade to Mac OS 12 is happening in March. I already started to look into the problem. I probably pin to Mac OS 11 for now.

PR #395, opened 2022-12-30
Fix and re-enable cppcheck

### LINKED ISSUE #188: Re-enable cppcheck

fangerer:
Resolves #188 .

This fixes and re-enables the `cppcheck` job.
I think the main reason why it broke before is that there was an update of the Ubuntu runner image (`ubuntu-latest` was updated to `ubuntu-22.04`) which also includes a newer cppcheck version.

Funnily, I first tried to reproduce the problem on Ubuntu 20.04 with cppcheck 1.90 and did fixes to make that run. Without changing HPy's source code, this did not pass on cppcheck 2.7 (which is available in Ubuntu 22.04). 
Long story short: the cppcheck job broke because of cppcheck. 

I therefore pinned the versions to Ubuntu 22.04 and cppcheck 2.7-1. This should work for a while (a few years).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -344,8 +344,6 @@ jobs:
     name: Cppcheck static analysis
     runs-on: 'ubuntu-20.04'
     continue-on-error: true
-    # Skip it, see issue gh-188
-    if: false
     steps:
       - uses: actions/checkout@v3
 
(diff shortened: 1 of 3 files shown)


mattip:
CI is green, let's give this a try. Thanks @fangerer

PR #398, opened 2023-01-30
Minor stylistic improvements to the top-level README

hodgestar:
Tweaks the opening paragraph of the README to avoid having a sentence with two colons in it.


mattip:
This is nice, but I think we should rewrite more. It should have less what we intend to do, and more what the project does. There should also be at least one link to the documentation.

fangerer (approved):

Issue #214, opened 2021-06-10
String builder API


antocuni:
XXX write me

TeamSpen210:
`pickle` actually contains what could be an excellent use case for a resizable builder. [`_Pickler_Write`](https://github.com/python/cpython/blob/304dfec8d3c0763734ea8b5fa2af1d9e1ce69ffa/Modules/_pickle.c#L1055-L1103) accumulates binary data into a bytes object, resizing it as required. (The frame part is unrelated, it's part of the pickle format.) In the case of file-dump mode, this is periodically ouputted, otherwise it collects the full pickle.

It's not necessarily in performance-critical code, but Cython uses an [inlined/optimised](https://github.com/cython/cython/blob/master/Cython/Utility/StringTools.c#L891-L959) version of `PyUnicode_Join` for [implementing](https://github.com/cython/cython/blob/0e4eeb27817577021ac3df4867200c1bc2ff6f11/Cython/Compiler/ExprNodes.py#L3242) F-strings.

steve-s:
Note I'd like to remove [docs/misc/str-builder-api.rst](https://github.com/hpyproject/hpy/pull/399/files#diff-61b4553a695a97d1ca1d19796c98e5c23c2b2846211ff641d99f60bc3fd16653) from user documentation, but leave a trace of it here.

PR #399, opened 2023-01-31
Small documentation improvements & quickstart guide

steve-s:


docs/api.rst
```diff
@@ -567,3 +546,40 @@ be considered in three places:
...
+contains forks of some popular Python extensions ported to HPy as
+a proof of concept/feasibility studies, such as the
+`Kiwi solver <https://github.com/hpyproject/kiwi-hpy>`_.
+Note that those forks may not be up to date with their upstream projects
+or with the upstream HPy changes.
+
+HPy unit tests
+~~~~~~~~~~~~~~
+
+HPy usually has tests for each API function. This means that there is lots of
```
steve-s, line 566: this paragraph is just shuffled at the end of this file

docs/index.rst
```diff
@@ -8,36 +8,58 @@ HPy: a better API for Python
...
+
+  - **Deployment**: it is possible to compile a single binary which runs unmodified on all
+    supported Python implementations and versions -- think "stable ABI" on steroids
+
+  - **Simplicity**: it is simpler and more manageable than the ``Python.h`` API, both for
+    the users and the Pythons implementing it
+
+  - **Debugging**: it provides an improved debugging experience. Debug mode can be turned
+    on at runtime without the need to recompile the extension or the Python running it.
+    HPy design is more suitable for automated checks.
```
steve-s, line 23: I shuffled this at the top, shortened the points, and added `think "stable ABI" on steroids`

docs/index.rst
```diff
@@ -8,36 +8,58 @@ HPy: a better API for Python
...
-    instead of refcounting, or to remove the GIL.
+  - experiment with new approaches inside CPython itself, for example:
 
-There are several advantages to write your C extension in HPy:
+    - use a tracing garbage collection instead of reference counting
+    - remove the global interpreter lock (GIL) to take full advantage of multicore architectures
+    - use tagged pointers to reduce memory footprint
+
+Where to go next:
+-----------------
```
steve-s, line 40: I think this could pitch some parts of the documentation depending on what you're looking for: either code samples or some more explanations/details

docs/misc/index.rst
```diff
@@ -2,8 +2,6 @@ Misc notes
 ==========
 
 .. toctree::
-   :maxdepth: 2
+   :maxdepth: 1
 
-   str-builder-api
    embedding
-   protocols
```
steve-s, line 9: These two were not a real documentation, but design notes. I don't think it fits in the docs.

docs/api-reference/inline-helpers.rst
```diff
@@ -5,5 +5,9 @@ Inline Helpers
 Those functions are usually small convenience functions that everyone could
 write but in order to avoid duplicated effort, they are defined by HPy.
 
+One category of inline helpers are functions that convert the commonly used
+but not fixed width C types, such as ``int``, or ``long long``, to HPy API
+that uses only fixed with types in order to ensure maximal ABI compatibility.
```
mattip, line 10:
```suggestion
but not fixed width C types, such as ``int``, or ``long long``, to HPy API.
The HPy API always uses well-defined fixed width types like ``int32`` or
``unsigned int8``.
```

docs/index.rst
```diff
@@ -8,36 +8,58 @@ HPy: a better API for Python
 
 HPy provides a new API for extending Python in C.
 
+There are several advantages to write your C extension in HPy:
```
mattip, line 11:
```suggestion
There are several advantages to writing C extensions in HPy:
```

docs/misc/index.rst
steve-s: I created an issue for the protocols: https://github.com/hpyproject/hpy/issues/400, and added a comment to the existing str builder issue: https://github.com/hpyproject/hpy/issues/214

docs/index.rst
```diff
@@ -8,36 +8,58 @@ HPy: a better API for Python
...
+  - implement it for other Python implementations (e.g. PyPy, GraalPy,
+    Jython, ...)
 
-  - to experiment with new things inside CPython itself: e.g. using a GC
-    instead of refcounting, or to remove the GIL.
+  - experiment with new approaches inside CPython itself, for example:
 
-There are several advantages to write your C extension in HPy:
+    - use a tracing garbage collection instead of reference counting
+    - remove the global interpreter lock (GIL) to take full advantage of multicore architectures
```
mattip, line 36: Is this true? Does using HPy make it any easier to be sure code does not have threading problems?

docs/misc/str-builder-api.rst
```diff
@@ -1,443 +0,0 @@
-bytes/str building API
-=======================
```
mattip, line 2: Maybe move this into a "design document" section?

docs/quickstart.rst
```diff
@@ -0,0 +1,52 @@
+HPy Quickstart
+=============
```
mattip, line 2: Doesn't this duplicate much of the [simple example](https://docs.hpyproject.org/en/latest/api.html#a-simple-example)

mattip (commented):
Nice improvements. I suggested a few small changes. I would be fine merging this as-is and considering the quickstart vs. the simple-example later.

I think we should keep the string builder document somewhere, at least until we have a string builder API.

docs/misc/index.rst
mattip: I saw this comment only after my review. Maybe we could move the string stuff to the wiki?

docs/quickstart.rst
steve-s: Yes, but I think the two can serve slightly different purposes: quickstart -- when I am impatient, I want hands on experience right now, I don't mind following instructions blindly for now, simple example -- when I want some additional context and explanations. I could not resist adding few explanatory comments to quickstart, but I tried to keep them very short.

docs/misc/str-builder-api.rst
steve-s: I've linked it to the relevant GitHub issue. I think that having design documents in the "main" user documentation is not good, because:
* it can be misleading and create false expectations
* we should primarily use GitHub issues (or something else, but a single location -- I haven't read it, but I suspect that `str-builder-api.rst` is not up-to-date, anyway)

docs/index.rst
steve-s: Yes, good point. This bullet point was here before, I just shuffled it and added "... to take full advantage of multicore architectures".

> HPy make it any easier to be sure code does not have threading problems

Depends on the point of view. My line of thinking is that it would make it easier to do the internal changes in CPython that are needed to remove the GIL. Still extensions would have to make sure that they operate correctly without GIL.  

I think it still falls into the category of "what is wrong with current C API", although better HPy would not solve the problem as a whole, only a part of it. Explaining all this in detail would make this bullet point too long. I like that "removing GIL" is something that many would clearly understand brings benefits and there is bit of a hype around it now :-) I think this simplification could be fine for such a pitch as the beginning of the documentation?

mattip:
Cool, thanks

PR #401, opened 2023-02-01
Autogen code also for HPyFunc_NOARGS and HPyFunc_NOARGS_O

steve-s:
It seems that there is no reason to not autogenerate them? Maybe they were historically manually written and stayed that way when autogeneration was added? In any case, all tests seem to pass :-)


mattip:
This now has conflicts. Otherwise LGTM

PR #403, opened 2023-02-06
Introduce HPy_Compile_s and HPy_EvalCode.

### LINKED ISSUE #312: Provide eval API
HPy does still not provide all necessary API functions to be able to migrate an arbitrary module.
We regularly hit this problem when migrating modules and keep adding the missing functions.
However, in particular for external people, it might sometimes be a blocker. Also, often it's not good to just add an exact equivalent API function for the missing C API function.

It would therefore be good to have the possibility to eval (in Python land known as: _compile and eval_) some Python code and receive the result.

CPython provides several functions for that: https://docs.python.org/3/c-api/veryhigh.html

I think what is already sufficient for this case are equivalent functions for `Py_CompileString` and `PyEval_EvalCode`.

fangerer:
Resolves #312 .

Also discussed in the [last dev call](https://github.com/hpyproject/hpy/wiki/dev-call-20230202#decisions-for-release).

### DIFF
--- a/hpy/devel/include/hpy.h
+++ b/hpy/devel/include/hpy.h
@@ -226,7 +226,7 @@ typedef enum {
      * Parse a single statement. This is the mode used for the interactive
      * interpreter loop.
      */
-    HPy_SourceKind_Stat = 2,
+    HPy_SourceKind_Single = 2,
 } HPy_SourceKind;
 
 #ifdef HPY_ABI_CPYTHON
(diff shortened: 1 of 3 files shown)


mattip (approved):
Seems quite straight-forward, and the documentation is clear. One minor nit about the enum name, feel free to ignore.

fangerer:
I seems that the Ubuntu mirror is having problems. That's why so many CI jobs are failing.

PR #407, opened 2023-02-23
ENH: add HPyContextVar* interfaces and document how it is done

mattip:
I added the missing `HPyContextVars_*` API functions.
- I documented what I did in `docs/api-reference/index.rst`
- I changed the following files then ran `make autogen, which failed:
  - `hpy/tools/autogen/public_api.h`
  - `hpy/tools/autogen/conf.py`
- I added `HPyContextVar_Get` to ``NO_WRAPPER`` in `hpy/tools/autogen/debug.py`, and then needed to modify/add these files:
  - `hyp/devel/src/runtime/ctx_contextvars.c` (and added that to `setup.py`)
  - `hpy/debug/src/debug_ctx.c`
  - `hpy/devel/include/hpy/runtime/ctx_funcs.h`
  - (edit) `hpy/devel/include/hpy/cpython/misc.h`
- I added a test. It passes. ~universal and debug mode, but fails to compile on cpython mode: it is missing the declaration for `HPyContextVar_Get`. Where do I need to add it?~


hpy/debug/src/debug_ctx.c
```diff
@@ -523,3 +523,16 @@ int debug_ctx_TypeCheck(HPyContext *dctx, DHPy obj, DHPy type)
...
+
+int32_t debug_ctx_ContextVar_Get(HPyContext *dctx, DHPy context_var, DHPy default_value, DHPy *result)
+{
+    HPyContext *uctx = get_info(dctx)->uctx;
+    UHPy uh_context_var = DHPy_unwrap(dctx, context_var);
+    UHPy uh_default_value = DHPy_unwrap(dctx, default_value);
+    UHPy uh_result;
+    assert(!HPy_IsNull(uh_context_var));
+    int32_t ret = HPyContextVar_Get(uctx, uh_context_var, uh_default_value, &uh_result);
+    if (ret < 0) return ret;
```
fangerer, line 535: I think we need to write ´HPy_NULL` to `*result` even in the error case.

hpy/tools/autogen/conf.py
```diff
@@ -139,4 +139,7 @@
     'HPyBool_FromBool': 'PyBool_FromLong',
     'HPy_Compile_s': None,
     'HPy_EvalCode': 'PyEval_EvalCode',
+    'HPyContextVar_New': 'PyContextVar_New',
```
fangerer, line 142: this one and the one for `HPyContextVar_Set` is not necessary; autogen will do that automatically

hpy/devel/src/runtime/ctx_contextvar.c
```diff
@@ -0,0 +1,17 @@
...
+   // for _h2py and _py2h
+#  include "handles.h"
+#endif
+
+_HPy_HIDDEN int32_t
+ctx_ContextVar_Get(HPyContext *ctx, HPy context_var, HPy default_value, HPy *result)
+{
+    PyObject * py_result = NULL;
+    int32_t ret = PyContextVar_Get(_h2py(context_var), _h2py(default_value), &py_result);
+    if (ret < 0) return ret;
```
fangerer, line 14: This branch can just go away. In case `ret < 0` we will have `py_result == NULL` and `_py2h` will correctly convert that to `HPy_NULL` which we need to write to `*result` in case of an error.
You can then also remove the initialization of `py_result = NULL`.

docs/api-reference/index.rst
```diff
@@ -66,3 +66,25 @@ write but in order to avoid duplicated effort, they are defined by HPy.
    :maxdepth: 2
 
    inline-helpers
+
+Adding New API
+==============
+
+1. Add the function to ``hpy/tools/autogen/public_api.h`` and add the mapping
+   from the function to the cpython equivalent to ``hpy/tools/autogen/conf.py``.
```
fangerer, line 74: The explicit mapping only needs to be done if the HPy and C API functions do not have the same name (except of the leading `H` in the HPy case). But for description, it is probably easier to put it this way.

docs/api-reference/index.rst
```diff
@@ -66,3 +66,25 @@ write but in order to avoid duplicated effort, they are defined by HPy.
...
+   ``hyp/devel/src/runtime/*.c`` (possibly adding the new file to ``setup.py``),
+   add a debug wrapper to ``hpy/debug/src/debug_ctx.c``, add a implementation
+   that uses the `ctx` variant to ``hpy/devel/include/hpy/cpython/misc.h`` and
+   add the declaration to ``hpy/devel/include/hpy/runtime/ctx_funcs.h``.
+
+3. Run ``make autogen`` which will turn the mapping into autogenerated functions
+
+4. Add a test for the functionality
+
+5. Build with ``python setup.py build_ext``. After that works, build with
```
fangerer, line 88: You should use `make` instead of `python setup.py build_ext`. If you really want to directly run `setup.py`, then it currently is:
```
python3 setup.py build_clib -f build_ext -if
```
Reason: `build_clib` will also build the static lib for the context that will speed up the tests a lot. But yes, your comamnd still works.

Also, installation is not required to run tests if you use `make` (which builds in-place).

fangerer (changes requested):
Thanks, @mattip, for this PR. I added some comments.
Most of my comments are not _strict_ except of the once where we need to write to the output var even in case of errors.

docs/api-reference/index.rst
mattip: I don't think `make` works on windows. I use the minimal `build_ext` on purpose, since it is the quickest way to flush out build errors, and once that works recommend using the command used in the CI jobs.

hpy/debug/src/debug_ctx.c
mattip: good catch

hpy/devel/src/runtime/ctx_contextvar.c
mattip: OK

hpy/tools/autogen/conf.py
mattip: OK

fangerer (approved):

docs/api-reference/index.rst
fangerer: I agree, `make` won't probably work on Windows (I think it could but that's rather tricky, irrc). Anyway, using `build_ext`to get compilation errors is fine. You can also run all tests, it's just slower.

PR #410, opened 2023-03-02
Introduce HPyType_GetName and HPyType_IsSubtype.

fangerer:
These new API functions are needed for Numpy/HPy. I recommend reviewing commit by commit.

The equivalent functions in CPython take `PyTypeObject *` as arguments which we cannot do in HPy but as we already do with `HPy_TypeCheck`, we check that in the debug mode.

In case of `HPyType_GetName`, I decided to return the type's name as a UTF-8 encoded C string instead of returning a Unicode object as CPython does. Reason: If a C extension wants to retrieve the name of a type, I assume it will work with the name in C. Also, there is a perfect option to get the type's name as Unicode object, which is: `HPy_GetAttr_s(ctx, type, "__name__")`.

This PR also contains a little doc improvement: The HPy API function index is now sorted alphabetically.


hpy/tools/autogen/public_api.h
```diff
@@ -402,6 +402,54 @@ HPy HPy_Type(HPyContext *ctx, HPy obj);
...
+ *     case, the behavior is undefined (verification of the argument is only
+ *     done in debug mode).
+ *
+ * :returns:
+ *     The name of the type as C string (UTF-8 encoded) or ``NULL`` in case of
+ *     an error. The returned pointer is read-only and guaranteed to be valid as
+ *     long as the handle ``type`` is valid.
+ */
+HPy_ID(253)
+const char *HPyType_GetName(HPyContext *ctx, HPy type);
```
mattip, line 426: Don't we have a utf8_t typedef to indicate the encoding?

fangerer: No, we didn't use that but named the arguments like `utf8_something`. That's, ofc, not possible for return types. The docs give the necessary hint. I hope that is enough.

mattip: OK, thanks

mattip:
LGTM, in it goes

PR #411, opened 2023-03-02
Introduce HPyUnicode_FromEncodedObject and HPyUnicode_Substring.

fangerer:
Also used in the Numpy/HPy port (builds on PR #410).
I recommend reviewing commit by commit.

### DIFF
--- a/hpy/tools/autogen/public_api.h
+++ b/hpy/tools/autogen/public_api.h
@@ -625,6 +625,22 @@ HPy HPyDict_New(HPyContext *ctx);
 HPy_ID(257)
 HPy HPyDict_Keys(HPyContext *ctx, HPy h);
 
+/**
+ * Creates a copy of the provided Python dict object.
+ *
+ * :param ctx:
+ *     The execution context.
+ * :param h:
+ *     A Python dict object. If this argument is ``HPy_NULL`` or not an
+ *     instance of a Python dict, a ``SystemError`` will be raised.
+ *
+ * :returns:
+ *     Return a new dictionary that contains the same key-value pairs as ``h``
+ *     or ``HPy_NULL`` in case of an error.
+ */
+HPy_ID(258)
+HPy HPyDict_Copy(HPyContext *ctx, HPy h);
+
 /* tupleobject.h */
 HPy_ID(203)
 int HPyTuple_Check(HPyContext *ctx, HPy h);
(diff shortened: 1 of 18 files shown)


hpy/tools/autogen/public_api.h
```diff
@@ -406,7 +406,7 @@ int HPy_TypeCheck(HPyContext *ctx, HPy obj, HPy type);
  * Return the type's name.
  *
  * Equivalent to getting the type's ``__name__`` attribute. If you want to
- * retrieve the type's name as Python unicode object, then just use
+ * retrieve the type's name as Python Unicode object, then just use
```
mattip, line 409: PyUnicodeObject?

mattip:
Somehow PR #412 appears together in this PR. Is that on purpose?

It might be nice to add more tests for the cases where the encoding or error are missing or where `start` < `end`

fangerer:
> Somehow PR #412 appears together in this PR. Is that on purpose?

Yes, sorry, for the confusion. This is a chain of PRs to make reviewing easier and avoid conflicts. The chains is:
#412 -> #411 -> #410 -> master

That was probably not a good idea. I won't do that in future.

> It might be nice to add more tests for the cases where the encoding or error are missing or where `start` < `end`

Okay, I'll add some more.

hpy/tools/autogen/public_api.h
fangerer: I wouldn't use `PyUnicodeObject` because this type does not exist in HPy. Should I write `Python string`?

mattip: I don't know. We need to have a terminology that makes sense, this is probably not the only place where we need this kind of construct. It is not a python `str` object since those do not exist in C, neither is it a `PyUnicodeObject`, so what is it?

fangerer: IMO, it is a handle that refers to a Python string. The [Python doc](https://docs.python.org/3/library/stdtypes.html#textseq) calls it: _str object_.

mattip: So "retrieve the type's name as a handle that refers to a `str`, then just use..."

It should be clear that handles refer to python objects.

mattip (approved):
LGTM. I still see the commits from #412 that was already merged. Strange.

fangerer:
> I still see the commits from #412 that was already merged. Strange.

That's fine because I've opened #412 to the source branch of this PR. So, we are now merging both into master.

PR #412, opened 2023-03-02
Introduce HPyDict_Keys and HPyDict_Copy.

fangerer:
Also used in the Numpy/HPy port (builds on PR #411).
I recommend reviewing commit by commit.


mattip:
LGTM. Simple and to the point. Maybe the tests could use better function names: `f` is a bit too simplistic, but no big deal.

mattip:
cppcheck is failing, as always

mattip:
Thanks @fangerer

PR #413, opened 2023-03-02
skip cppcheck

mattip:
Related to #188 and #395. The check has been failing for a few months. It seems to check way too much


fangerer (approved):
I absolutely agree because if the check is not a blocker and we know that it is failing, no one is ever looking at the log output to check if there are new problems.  So it is useless.

PR #414, opened 2023-03-08
Remove support for Python 3.7.

fangerer:
Python 3.7 is soon EOL (see https://devguide.python.org/versions/).
This is also a preparation for the vectorcall PR #390 because 3.7 did not have vectorcall API yet.


mattip:
I think this line needs to change as well https://github.com/hpyproject/hpy/blob/6b54518266a2b857f42712838db797d75aa9f7f2/setup.py#L266

mattip:
Thanks @fangerer

PR #415, opened 2023-03-08
Add struct sequence API as HPy helpers.

fangerer:
This PR adds some HPy helper functions to work with _struct sequences_.
Since CPython's struct sequences and namedtuple are very similar, the idea is to map the API to CPython's `PyStructSequence_*` functions in CPython ABI mode and to use `namedtuple` in universal ABI mode.

However, there are subtle differences regarding the hidden elements and the error behavior. HPy's struct sequence API is therefore a little bit smaller.

I admit, the implementation for the universal mode doesn't look very nice but the big advantage is that this won't need any more implementation on the runtime side.


hpy/devel/include/hpy/runtime/structseq.h
```diff
@@ -0,0 +1,190 @@
+#ifndef HPY_COMMON_RUNTIME_STRUCTSEQ_H
+#define HPY_COMMON_RUNTIME_STRUCTSEQ_H
+
+#include "hpy.h"
+
+/*
+ * Struct sequences are relatively simple subclasses of tuple, so we provide the
```
mattip, line 7:
```suggestion
 * Struct sequences are subclasses of tuple, so we provide a
```

hpy/devel/include/hpy/runtime/structseq.h
```diff
@@ -0,0 +1,190 @@
...
+
+    /**
+     * Docstring of the field (UTF-8 encoded); may be ``NULL``.
+     */
+    const char *doc;
+} HPyStructSequence_Field;
+
+/**
+ * Contains the meta information of a struct sequence type to create.
+ * Struct sequences are subclasses of tuple. All fields are Python objects. The
```
mattip, line 36: Is this a copy-paste or did you really mean that the fields are Python objects?

test/test_hpystructseq.py
```diff
@@ -0,0 +1,132 @@
...
+            {
+                HPyStructSequenceBuilder builder = HPyStructSequenceBuilder_New(ctx, type);
+                if (HPyStructSequenceBuilder_IsNull(builder))
+                    return HPy_NULL;
+                HPy tmp = HPyLong_FromLong(ctx, 1);
+                HPyStructSequenceBuilder_Set(ctx, builder, 0, tmp);
+                HPy_Close(ctx, tmp);
+                HPyStructSequenceBuilder_Set_i(ctx, builder, 1, 2);
+                HPyStructSequenceBuilder_Set_i(ctx, builder, 2, 3);
+                return HPyStructSequenceBuilder_Build(ctx, builder, type);
```
mattip, line 50: What happens if a different `type` is used here? Shouldn't the `type` be redundant since it was used in `HPyStructSequenceBuilder_New`?

mattip (commented):
Why do we need the calls to`HPyStructSequenceBuilder_Set*`? Since the spec is passed in, couldn't the call to `HPyStructSequenceBuilder_New` fill in the fields?

hpy/devel/include/hpy/runtime/structseq.h
fangerer: Right, I've basically copied that from the CPython docs.
If the phrase confuses you, I think we can just remove it. As in the same line stated: a struct sequence is a subclass of tuple. So it should be clear what are the elements of it.

test/test_hpystructseq.py
fangerer: > What happens if a different type is used here?

If the types then have a different number of elements, construction would not work. Depending on the ABI mode, it would maybe just crash or raise an error. Hence, behavior is undefined.

I agree, it shouldn't be necessary to pass the type to `HPyStructSequenceBuilder_Build`. And yes, behavior should be defined and consistent. Sorry, I was already a bit tired of merging behavior of `namedtuple` and `structseq`.

fangerer:
> Why do we need the calls to`HPyStructSequenceBuilder_Set*`? Since the spec is passed in, couldn't the call to `HPyStructSequenceBuilder_New` fill in the fields?

We could, ofc, do that but it's a question of convenience. I think in Numpy it wouldn't be a big deal. This is how we use it in Numpy/HPy right now (in file [typeinfo.c](https://github.com/hpyproject/numpy-hpy/blob/graal-team/hpy-0.0.5/numpy/core/src/multiarray/typeinfo.c#L59)):
``` 
    HPyStructSequenceBuilder entry = HPyStructSequenceBuilder_New(ctx, PyArray_typeinfoType);
    HPyStructSequenceBuilder_Set_i(ctx, entry, 0, typechar);
    HPyStructSequenceBuilder_Set_i(ctx, entry, 1, typenum);
    HPyStructSequenceBuilder_Set_i(ctx, entry, 2, nbits);
    HPyStructSequenceBuilder_Set_i(ctx, entry, 3, align);
    HPyStructSequenceBuilder_Set(ctx, entry, 4, type_obj);
```
In this case, you would need to allocate a temporary array (could be on stack).

You are probably right, we should decouple the builder and creating the actual instance. I'll introduce a `HPyStructSequence_New(HPyContext *ctx, HPy type, HPy_ssize_t n, HPy const *args)` to be able to create the instance from an array.
I'll probably just kill the builder for now since I don't see a big use case for that right now.

hpy/devel/src/runtime/structseq.c
```diff
@@ -0,0 +1,396 @@
...
+
+static HPy
+build_array(HPyContext *ctx, HPy type, const char **fmt, va_list *values, HPy_ssize_t size)
+{
+#ifdef _WIN32
+    HPy *arr = _malloca(size);
+    HPy result;
+#else
+    HPy arr[size];
+#endif
```
mattip, line 363: Why use `_malloca` on windows, it seems to complicate the code paths?

fangerer: Right but I was reading through [the manual](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/alloca?view=msvc-170) and it says:
```
Allocates memory on the stack. This function is deprecated because a more secure version is available; see _malloca.
```
The problem is: if `size` is unexpectedly large, then you would kill the stack (since max stack sizes are usually much smaller than heap sizes). `_malloca` is actually exactly that function I would also prefer on Linux and other systems. It decides whether to allocate on stack or heap depending on the size. But I haven't seen such a function.

IMO, the increased complexity isn't that much. We need to call `_freea` (like we would need to do if we'd used `malloc`). I actually plan to use `_malloca` everywhere instead of `_alloca`.

mattip: using `_malloca` (or `_alloca`) only on windows does not solve the stack problem, since it does not cover uses on non-windows. If there is no generic solution, I think I would prefer not to create different code paths for different runtme libraries. If you insist on using this, you should guard it with `#ifdef _MSC_VER` to prevent using this with non-MSVC compilers on windows.

steve-s: Note that we also use plain `alloca` elsewhere and just assume the stack has enough space. Maybe we should have some thin abstraction (also for `strncpy`, `snprintf`, ...), but not necessarily in this PR.

fangerer: > Maybe we should have some thin abstraction (also for strncpy, snprintf, ...), but not necessarily in this PR.

Agreed. I was also thinking about that. Should be easy for `strncpy`, `snprint` and such but for `_malloca`, it's not that easy. That needs to be a macro but that should also be possible.

Anyway, didn't do that since I was concentrating on getting the APIs for Milestone 1 in. I would do such enhancements after that.

fangerer: > using _malloca (or _alloca) only on windows does not solve the stack problem, since it does not cover uses on non-windows. If there is no generic solution, I think I would prefer not to create different code paths for different runtme libraries. 

You are right, we should have the same behavior as far as possible.

> If you insist on using this, you should guard it with #ifdef _MSC_VER to prevent using this with non-MSVC compilers on windows.

No, I don't insist on that. I'll remove it and will come up with a proper solution as Stepan suggested.

fangerer:
@mattip I've removed `HPyStructSequence_NewFromFormat` (which also removes usage of `_malloca`). As already mentioned, the structseq API cannot be simpler than that.

This PR also includes the promised fixup for #416 (see https://github.com/hpyproject/hpy/pull/415/commits/4c5f6f19ebb28235a86c247553b1b20693643b27).

Please do (a hopefully) last review round 🙂.

mattip:
Thanks @fangerer

PR #416, opened 2023-03-13
Introduce HPySlice_Unpack and HPySlice_AdjustIndices.

fangerer:
Two more functions required for NumPy/HPy:

* `HPySlice_Unpack` implemented as full API/ABI function
* `HPySlice_AdjustIndices` implemented as inline helper function. This one would not even need the context since it is actually just doing some arithmetic on the given arguments but for consistency, it still has a `HPyContext *ctx` argument.

This PR also includes some other small fixes:

* Minor fixes for Python 3.12 compatibility
* Fix missing `NULL` terminator in test template
* Fix caption level in contributing docs
* Implement format specifier `'n'` (`HPy_ssize_t`) in `HPy_BuildValue`. Not sure why we didn't have it since we already have it in `HPyArg_Parse`.
* We incorrectly used return type `int` for `ctx_ContextVar_Get` in `ctx_funcs.h` (should be type `int32_t`).

### DIFF
--- a/hpy/devel/include/hpy/inline_helpers.h
+++ b/hpy/devel/include/hpy/inline_helpers.h
@@ -395,4 +395,67 @@ HPyBool_FromLong(HPyContext *ctx, long v)
     return HPyBool_FromBool(ctx, (v ? true : false));
 }
 
+/**
+ * Adjust start/end slice indices assuming a sequence of the specified length.
+ *
+ * Out of bounds indices are clipped in a manner consistent with the handling of
+ * normal slices. This function cannot fail and does not call interpreter
+ * routines.
+ *
+ * :param ctx:
+ *     The execution context.
+ * :param length:
+ *     The length of the sequence that should be assumed for adjusting the
+ *     indices.
+ * :param start:
+ *     Pointer to the start value (must not be ``NULL``).
+ * :param stop:
+ *     Pointer to the stop value (must not be ``NULL``).
+ * :param step:
+ *     The step value of the slice (must not be ``0``)
+ *
+ * :return:
+ *     Return the length of the slice. Always successful. Doesn’t call Python code.
+ */
+HPyAPI_INLINE_HELPER HPy_ssize_t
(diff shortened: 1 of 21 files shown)


hpy/devel/include/hpy/inline_helpers.h
```diff
@@ -395,4 +395,67 @@ HPyBool_FromLong(HPyContext *ctx, long v)
...
+ * :param stop:
+ *     Pointer to the stop value (must not be ``NULL``).
+ * :param step:
+ *     The step value of the slice (must not be ``0``)
+ *
+ * :return:
+ *     Return the length of the slice. Always successful. Doesn’t call Python code.
+ */
+HPyAPI_INLINE_HELPER HPy_ssize_t
+HPySlice_AdjustIndices(HPyContext *ctx, HPy_ssize_t length, HPy_ssize_t *start, HPy_ssize_t *stop, HPy_ssize_t step)
```
mattip, line 421: Do we have an UNUSED macro to mark `ctx`?

hpy/tools/autogen/public_api.h
```diff
@@ -648,6 +648,40 @@ HPy_ID(204)
...
+ * The slice members may be arbitrary int-like objects. If they are not Python
+ * int objects, they will be coerced to int objects by calling their
+ * ``__index__`` method.
+ *
+ * If a slice member value is out of bounds, it will be set to the maximum value
+ * of ``HPy_ssize_t`` if the member was a positive number, or to the minimum
+ * value of ``HPy_ssize_t`` if it was a negative number.
+ *
+ * :param ctx:
+ *     The execution context.
```
mattip, line 666:
```suggestion
 *     The execution context, not used but kept for compatibility with other signatures
```

mattip (approved):
LGTM. Perhaps could emphasize that `ctx` is unused in `HPySlice_AdjustIndices`

hpy/devel/include/hpy/inline_helpers.h
fangerer: we only have
```
#define _HPy_UNUSED __attribute__((unused))
```
in `hpy.h`. I'll introduce one. Since all tests are green, let me do the fixup in #415 (since that needs an update anyway).

PR #417, opened 2023-03-14
GitHub Action to lint Python code

cclauss:
As suggested at https://github.com/hpyproject/hpy/wiki/dev-call-20201105#topics-deferred-to-the-next-dev-call

[Ruff](https://beta.ruff.rs/) supports [over 500 lint rules](https://beta.ruff.rs/docs/rules) including bandit, isort, pylint, pyupgrade, and flake8 plus its plugins, and is written in Rust for speed.

The `ruff` Action uses minimal steps to run in ~10 seconds, rapidly providing intuitive GitHub Annotations to contributors.

image

Python file edits were done with:
% [`ruff --select=E703,E711,E712,F841 --fix .`](https://beta.ruff.rs/docs/rules/#error-e)

### DIFF
--- a/.github/workflows/ruff.yml
+++ b/.github/workflows/ruff.yml
@@ -1,0 +1,16 @@
+# https://beta.ruff.rs
+name: ruff
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+jobs:
+  ruff:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - run: pip install --user ruff
+    - run: ruff --format=github .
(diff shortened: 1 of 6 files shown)


fangerer (approved):
LGTM. Thank you, @cclauss.

fangerer (changes requested):
Sorry, I was a bit too fast in approving. I had some discussion with other HPy devs and they convinced me that can be helpful but we need to carefully select the rules.
Let's wait for some other opinions.

hodgestar:
Given that no big changes were required for the current set of rules to pass, I'm happy with the PR as is. I'm also happy with slightly more relaxed rules though.

mattip:
I think we should reject this PR. The only time code formatting has become a discussion topic is when use of a tool to format code is proposed. This seems to be the flip side of the "code formatters __prevent__ discussions around formatting": to become "code formatters __generate__ unhelpful and long discussions around code formatting".

PR #425, opened 2023-04-18
Put hyperlink to API reference into readme and improve autogen of API mapping table.

### LINKED ISSUE #323: What is the actual API?
If I were to start writing a new module in C, I wouldn't know where to start.
The actual header file seems to be quite deep in the source tree at https://github.com/hpyproject/hpy/blob/master/hpy/devel/include/hpy.h which includes a lot of other files, so I'm already a bit lost.

The docs at https://docs.hpyproject.org/en/latest/# describe the high-level concepts, porting and various other bits and pieces, but still no actual API.

The CPython API, for all its flaws, is more navigable thanks to the online documentation at https://docs.python.org/3/c-api/index.html.

HPy may provide a better API than CPython's, but without documentation, it is a challenge to use.
I wouldn't recommend duplicating all the API in parallel docs, like CPython does, but adding some parseable comments to the source and using a tool like doxygen would produce something usable.

fangerer:
Should finally resolve #323 and also improves the auto-generation of the C API - HPy API function mapping table.
The table is now written into the document that contains it (currently `docs/porting-guide.rst`). The new approach replaces everything between two markers (i.e. `.. mark: BEGIN API MAPPING\n` and `.. mark: END API MAPPING\n`).

Previously, I manually merged an auto-generated part of the table with some manually defined mappings. This is now also done by autogen. Therefore, one needs to specify the manual mappings in dict `DOC_MANUAL_API_MAPPING`(file `hpy/tools/autogen/conf.py`).


docs/api-reference/index.rst
```diff
@@ -3,8 +3,8 @@ API Reference
 
 HPy's public API consists of three parts:
 
-1. The **Core API** as defined in file ``public_api.h``
-2. **HPy Helper** functions 
+1. The **Core API** as defined in file :doc:`public_api`
```
hodgestar, line 6:
```suggestion
1. The **Core API** as defined in the :doc:`public_api`
```

docs/api-reference/public_api.rst
```diff
@@ -0,0 +1,12 @@
+:orphan:
```
hodgestar, line 1: Shouldn't we just add a toctree to `api-reference/index.rst` somewhere rather than marking this as an orphan page?

docs/api-reference/public_api.rst
```diff
@@ -0,0 +1,12 @@
...
+
+Public API Header
+=================
+
+`public_api.h
+<https://github.com/hpyproject/hpy/blob/master/hpy/tools/autogen/public_api.h>`_
+(GitHub)
```
hodgestar, line 8:
```suggestion
The core API is defined in `public_api.h
<https://github.com/hpyproject/hpy/blob/master/hpy/tools/autogen/public_api.h>`_:
```

hodgestar (commented):
I left a couple of small comments, but I think this is a step forward on the documentation front. Thank you!

docs/api-reference/public_api.rst
fangerer: Sure, no objections. I just used the same scheme as for `docs/porting-example/steps/step_*.rst`.

hodgestar: The step files were only removed because I didn't want them to distract people from the text of the porting guide or clutter the side bar menu. I think the core API is worthy of being in a toctree though.

PR #427, opened 2023-04-21
Prepare next release (version 0.9.0).

fangerer:
This PR prepares the next release (version 0.9).
Once this is merged, I'll do a test by creating a release candidate `0.9.0rc1` or something.

### Refactor GitHub action `Release to PyPI`

I've refactor our GitHub action that builds the distributions and uploads them  to PyPI.
Previously, we used the GitHub `Release published` event. However, that did not work the last time. So, I switched to the `push tag` event. This means, whenever we push a tag in form of `*.*.*` (e.g. `0.9.0`) we will build and upload the distributions.

Further, I plan to do the whole action on every PR (maybe just `post-merge`) but only uploading to Test PyPI. This is still disabled because we don't own https://test.pypi.org/project/hpy. I've asked the current owner to transfer ownership but didn't get an answer yet.

I'm now also using a new GitHub action `softprops/action-gh-release@v1` that will also create a release on GitHub.

### Release Notes

I, ofc, wrote release notes going through all PRs we merged since `0.0.4`

### DIFF
--- a/.github/workflows/release-pypi.yml
+++ b/.github/workflows/release-pypi.yml
@@ -1,20 +1,36 @@
 ---
-name: Release to PyPI
+name: Publish to PyPI
 
+# manually trigger this workflow
 on:
-  release:
-    types: [created]
-
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'The Git tag to create the release for'
+        required: true
+        type: string
+      official:
+        description: 'Publish to official PyPI'
+        default: false
+        type: boolean
 
 jobs:
   build_sdist:
     name: Build source package
     runs-on: 'ubuntu-latest'
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
+      with:
+        ref: ${{ inputs.tag }}
+
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        # HPy requires at least Python 3.8; we mostly work with >=3.10
+        python-version: '>=3.10'
 
     - name: Install/Upgrade Python dependencies
-      run: python -m pip install --upgrade pip wheel
+      run: python -m pip install --upgrade pip wheel 'setuptools>=60.2'
 
     - name: Build and install Python source package
       run: |
@@ -23,12 +39,14 @@ jobs:
 
     - name: Run tests
       run: |
-        pip install pytest pytest-xdist
+        make
+        pip install pytest pytest-xdist filelock
         pytest -n auto test/
 
-    - uses: actions/upload-artifact@v2
+    - uses: actions/upload-artifact@v3
       with:
         path: dist/*.tar.gz
+        retention-days: 5
 
 
   build_bdist:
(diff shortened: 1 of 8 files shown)


docs/conf.py
```diff
@@ -26,7 +26,7 @@
 author = "HPy Collective"
 
 # The full version, including alpha/beta/rc tags
-release = "0.0.4"
+release = "0.9"
```
auvipy, line 29: jump in version?

.github/workflows/release-pypi.yml
```diff
@@ -37,46 +47,86 @@ jobs:
...
-        CIBW_TEST_REQUIRES: pytest pytest-xdist
-        CIBW_TEST_COMMAND: pytest -n auto {project}/test/
-
-    - uses: actions/upload-artifact@v2
+        # cibuildwheel automatically reads 'python_requires' from 'setup.py'
+        CIBW_BUILD: 'cp*'  # no PyPy builds
+        CIBW_ARCHS_LINUX: "x86_64" # only Intel 64-bit
+        CIBW_ARCHS_MACOS: "x86_64" # only Intel 64-bit
+        # CIBW_TEST_REQUIRES: pytest pytest-xdist filelock 'setuptools>=60.2'
+        # CIBW_TEST_COMMAND: pytest --basetemp=.tmpdir -n auto {project}/test/
```
hodgestar, line 71: What is the status of these commented out lines?

docs/changelog.rst
```diff
@@ -1,6 +1,61 @@
 Changelog
 =========
 
+Version 0.9 (April 21st, 2023)
+------------------------------
+
+New Features/API
```
hodgestar, line 7: I'd like to see this reorganized a bit to make it much clearer how much work has been done since the last release and of what kind. Concrete proposal:

- Add a very short paragraph at the top that says roughly "this release adds a huge number of major features and the next major release will be 1.0". 
- Start with a section on major new features and add a one sentence explanation of each (and maybe a link to docs if any). Have one item that mentions the documentation improvements as a whole and refers to the full list further down.
- Then have a section on all the new APIs added.
- Then list the debug mode improvements (to remind people debug mode is awesome).
- Only then list the Python version support changes.

hodgestar (changes requested):
Thanks @fangerer.

I left some comments on the changelog. We need to be clearer that a lot of great work has been done (and a lot of it by you!).

I don't think we should upload to test pypi on every push. It's not really intended to be a heavily used service, just a place to try things out and I wouldn't want to have many hundreds of builds there.

Instead, I suggest that we add the ability to manually trigger a publish to either pypi test (default) or pypi. The manual push to test pypi allows us to try out big release process changes (hopefully infrequent). A manual push to the real pypi allows us to redo failed releases easily (hopefully also infrequent).

Auto release on tags is great, so let's keep that.

docs/conf.py
fangerer: Yes. We used version `0.0.x` to indicate that we are in an early (alpha) phase. We now think that we left the _alpha_ phase and want to indicate this by jumping to `0.x`. We actually discussed if this release should be `1.0` but we wanted to leave some _space_.

.github/workflows/release-pypi.yml
fangerer: I plan to re-enable this tests in this PR but ran out of time on Friday. The main problem is: we usually run tests _in-place_ but here, we really want to test the wheel. I'll finish that on Monday.

fangerer:
> I don't think we should upload to test pypi on every push. It's not really intended to be a heavily used service, just a place to try things out and I wouldn't want to have many hundreds of builds there.

You are probably right. I just saw it in some examples and thought it is maybe a good idea because whenever I do a release, I realize that the build-and-upload-wheel machinery is broken for some reason.

> Instead, I suggest that we add the ability to manually trigger a publish to either pypi test (default) or pypi.

Agreed. I suggest that we also do it with tags. We could have a filter and if a tag is in form, let's say, `test-*.*.*`, we upload to Test PyPI. I'm not sure if we should also create a GitHub draft release in this case. What do you think?

Instead of tags, we could also use branches. E.g all branches that start with `test-release/` will upload to Test PyPI.

.github/workflows/release-pypi.yml
```diff
@@ -37,46 +45,88 @@ jobs:
...
-        CIBW_TEST_COMMAND: pytest -n auto {project}/test/
-
-    - uses: actions/upload-artifact@v2
+        # cibuildwheel automatically reads 'python_requires' from 'setup.py'
+        CIBW_BUILD: 'cp*'  # no PyPy builds
+        CIBW_ARCHS_LINUX: "x86_64" # only Intel 64-bit
+        CIBW_ARCHS_MACOS: "x86_64" # only Intel 64-bit
+        CIBW_TEST_REQUIRES: pytest pytest-xdist filelock setuptools>=60.2
+        # build static libs; those are required for testing
+        CIBW_BEFORE_TEST: python {project}/setup.py build_clib build_ext -i
```
fangerer, line 70: I'm a bit unhappy about that and plan to fix that for the next minor release.
`cibuildwheel` is already very clever and changes the working directory to something else than the project root when running the tests. However, it seems that `pytest` discovers the project root and then also prepends it to `sys.path`. As far as I could find out locally, this will still run the tests _from source_ (so, not using the installed HPy package). 

I tried following things to mitigate that:
* Set `--import-mode=append`; fails because with:
  ```text
  ImportError while loading conftest '{project}/test/conftest.py'.
  ModuleNotFoundError: No module named 'test.conftest'
  ```
* Set `--import-mode=importlib`; not possible for us because our test modules have relative imports (as far as I understood from docs)
* Set `--root-dir` to something else; failed

The only thing that worked as expected for me (locally) was to copy `{project}/test` to some temp dir and use that. Maybe that's what we should do here as well but I first need to verify that we can do that.

Do you know what is the proper way to do this with `pytest`?

hodgestar:
@fangerer I also don't want lots of `test-....` tags. I was thinking we could use the GitHub Actions `workflow_dispatch` option to run test builds from the GitHub Actions UI.

.github/workflows/release-pypi.yml
hodgestar: We definitely shouldn't run `setup.py build_XXX` when testing the wheel. It is already built and installed. Perhaps just copy the test folder across to wherever the tests are being run?

fangerer:
Btw. the previous owner of https://test.pypi.org/project/hpy deleted the project but I still cannot upload to it:
```
403 The user 'fangerer' isn't allowed to upload to project 'hpy'. See  
         https://test.pypi.org/help/#project-name for more information.  
```
Any ideas why this is the case?

hodgestar:
@fangerer The docs seem to indicate that `uses` will run the called workflow again, thus building different wheels for GitHub and PyPI releases, which is not what we want I think.

hodgestar:
> I've also fixed the wheel testing. In order to do that, we will now ship the static libraries since they are strictly required by the test suite.

@fangerer I couldn't figure out what if the diff implements this change. Would you mind explaining a bit more?

fangerer:
@hodgestar btw. here is a run where you can see that tests were actually executed with `cibuildhweel`: https://github.com/hpyproject/hpy/actions/runs/4796648081/jobs/8532661365

GalaxySnail:
> The docs seem to indicate that `uses` will run the called workflow again, thus building different wheels for GitHub and PyPI releases, which is not what we want I think.

What about [reproducible builds](https://reproducible-builds.org/)? I'm not sure if it is too difficult or takes a lot of effort to do. But it will be great if we can achieve reproducible builds.

.github/workflows/release-pypi.yml
```diff
@@ -1,20 +1,36 @@
...
-    types: [created]
-
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'The Git tag to create the release for'
+        required: true
+        type: string
+      official:
+        description: 'Publish to official PyPI'
```
hodgestar, line 13:
```suggestion
        description: 'If true, publish to official PyPI and create GitHub release. Otherwise publish only to test PyPI.'
```

.github/workflows/release-pypi.yml
```diff
@@ -1,20 +1,36 @@
...
 
+# manually trigger this workflow
 on:
-  release:
-    types: [created]
-
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'The Git tag to create the release for'
```
hodgestar, line 9:
```suggestion
        description: 'The Git tag or branch to create or test a release for. Official releases should always be made from an existing tag.'
```

fangerer: This is not exactly true. I'm using action `softprops/action-gh-release` and the `tag` input will be passed to this action to the `tag_name` param. If the specified tag does not exist, it will create the tag from the current revision. I wanted to keep it simple but we can also have an additional input `ref` that (if set) will be the ref (i.e. SHA/branch/etc) to be used to create a new tag (if the tag does not yet exist).

fangerer: I just dropped the `or branch` for now.

hodgestar (approved):
It would be nice to be able to build from a branch for testing purposes, but +1 this for now so that it can be merged. We can improve things later as needed.

fangerer:
> It would be nice to be able to build from a branch for testing purposes, but +1 this for now so that it can be merged. We can improve things later as needed.

Agreed. Let's get the release out and I promise to add this ability. We can discuss that in the dev call.

PR #432, opened 2023-05-11
Add structseq.c to extra sources.

fangerer:
This is an important bug fix for HPy `0.9.0rc1` (and I will create `rc2` after this is merged).

When I've added the helper functions `HPyStructSequence_New/NewType` in PR #415 , I obviously forgot to add it to the list of extra sources in `hpy.devel`.  This may result in an undefined symbol error when linking.

It doesn't show up in the tests because we build a static lib containing the helpers and I did add the new source file to there (in `setup.py`).

In order to avoid such problems in future, I plan to have the list of helper sources just in one place.

### DIFF
--- a/hpy/devel/__init__.py
+++ b/hpy/devel/__init__.py
@@ -74,6 +74,7 @@ class HPyDevel:
             self.src_dir.joinpath('buildvalue.c'),
             self.src_dir.joinpath('format.c'),
             self.src_dir.joinpath('helpers.c'),
+            self.src_dir.joinpath('structseq.c'),
         ]))
 
     def _scan_static_lib_dir(self):
--- a/hpy/devel/__init__.py
+++ b/hpy/devel/__init__.py
@@ -188,8 +188,11 @@ def handle_hpy_ext_modules(dist, attr, hpy_ext_modules):
     dist.__class__.hpy_abi = DEFAULT_HPY_ABI
     dist.__class__.hpy_use_static_libs = False
     dist.__class__.global_options += [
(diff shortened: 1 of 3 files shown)


hpy/devel/__init__.py
```diff
@@ -187,8 +188,11 @@ def handle_hpy_ext_modules(dist, attr, hpy_ext_modules):
     dist.__class__.hpy_abi = DEFAULT_HPY_ABI
     dist.__class__.hpy_use_static_libs = False
     dist.__class__.global_options += [
-        ('hpy-abi=', None, 'Specify the HPy ABI mode (default: %s)' % DEFAULT_HPY_ABI),
-        ('hpy-no-static-libs', None, 'Compile context and extra sources with extension (default: False)')
+        ('hpy-abi=', None, 'Specify the HPy ABI mode (default: %s)'
+                           % DEFAULT_HPY_ABI),
+        ('hpy-use-static-libs', None, 'Use static library containing context '
+                                      'and helper functions for building '
+                                      'extensions (default: False)')
```
mattip, line 195: Is this a typo fix?

mattip:
LGTM, just trying to understand the `hpy-no-static-libs` change.

hpy/devel/__init__.py
fangerer: Yes, I first called it `hpy-no-static-libs` but then we changed it to semantically _do not use static libs by default_. So, I renamed the attribute but didn't rename the option 😞 .
This just shows that everything that is not tested will eventually fail. I will also add some tests for this as soon as I find the time to do so.

PR #440, opened 2023-06-20
Make autogen work under Windows+MSVC

KubaO:
1. Default encoding on Windows isn't UTF-8. This caused autogen to mangle the documentation files.
2. Add detection of MSVC so that autogen will work under msvc command line.

This helps when using hpy under MSVC.

### DIFF
--- a/hpy/tools/autogen/parse.py
+++ b/hpy/tools/autogen/parse.py
@@ -4,6 +4,7 @@ import attr
 import re
 import py
 import pycparser
+import shutil
 from pycparser import c_ast
 from pycparser.c_generator import CGenerator
 from distutils.sysconfig import get_config_var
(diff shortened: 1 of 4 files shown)


hpy/tools/autogen/parse.py
```diff
@@ -202,16 +203,26 @@ class HPyAPI:
                             re.DOTALL | re.MULTILINE)
 
     def __init__(self, filename):
-        cpp_cmd = get_config_var('CC').split(' ')
+        cpp_cmd = get_config_var('CC')
+        if cpp_cmd:
+            cpp_cmd = cpp_cmd.split(' ')
+        elif sys.platform == 'win32':
```
fangerer, line 209: Interesting. Is conf variable `CC`not set on Windows? I did it this way assuming that we would get `cl.exe` from `CC. I thought I've tested that on Windows but since it's not tested in the CI, I can imagine that this changed or is different on other Windows versions.

hpy/tools/autogen/parse.py
```diff
@@ -202,16 +203,26 @@ class HPyAPI:
...
+            cpp_cmd = [shutil.which("cl.exe")]
         if sys.platform == 'win32':
             cpp_cmd += ['/E', '/I%s' % CURRENT_DIR]
         else:
             cpp_cmd += ['-E', '-I%s' % CURRENT_DIR]
 
+        msvc = "cl.exe" in cpp_cmd[0].casefold()
```
fangerer, line 216: I don't see any usage of variable `msvc`.

hpy/tools/autogen/parse.py
KubaO: CC is empty on Windows - perhaps because finding MSVC has been a pain nobody bothered with? Even in distutils it was a bit of an unreliable mess.

fangerer:
> We don't care about Python 2.7 anymore, right?

Hi @KubaO, please excuse my very late answer. We don't really _care_ about Python 2.7 in the sense that we don't run or support HPy on Python 2.7. However, we are running tests on PyPy and some of the test files therefore also need to be Python 2.7 compatible. Anyway, @mattip already has a fix (see #444 ).

mattip:
Close/reopen to trigger CI

PR #444, opened 2023-07-22
fix typo: HPy_mod_execute -> HPy_mod_exec

### LINKED ISSUE #443: Add an example or test that uses HPy_mod_execute
In porting hpy-0.9 to PyPy, I could [not find](https://github.com/search?q=org%3Ahpyproject%20HPy_mod_execute&type=code) an example of using the `HPy_mod_execute` slot. It is mentioned in the [documentation](https://docs.hpyproject.org/en/latest/quickstart.html?highlight=HPy_mod_execute#hpy-quickstart)
> The Python interpreter will create the module for us from the `HPyModuleDef` specification. Additional initialization can be done in the `HPy_mod_execute` slot

but there is no example of using that slot.

mattip:
Closes #443 

HPy_mod_execute in documentation should be HPy_mod_exec

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -322,10 +322,16 @@ jobs:
     #   parameters:
     #     pythonVersion: "2.7"
 
-      - name: Set up Python
-        uses: actions/setup-python@v4
-        with:
-          python-version: '2.7'
+      - name: Set up Python2
+        # Copied from cython's ci.yml
+        run: |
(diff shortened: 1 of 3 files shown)


mattip:
The Python2.7 test is failing since python2.7 was removed from the `actions/setup-python@v4` github action. How can we still install it to run the tests?

mattip:
CI is passing, if someone wants to take a look

fangerer (approved):

PR #449, opened 2023-09-15
Final changes for stable release 0.9.

### LINKED ISSUE #448: debug ctx_Type_GetName does not take trailing NULL into account
In this call the trailing NULL is not copied back out of `protect_and_associate_data_ptr`, so I am getting garbage on the end of the returned `char*`. I think `n_name` should be incremented to include the trailing NULL.  https://github.com/hpyproject/hpy/blob/8310a762d78e3412464b1869959a77da013e6307/hpy/debug/src/debug_ctx.c#L559-L564

fangerer:
Fixes #448 and prepares for Python 3.12 as well.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,8 @@ jobs:
             compiler: 'g++'
           - os: ubuntu-latest
             python-version: '3.11'
+          - os: ubuntu-latest
+            python-version: '3.12'
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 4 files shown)


fangerer:
@mattip Turns out that there was a small issue with Python 3.12. I've added 3.12 to the test matrix (but not yet to `cibuildwheel`).

hodgestar (approved):
Looks good so far to me.

mattip:
Thanks @fangerer

Issue #452, opened 2023-10-19
Duplicate supports_ref_counts


mattip:
A second implementation of the `supports_ref_counts` function in `support.py` was added in PR #390, I think this was supposed to be named something related to vectorcall?
https://github.com/hpyproject/hpy/blob/f6114734b4a5d03cfb9d24d3e8c71f77e8803881/test/support.py#L498-L505

@fangerer

fangerer:
True, that was a mistake. I'll fix it.

mattip:
Closed by #455

PR #455, opened 2023-11-03
backport changes from PyPy

mattip:
After merging the hpy-0.9 branch in PyPy, these were the relevant changes I made.

### DIFF
--- a/hpy/debug/src/_debugmod.c
+++ b/hpy/debug/src/_debugmod.c
@@ -10,7 +10,7 @@
 static UHPy new_DebugHandleObj(HPyContext *uctx, UHPy u_DebugHandleType,
                                DebugHandle *handle);
 
-HPY_MOD_EMBEDDABLE(_trace)
+HPY_MOD_EMBEDDABLE(_debug)
 
 HPyDef_METH(new_generation, "new_generation", HPyFunc_NOARGS)
 static UHPy new_generation_impl(HPyContext *uctx, UHPy self)
--- a/hpy/devel/include/hpy/hpytype.h
+++ b/hpy/devel/include/hpy/hpytype.h
@@ -203,14 +203,15 @@ typedef struct {
 
 /** Convenience macro which is equivalent to:
     ``HPyType_HELPERS(TYPE, HPyType_BuiltinShape_Legacy)`` 
-    For instance, HPyType_LEGACY_HELPERS(DummyMeta) will produce
+    For instance, HPyType_LEGACY_HELPERS(DummyMeta) will produce::
 
-    enum { DummyMeta_SHAPE = (int)HPyType_BuiltinShape_Legacy }; 
-    __attribute__((unused)) static inline 
-    DummyMeta * 
-    DummyMeta_AsStruct(HPyContext *ctx, HPy h) { 
-        return (DummyMeta *) _HPy_AsStruct_Legacy(ctx, h); 
-    } */
+        enum { DummyMeta_SHAPE = (int)HPyType_BuiltinShape_Legacy }; 
+        __attribute__((unused)) static inline 
+        DummyMeta * 
+        DummyMeta_AsStruct(HPyContext *ctx, HPy h) { 
+            return (DummyMeta *) _HPy_AsStruct_Legacy(ctx, h); 
+        }
+*/
 #define HPyType_LEGACY_HELPERS(TYPE) \
     HPyType_HELPERS(TYPE, HPyType_BuiltinShape_Legacy)
 
--- a/hpy/devel/include/hpy/hpytype.h
+++ b/hpy/devel/include/hpy/hpytype.h
@@ -202,7 +202,15 @@ typedef struct {
 #define HPy_TPFLAGS_HAVE_GC (1UL << 14)
 
 /** Convenience macro which is equivalent to:
-    ``HPyType_HELPERS(TYPE, HPyType_BuiltinShape_Legacy)`` */
+    ``HPyType_HELPERS(TYPE, HPyType_BuiltinShape_Legacy)`` 
+    For instance, HPyType_LEGACY_HELPERS(DummyMeta) will produce
+
+    enum { DummyMeta_SHAPE = (int)HPyType_BuiltinShape_Legacy }; 
+    __attribute__((unused)) static inline 
+    DummyMeta * 
+    DummyMeta_AsStruct(HPyContext *ctx, HPy h) { 
+        return (DummyMeta *) _HPy_AsStruct_Legacy(ctx, h); 
+    } */
 #define HPyType_LEGACY_HELPERS(TYPE) \
     HPyType_HELPERS(TYPE, HPyType_BuiltinShape_Legacy)
 
--- a/test/debug/test_charptr.py
+++ b/test/debug/test_charptr.py
@@ -1,6 +1,6 @@
 import os
 import pytest
-from test.support import SUPPORTS_SYS_EXECUTABLE, SUPPORTS_MEM_PROTECTION
+from ..support import SUPPORTS_SYS_EXECUTABLE, SUPPORTS_MEM_PROTECTION
 
 # Tests detection of usage of char pointers associated with invalid already
 # closed handles. For now, the debug mode does not provide any hook for this
--- a/test/support.py
+++ b/test/support.py
@@ -50,15 +50,23 @@ def make_hpy_abi_fixture(ABIs, class_fixture=False):
         hpy_abi = make_hpy_abi_fixture('with hybrid', class_fixture=True)
     """
     if ABIs == 'default':
-        if sys.implementation.name == 'cpython':
-            ABIs = ['cpython', 'universal', 'debug']
-        else:
-            ABIs = ['universal', 'debug']
-    elif ABIs == 'with hybrid':
-        if sys.implementation.name == 'cpython':
+        try:
+            if sys.implementation.name == 'cpython':
+                ABIs = ['cpython', 'universal', 'debug']
+            else:
+                ABIs = ['universal', 'debug']
+        except:
+            # CPython2.7: no sys.implementation
             ABIs = ['cpython', 'hybrid', 'hybrid+debug']
-        else:
-            ABIs = ['hybrid', 'hybrid+debug']
+    elif ABIs == 'with hybrid':
+        try:
+            if sys.implementation.name == 'cpython':
+                ABIs = ['cpython', 'hybrid', 'hybrid+debug']
+            else:
+                ABIs = ['hybrid', 'hybrid+debug']
+        except:
+            # CPython2.7: no sys.implementation
+            ABIs = ['cpython', 'hybrid', 'hybrid+debug']
     elif isinstance(ABIs, list):
         pass
     else:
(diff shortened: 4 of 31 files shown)


hpy/devel/include/hpy/hpytype.h
```diff
@@ -202,7 +202,15 @@ typedef struct {
...
-    ``HPyType_HELPERS(TYPE, HPyType_BuiltinShape_Legacy)`` */
+    ``HPyType_HELPERS(TYPE, HPyType_BuiltinShape_Legacy)`` 
+    For instance, HPyType_LEGACY_HELPERS(DummyMeta) will produce
+
+    enum { DummyMeta_SHAPE = (int)HPyType_BuiltinShape_Legacy }; 
+    __attribute__((unused)) static inline 
+    DummyMeta * 
+    DummyMeta_AsStruct(HPyContext *ctx, HPy h) { 
+        return (DummyMeta *) _HPy_AsStruct_Legacy(ctx, h); 
+    } */
```
mattip, line 213: I found it helpful to decode what the macro actually produces

test/support.py
```diff
@@ -495,15 +501,6 @@ def supports_ordinary_make_module_imports(self):
...
 
-    def supports_refcounts(self):
-        """ Returns True if the underlying Python implementation supports
-            the vectorcall protocol.
-
-            By default, this returns True for Python version 3.8+ on all
-            implementations.
-        """
-        return sys.version_info >= (3, 8)
-
```
mattip, line 506: duplicate code

test/support.py
```diff
@@ -572,7 +569,7 @@ def _build(tmpdir, ext, hpy_devel, hpy_abi, compiler_verbose=0, debug=None):
     dist.hpy_abi = hpy_abi
     # For testing, we want to use static libs to avoid repeated compilation
     # of the same sources which slows down testing.
-    dist.hpy_use_static_libs = True
+    dist.hpy_use_static_libs = False
```
mattip, line 572:
```suggestion
    dist.hpy_use_static_libs = True
```

test/test_argparse.py
```diff
@@ -75,7 +75,7 @@ def test_s(self):
             "function a str is required"
         )
 
-    def test_B(self):
+    def test_upper_b(self):
```
mattip, line 78: PyPy uses an older version of pytest that does not support upper case test names, so only one of `test_b`, `test_B` ran

test/test_hpyerr.py
```diff
@@ -461,14 +461,20 @@ def test_HPyErr_WarnEx(self):
...
         # NOTE: trampoline is defined in support.py
+        import sys
+        ISPYPY = "__pypy__" in sys.modules
         def check_warning(arg, category, message, file):
             with warnings.catch_warnings(record=True) as warnings_list:
-                trampoline(mod.f, arg)
+                if ISPYPY:
+                    mod.f(arg)
+                else:
+                    trampoline(mod.f, arg)
```
mattip, line 471: Did the original author mean to call the legacy function here?

test/test_hpyfield.py
```diff
@@ -326,6 +326,7 @@ def test_tp_finalize(self):
             // nicely with pytest, but if the finalizer gets called after the
             // test has finished, we have no choice but to abort to signal
             // the error
+            static void on_unexpected_finalize_call(void);
```
mattip, line 329: removes a warning about missing declarations

test/test_hpytype_legacy.py
```diff
@@ -257,6 +257,56 @@ def test_call_zero_basicsize(self):
...
+            @INIT
+        """)
+        # Calls __class_getitem__
+        t = mod.Point[int]
+        assert str(t) == "mytest.Point[int]"
+        pt = mod.Point()
+        assert pt.x == 0
+        assert pt.y == 0
+
+
```
mattip, line 309: added a test that `__class_getitem__` is handled properly

test/support.py
```diff
@@ -50,9 +50,23 @@ class TestFoo(HPyTest):
...
     if ABIs == 'default':
-        ABIs = ['cpython', 'universal', 'debug']
+        try:
+            if sys.implementation.name == 'cpython':
+                ABIs = ['cpython', 'universal', 'debug']
+            else:
+                ABIs = ['universal', 'debug']
+        except:
+            # CPython2.7: no sys.implementation
+            ABIs = ['cpython', 'hybrid', 'hybrid+debug']
```
hodgestar, line 60:
```suggestion
            ABIs = ['cpython', 'universal', 'debug']
```
The list here was copy and pasted incorrectly.

We should simplify this somehow. Perhaps there can just be a pytest flag that adds `cpython` to the list? Or make a flag here like `is_cpython = not hasattr(sys, "implementation") or sys.implementation.name == 'cpython'` so at least we're not repeating strange logic.

test/test_argparse.py
hodgestar: A bit strange, but okay. :)

test/test_hpyerr.py
```diff
@@ -461,14 +461,20 @@ def test_HPyErr_WarnEx(self):
...
                 assert len(warnings_list) == 1, str(category)
                 w = warnings_list[-1]
                 assert issubclass(w.category, category), str(category)
                 assert str(w.message) == message, str(category)
-                assert w.filename.endswith(file), str(category)
+                if not ISPYPY:
+                    assert w.filename.endswith(file), str(category)
```
hodgestar, line 477: Why are the filenames different on PyPy?

test/test_hpyerr.py
hodgestar: I have no idea why `trampoline` was invoked here. Perhaps we can ask in the HPy Discord.

test/test_slots.py
```diff
@@ -419,6 +419,10 @@ def test_buffer(self):
                 assert sys.getrefcount(arr) == init_refcount + 1
             for i in range(12):
                 assert mv[i] == i
+        del mv
+        import gc
+        for i in range(3):
+            gc.collect()
```
hodgestar, line 425: Why do we need to do `del mv`? I would have thought that the memoryview would cease to link to `arr` once the with block exits? In fact it must so that `mv2` can be created lower down, no?

It's also a bit unclear to me why the `gc.collect()` is needed here? Isn't `self.supports_refcounts()` sufficient?

test/test_hpyerr.py
mattip: This is a quirk in the untranslated tests. Removing the change.

test/test_slots.py
mattip: We need the gc to collect `mv` since on PyPy the buffer is only released when the gc calls the finalizer.

mattip:
I updated the PR after the review.

mattip:
CI is passing after review fixes

hpy/debug/src/_debugmod.c
```diff
@@ -10,7 +10,7 @@
 static UHPy new_DebugHandleObj(HPyContext *uctx, UHPy u_DebugHandleType,
                                DebugHandle *handle);
 
-HPY_MOD_EMBEDDABLE(_trace)
+HPY_MOD_EMBEDDABLE(_debug)
```
fangerer, line 13: oh, thank you

test/debug/test_charptr.py
```diff
@@ -1,6 +1,6 @@
 import os
 import pytest
-from test.support import SUPPORTS_SYS_EXECUTABLE, SUPPORTS_MEM_PROTECTION
+from ..support import SUPPORTS_SYS_EXECUTABLE, SUPPORTS_MEM_PROTECTION
```
fangerer, line 3: I was already wondering because in GraalPy we have a stdlib module `test` (afaik, imported from CPython) and I needed to rename HPy tests in GraalPy. The relative import does unfortunately not work for us because we have the tests in a separate tree.

test/test_hpytype.py
```diff
@@ -1102,10 +1102,12 @@ def test_call_invalid_specs(self):
...
             @INIT
         """)
-        with pytest.raises(TypeError):
+        with pytest.raises(TypeError) as err:
             mod.create_var_type()
-        with pytest.raises(TypeError):
+        assert "Cannot use HPy call protocol with var" in str(err.value)
+        with pytest.raises(TypeError) as err:
             mod.create_call_and_vectorcalloffset_type()
+        # assert "Cannot have HPy_tp_call and explicit member" in str(err.value)
```
fangerer, line 1110: left-over?

test/test_hpytype.py
```diff
@@ -1508,7 +1510,7 @@ def test_get_name(self):
         assert mod.Dummy.__name__ == "Dummy"
         assert mod.get_name(mod.Dummy) == "Dummy"
         assert mod.get_name(str) == "str"
-        assert mod.get_name(array.array) == "array"
+        assert mod.get_name(struct.Struct) == "Struct"
```
fangerer, line 1513: is `array.array` problematic for PyPy?

fangerer (approved):
Thanks. Sorry, we didn't pay enough attention to PyPy's restrictions (in particular that imports should be in the test function).

test/debug/test_charptr.py
mattip: PyPy has this layout, so the relative imports work
```
- tests
  - conftest.py  # specific for PyPY
  - support.py   # specific for PYPy
  |- pypy_tests  # specific for PyPy
  |- _vendored
   - conftest.py  # copied from hpy
   - support.py   # copied from hpy
   - test_*.py    # copied from hpy
   - | debug      # copied from hpy
   - | trace      # copied from hpy
```

test/test_hpytype.py
mattip: There is a difference in the parsing code between PyPy and CPython, so the error is raised but differently. I am leaving this here so someone who cares deeply can try to fix it all to be consistent. I don't remember off the top of my head what the different messages are.

test/test_hpytype.py
mattip: Not so much problematic as awkward. Using it means we need to add another module to the untranslated tests, which slows the testing down ever so slightly (PyPy's untranslated tests build "just enough" of the interpreter to run tests, with a limited set of modules). `struct` is part of the core modules we require always.

test/test_hpyerr.py
steve-s: I think we need to call `trampoline`, such that the `support.py` file ends up on the call-stack when the warning is signaled. This helper method is [called like this](https://github.com/hpyproject/hpy/pull/455/files/f90e45bd2b65a8aac2e646b84a00e0e1f8112459#diff-193925ea633541353994533e8c9fa601c71e47456b6210a3b451aa5ec9622f75R479):
```
check_warning(0, RuntimeWarning, "warn qzp", "support.py")
```
and the assert that you disabled for PyPy should actually check that "support.py" in this case is as a filename in the warning object:
```
assert w.filename.endswith(file) # where file will be `support.py` for us
```

test/test_slots.py
hodgestar: But shouldn't the context manager for `mv` release the buffer when the context is left independent of the GC? Isn't that the point of having a context manager for memoryviews? Or am I not understanding something?

mattip: Maybe this is a bug in PyPy. I will back it out, in the interest of finishing this Pr, and we can visit it again in the future.

test/support.py
mattip: Closes #452

test/test_slots.py
mattip: I found the problem in PyPy - indeed it was a bug

mattip:
CI is passing after review fixes

PR #461, opened 2023-12-04
add a failing test to replicate numpy scalar type scheme

mattip:
This adds a test to try to recreate the NumPy DType type hierarchy. It takes the type spec from [scalartypes.c.src](https://github.com/hpyproject/numpy-hpy/blob/23f4a1898d37af3dd3525144eb800c64abfdb495/numpy/core/src/multiarray/scalartypes.c.src#L82) and the double-inheritance from [multiarraymodule.c](https://github.com/hpyproject/numpy-hpy/blob/23f4a1898d37af3dd3525144eb800c64abfdb495/numpy/core/src/multiarray/multiarraymodule.c#L6153) to create a `float64` dtype. It is locally failing to create the correct tpye heirarchy, which on NumPy is 
```
[<class 'numpy.float64'>, <class 'numpy.floating'>, <class 'numpy.inexact'>,
 <class 'numpy.number'>, <class 'numpy.generic'>, <class 'float'>, <class 'object'>]
```

but in this test is only `[<class 'float'>, <class 'object'>]`. I am not sure why that is.

It then has two methods for creating a new `float64` object: via calling, i.e. `float64(n)`, and via using the `tp_alloc` slot from the `PyTypeObject`, like in `PyArray_Scalar`. This currently fails on PyPy, for a myriad of reasons, which is why I wanted a stand alone test to replicate the type creation.

Any thoughts are welcome.

### DIFF
--- a/test/test_hpytype_legacy.py
+++ b/test/test_hpytype_legacy.py
@@ -258,7 +258,7 @@ class TestLegacyType(_TestType):
                 @INIT
             """)
 
-    @pytest.mark.skipif(sys.version_info < (3, 9), reason="not for python<3.9") 
+    @pytest.mark.skipif(sys.version_info < (3, 9), reason="not for python<3.9")
     def test_legacy_class_method(self):
         mod = self.make_module("""
             @DEFINE_PointObject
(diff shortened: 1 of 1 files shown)


test/test_hpytype_legacy.py
```diff
@@ -397,3 +397,165 @@ def test_legacy_inherits_from_pure_raises(self):
...
+
+            HPyDef_METH(new_float64, "new_float64", HPyFunc_O)
+            static HPy new_float64_impl(HPyContext *ctx, HPy self, HPy arg)
+            {
+                double a = HPyFloat_AsDouble(ctx, arg);
+                HPy h_PyType = HPyGlobal_Load(ctx, float64global);
+                if (HPy_IsNull(h_PyType))
+                    return HPy_NULL;
+                PyTypeObject *pytype = (PyTypeObject *)HPy_AsPyObject(ctx, h_PyType);
+                PyObject *val=pytype->tp_alloc(pytype, 0);
```
hodgestar, line 478: What happens if we use `tp_new` here instead of `tp_alloc`?

mattip:
Closing, the problem was elsewhere

PR #466, opened 2023-12-26
Add a GitHub Action to run codespell and ruff

cclauss:
This GitHub Action uses minimal steps to run in ~5 seconds to rapidly:
1. look for typos in the codebase using codespell, and
2. lint Python code using ruff and provide intuitive GitHub Annotations to contributors.

---

* https://github.com/codespell-project/codespell#readme
* https://docs.astral.sh/ruff


fangerer (approved):
I'm fine with that but would like to wait for a second opinion.
@mattip @hodgestar @antocuni What do you think?

steve-s:
I like that the codespell caught real errors. I don't have a strong opinion on ruff, but there was [already a declined PR](https://github.com/hpyproject/hpy/pull/417) to enable it. Has anything changed since then? Does this PR enable it in some other way than the previous?

cclauss:
I would say that a ton has changed given the rapid adoption of ruff.  It now has more than 700 lint rules and a code formatter that is often preferred over black.

thedrow (approved):

mattip:
As I commented in the other PR I think we are not in need of ruff right now. I said there: 
> I think we should reject this PR. The only time code formatting has become a discussion topic is when use of a tool to format code is proposed. This seems to be the flip side of the "code formatters prevent discussions around formatting": to become "code formatters generate unhelpful and long discussions around code formatting".

mattip:
PRs that contribute to the adoption and development of HPy would be much more helpful.

cclauss:
Linting with ruff has been removed.

PR #467, opened 2023-12-27
ci.yml: macos-latest

cclauss:


fangerer (changes requested):
Thanks for the effort but I think this should not be merged as it is.
Using `macos-latest` is, ofc, fine and we should do that.
However, the current solution just passes tests because of the `xfail` annotations. This means, we are no longer running the marked tests because they all fail on `macos-12` and if they break on `macos-11`, we won't notice.

Until I find the time to investigate and properly fix that, I would be fine if we add a CI job that runs just those tests on `macos-11`.

cclauss:
All of this is a bit strange on GitHub Actions because the current macOS is v14.2 but given that GHA runs on Intel Macs, the highest version we can test on is 12.

cclauss:
Did [Add main_tests os: macos-11 on python-version: 3.12](https://github.com/hpyproject/hpy/pull/467/commits/0e8839830cd935ebbe3aede53f3bd410db1a6d97) do what you were requesting or did I misunderstand?

cclauss:
`macos-latest` is converting to M1…

https://github.blog/changelog/2024-04-01-macos-14-sonoma-is-generally-available-and-the-latest-macos-runner-image

PR #468, opened 2024-01-25
Enable slot HPy_tp_descr_get.

fangerer:
This slot is needed for Cython/HPy because Cython has its own function type named `cython_function_or_method`.
In this PR, I also refactor autogen a bit such that it is a proper package that can be imported. We will use that in GraalPy to inject our own source code generators.

### DIFF
--- a/docs/api-reference/builder.rst
+++ b/docs/api-reference/builder.rst
@@ -1,8 +1,0 @@
-Building tuples and lists
-=========================
-
-.. autocmodule:: autogen/public_api.h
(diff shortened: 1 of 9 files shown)


hodgestar (approved):
The HPy_tp_descr_get implementation and autogen refactoring look good.

@mattip Pinging you so that you know about the autogen changes in case they surprise you when you update the vendored copy in PyPy (assuming you still have one).

mattip:
Thanks for the heads-up.

PR #469, opened 2024-01-25
Add HPyList_Insert

fangerer:
Also an API we needed in Cython/HPy for performance because calling the method `list.insert` was just too slow.


hodgestar (approved):
Looks good.

PR #470, opened 2024-01-26
Add HPy_(Get|Set|Del)Slice

fangerer:
Again, we needed this for Cython/HPy (in particular, for the `fannkuch` benchmark) because getting and setting slices required multiple upcalls like this:
```c
// create a slice object
HPy start = HPyLong_FromLong(lstart);
HPy stop = HPyLong_FromLong(lsop);
HPy ctor_args[] = { start, stop };
HPy slice = HPy_Call(ctx, ctx->h_SliceType, ctor_args, 2, HPy_NULL);
HPy res = HPy_GetItem(ctx, seq, slice);
HPy_Close(ctx, slice);
HPy_Close(ctx, stop);
HPy_Close(ctx, start);
```
Now, this can be done with:
```c
HPy res = HPy_GetSlice(ctx, seq, lstart, lstop);
```

The proposed APIs are probably worth some discussion. E.g. we may think about introducing a third argument to each denoting the slice's `step`. Right now, this will always be `1`.


hodgestar (approved):
I'm happy with the slice methods not supporting a step size here unless someone else sees a strong use case (e.g. striding numpy arrays).

If needed, we can always add our own `HPy_GetSliceStride`, etc later?

PR #471, opened 2024-01-29
Improve implementation of HPy_(Type | Is | GetItem_i) in CPython ABI mode.

fangerer:
Some time ago, I've mentioned that the call overhead in the implementation of `HPy_Type` was crucial (in CPython ABI mode) for the NumPy/HPy demo. The same applies to `HPy_Is`. Therefore, I'm essentially _inlining_ the implementation into the appropriate API functions.
I also did a little improvement in `HPy_GetItem_i` to avoid unnecessary allocation of long objects.


hodgestar (approved):
Looks good to me.

PR #481, opened 2024-06-06
Added HPySlice_New

DuToitSpies:
Added HPySlice_New - still need to add tests. Needed for benchmarks for Cython.

### DIFF
--- a/hpy/tools/autogen/public_api.h
+++ b/hpy/tools/autogen/public_api.h
@@ -963,6 +963,15 @@ HPy HPyTuple_FromArray(HPyContext *ctx, const HPy items[], HPy_ssize_t n);
  *
  * :param ctx:
  *     The execution context.
+ * 
+ * :param start:
+ *     A pointer to a variable where to write the unpacked slice start. Must not
+ *     be ``NULL``.
+ * :param end:
+ *     A pointer to a variable where to write the unpacked slice end. Must not
+ * :param step:
+ *     A pointer to a variable where to write the unpacked slice step. Must not
+ *     be ``NULL``.
  *
  * :returns:
  *     A handle to the new and empty Python slice object or ``HPy_NULL`` in case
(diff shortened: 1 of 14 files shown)


hpy/tools/autogen/public_api.h
```diff
@@ -958,6 +958,19 @@ HPy HPyTuple_FromArray(HPyContext *ctx, const HPy items[], HPy_ssize_t n);
 
 /* sliceobject.h */
 
+/**
+ * Creates a new empty Python slice object.
+ *
+ * :param ctx:
+ *     The execution context.
+ *
```
fangerer, line 966: Please add `:param` for `start`, `stop`, and `step` as well.

fangerer (commented):
LGTM so far. As you mentioned, please add tests. This should be very easy since this function only fails if the object cannot be allocated and we don't test for out of memory problems.
You should just test if `HPy_NULL` is correctly handled if passed as `start`, `stop`, or `step`.

test/test_hpyslice.py
```diff
@@ -108,3 +108,32 @@ def test_adjust_indices(self):
...
+                        "expected exactly 3 arguments");
+                    return HPy_NULL;
+                }
+
+                if (HPyArg_Parse(ctx, NULL, args, nargs, "OOO",
+                        &start, &stop, &step) < 0) {
+                    return HPy_NULL;
+                }
+                HPy length = HPySlice_New(ctx, start, stop, step);
+                return length;
```
fangerer, line 130: minor: this could just be `return HPySlice_New(...)`

hpy/tools/autogen/public_api.h
```diff
@@ -963,6 +963,15 @@ HPy HPyTuple_FromArray(HPyContext *ctx, const HPy items[], HPy_ssize_t n);
  *
  * :param ctx:
  *     The execution context.
+ * 
+ * :param start:
+ *     A pointer to a variable where to write the unpacked slice start. Must not
+ *     be ``NULL``.
```
fangerer, line 969: I guess this is just a copy-and-paste-error but those parameters are not *pointers*.
I would write something like: "A handle to an object to be used as slice start" or something (your English is better than mine :slightly_smiling_face:). Also, we allow `HPy_NULL` !
(Same for the other params.)

fangerer (approved):

PR #482, opened 2025-01-22
fixes from pypy

mattip:
- add proper function declarations
- skip a few tests that fail on PyPy

I found these when updating PyPy to work with latest HEAD here. It might be nice to release 0.9.1 so there is a tag to use for the cython benchmarks at https://github.com/DuToitSpies/msc_cython_benchmarks

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -254,7 +254,7 @@ jobs:
       # - template: azure-templates/python.yml
 
       - name: Install / Upgrade system requirements
-        run: sudo apt update && sudo apt install -y libclang-11-dev
+        run: sudo apt update && sudo apt install -y libclang-17-dev
 
       - name: Install / Upgrade Python requirements
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       # Duplicate changes to this matrix to 'poc_tests'
       matrix:
-        os: [ubuntu-latest, macos-11, windows-latest]
+        os: [ubuntu-latest, macos-13, windows-latest]
         python-version: ['3.9', '3.10']
         compiler: [""]
         include:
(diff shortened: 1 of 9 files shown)


mattip:
The image changed to ubuntu24.04 and now python3.10 installation is different?
```
sudo apt-get install -y --no-install-recommends python3.10-dev python3.10-venv python3-distutils python3.10-dbg
  Reading package lists...
  Building dependency tree...
  Reading state information...
  Package python3-distutils is not available, but is referred to by another package.
  This may mean that the package is missing, has been obsoleted, or
  is only available from another source
```

mattip:
And on windows, the build also fails
```
D:\a\hpy\hpy\hpy\universal\src\autogen_ctx_def.h(181): \
    error C2220: the following warning is treated as an error
  D:\a\hpy\hpy\hpy\universal\src\autogen_ctx_def.h(181): \
                     warning C4113: 'void (__cdecl *)(HPyContext *,HPyFunc_Signature,void *(__cdecl *)(),void *)' 
    differs in parameter lists from 'void (__cdecl *)(HPyContext *,HPyFunc_Signature,HPyCFunction,void *)'
  error: command 'C:\\Program Files\\Microsoft Visual Studio\\2022\\Enterprise\\VC\\Tools\\MSVC\\14.42.34433\\bin\\HostX86\\x64\\cl.exe' failed with exit code 2
```

hodgestar (approved):
Looks good to me.

mattip:
I am not sure what changed in the valgrind tests, the logs from the last successful run are gone. I think
- ubuntu-latest has changed
- python3 has changed.

I tried pinning to python3.11 but that did not help.

mattip:
Python 3.9 passes the valgrind tests, newer versions fail. Not clear if it is a false-positive or whether there is some refounting bug on later versions.

CI is passing, I will self-merge since @hodgestar approved the real changes.

PR #484, opened 2025-04-09
Allow building the C extension on GraalPy and running tests with it.

timfel:
With these changes we can build the HPy C extension on GraalPy and run the tests.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -1,4 +1,5 @@
 import pytest
+import sys
 from .support import ExtensionCompiler, DefaultExtensionTemplate,\
     PythonSubprocessRunner, HPyDebugCapture, make_hpy_abi_fixture
 from pathlib import Path
@@ -31,6 +32,13 @@ def pytest_configure(config):
     config.addinivalue_line(
         "markers", "syncgc: Mark tests that rely on a synchronous GC."
     )
+
+
+def pytest_runtest_setup(item):
+    if any(item.iter_markers(name="syncgc")):
+        if sys.implementation.name in ("pypy", "graalpy"):
+            pytest.skip("requires synchronous garbage collector")
+
 
 # this is the default set of hpy_abi for all the tests. Individual files and
 # classes can override it.
--- a/test/test_cpy_compat.py
+++ b/test/test_cpy_compat.py
@@ -1,4 +1,5 @@
-from .support import HPyTest, make_hpy_abi_fixture
+import pytest
+from .support import HPyTest, make_hpy_abi_fixture, IS_GRAALPY
 
 class TestCPythonCompatibility(HPyTest):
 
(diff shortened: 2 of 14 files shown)


test/conftest.py
```diff
@@ -32,6 +33,13 @@ def pytest_configure(config):
         "markers", "syncgc: Mark tests that rely on a synchronous GC."
     )
 
+
+def pytest_runtest_setup(item):
+    if any(item.iter_markers(name="syncgc")):
+        if sys.implementation.name in ("pypy", "graalpy"):
+            pytest.skip("requires syngc")
```
hodgestar, line 40:
```suggestion
            pytest.skip("requires synchronous garbage collector")
```

timfel:
> The previous PR merged to main was green, and now valgrind and python2.7 is failing. What changed?

The 2.7 is just that i added a skip using sys.implementation without guarding for 2.7

The valgrind issue seems this: https://github.com/hpyproject/hpy/actions/runs/14363896576/job/40272010157?pr=484#step:6:13
```
python3 setup.py build_clib -f build_ext -if
/home/runner/work/hpy/hpy/.eggs/setuptools_scm-8.2.0-py3.9.egg/setuptools_scm/_integration/setuptools.py:31: RuntimeWarning: 
ERROR: setuptools==58.1.0 is used in combination with setuptools-scm>=8.x

Your build configuration is incomplete and previously worked by accident!
setuptools-scm requires setuptools>=61
```

Looking into it 👀

timfel:
For valgrind I opened https://github.com/hpyproject/hpy/pull/485/files

mattip:
#485 is in, do you want to rebase to get that here?

test/test_cpy_compat.py
```diff
@@ -1,4 +1,5 @@
-from .support import HPyTest, make_hpy_abi_fixture
+import pytest
```
timfel, line 1: @mattip do these need to run as apptests in pypy?

test/test_hpyerr.py
```diff
@@ -1,4 +1,5 @@
-from .support import HPyTest, SUPPORTS_SYS_EXECUTABLE, trampoline
+import pytest
```
timfel, line 1: @mattip same question as for test_cpy_compat, do these need to run as apptests in pypy? If so I will pull the skip into the body of the methods like in test_object.py etc

timfel:
> The changes look good from my side. Happy to review again later once tests are passing.

Thanks, tests are passing now, just need someone to check if this'll work for PyPy apptests.

mattip:
All the `test_*` files are run untranslated on PyPy, so we cannot import pytest in the global level. On the other hand, it only becomes an issue if we pull latest HPy HEAD into PyPy, which will only happen if there is more development of PyPy + HPy. I am not sure that will happen, so I think we can merge the code as you have it here and refactor for PyPy when/if needed.

mattip:
It might be nice to make a release of HPy since HEAD includes the changes needed for Cython.

PR #490, opened 2025-05-16
Update supported Python versions (from 3.10 to 3.13)

paugier:
Mostly to check how much it is broken with Python 3.13.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,11 +12,9 @@ jobs:
       # Duplicate changes to this matrix to 'poc_tests'
       matrix:
         os: [ubuntu-latest, macos-13, windows-latest]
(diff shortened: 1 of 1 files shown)


hodgestar:
I would say let's start by cleaning up the simple issues and then have a look at the scarier one when the dust settles.

One simple route for debugging segfault's is the faulthandler module -- https://docs.python.org/3/library/faulthandler.html. After that it's GDB, I guess?

rlamy:
I think that all the segfaults are caused by CPython [now using](https://github.com/python/cpython/pull/116115) bit 2 of `tp_flags`, which clashes with `HPy_TPFLAGS_INTERNAL_IS_HPY_TYPE`. Unfortunately, there are no free bits left, so we can't just redo #304...

PR #491, opened 2025-05-21
Microbench: compatibility CPython 3.8 and example TestType::test_allocate_obj

paugier:
Tiny but useful because it shows that PyPy is much slower (typically 6 times slower) than CPython (3.8 and 3.12) for the benchmark `TestType::test_allocate_obj`.

It seems that it is a regression in PyPy (see https://github.com/pypy/pypy/issues/3816#issuecomment-2895973948).


mattip:
Thanks @paugier

Issue #493, opened 2026-01-12
Title: HPy Universal ABI crashes on Python 3.13 when calling type methods


koubaa:
HPy extensions using the Universal ABI crash with an access violation when calling methods on custom types in Python 3.13. The crash occurs during attribute lookup on instances (e.g., `obj.method()` or `getattr(obj, 'method')`), but calling via `type(obj).__dict__['method'].__get__(obj, type(obj))()` works correctly.

## Environment
Python version: 3.13.3 (MSC v.1943 64 bit AMD64)
HPy version: 0.9.0 (also reproduced with 0.9.1.dev from main branch)
OS: Windows 10/11 x64
Compiler: MSVC (Visual Studio 2022)
Steps to Reproduce
Run HPy's own test suite with Python 3.13:

## Expected Behavior
Test passes (as it does with Python 3.11)

## Actual Behavior
Additional Observations
CPython ABI works: `test_HPyDef_METH`[cpython] passes on Python 3.13
Python 3.11 works: Both Universal and CPython ABI tests pass on Python 3.11.4
Module-level methods work: `HPyDef_METH` at module level works correctly in Universal mode
Type slots work: `HPy_tp_new` slot works correctly - objects can be created
Explicit descriptor call works: `type(obj).__dict__['method'].__get__(obj, type(obj))()` succeeds
Instance attribute access crashes: obj.method crashes before the method is even called

## Minimal Reproduction
```
#include <hpy.h>

typedef struct { int value; } MyObject;
HPyType_HELPERS(MyObject)

HPyDef_METH(MyObject_test, "test", HPyFunc_NOARGS, .doc = "test")
static HPy MyObject_test_impl(HPyContext *ctx, HPy self) {
    return HPyLong_FromLong(ctx, 42);
}

HPyDef_SLOT(MyObject_new, HPy_tp_new)
static HPy MyObject_new_impl(HPyContext *ctx, HPy cls, const HPy *args,
                             HPy_ssize_t nargs, HPy kw) {
    MyObject *obj;
    HPy h = HPy_New(ctx, cls, &obj);
    if (HPy_IsNull(h)) return HPy_NULL;
    obj->value = 123;
    return h;
}

static HPyDef *MyObject_defines[] = { &MyObject_new, &MyObject_test, NULL };
static HPyType_Spec MyObject_spec = {
    .name = "mytest.MyObject",
    .basicsize = sizeof(MyObject),
    .defines = MyObject_defines,
};
[...]

PR #494, opened 2026-01-12
fix issue #493

koubaa:
HPy-defined types crashed with a segmentation fault on Python 3.13 when creating or instantiating types using the Universal ABI.


koubaa:
@vstinner could you please review this?

vstinner (commented):
The overall change LGTM, but I'm not a HPy developer so I cannot properly review the change.

paugier:
@koubaa 

Note that HPy does not support Python 3.13 (see https://github.com/hpyproject/hpy/pull/490).

koubaa:
> @koubaa
> 
> Note that HPy does not support Python 3.13 (see #490).

Understood, this fixes some of the issues for python3.13, this doesn't attempt to support it fully

